PR #3, opened 2019-10-03
Initial web client skeleton

zachmullen:


zachmullen:
@mgrauer if you're OK with where I've placed this, let's go ahead and merge it. The app code is going to get removed anyway on future topics.

mgrauer:
LGTM

PR #9, opened 2019-10-04
Ansible

dorukozturk:


ansible/playbook.yml
```diff
@@ -0,0 +1,44 @@
...
+      copy:
+        src: files/config
+        dest: "{{ ansible_env.HOME }}/.aws/config"
+
+    - name: Install Girder plugins
+      pip:
+        name:
+          - girder-oauth
+          - girder-sentry
+          - girder-hashsum-download
```
yarikoptic: should we add some analytics (google or just downloads) right away?

mgrauer: Good question.

Google analytics (the Girder plugin) will track interactions with the Girder web client, which we won't expose to end users, so there isn't a need for that plugin to be enabled.

The download statistics Girder plugin will track the number of times a file or item has been downloaded, but no more detailed statistics than that. We could enable this one very easily, but I'd like us to have clear requirements and expectations around any download stats tracking.

yarikoptic: would frontend have its own log (for future analysis, alike apache logs) of interactions with it?

mgrauer: the frontend currently doesn't have that. we would need to clarify requirements there if that is desired. it definitely won't happen before SfN.

ansible/playbook.yml
```diff
@@ -0,0 +1,44 @@
...
+        name: awscli
+        update_cache: true
+      become: true
+
+    - name: Create .aws directory
+      file:
+        path: "{{ ansible_env.HOME }}/.aws"
+        state: directory
+
+    - name: Copy aws config file
```
mgrauer: can you extend this `name` to `Copy aws config file required for public S3 bucket in separate aws account`?

ansible/playbook.yml
mgrauer: I will add download statistics plugin, see #11.

PR #17, opened 2019-10-07
Start of app list view

zachmullen:
* Login component
* Data browser
* Search component (placeholder)
* Nav bar with logout button


zachmullen:
@mgrauer much of the remaining front-end work here depends on stuff upstream in GWC. While I wait for that to get merged and/or released, I plan to move onto the backend stuff that needs to be done, so I'm OK with merging my current progress on the UI.

dandi_ui

mgrauer:
OK let's merge this. I assume we'll have to debug netlify/OAuth/CORS or something, might as well start now.

PR #19, opened 2019-10-08
Install dandi plugin as part of ansible playbook

dorukozturk:
I already ran the playbook and the plugin is installed on the production girder instance.

### DIFF
--- a/ansible/playbook.yml
+++ b/ansible/playbook.yml
@@ -31,6 +31,12 @@
         src: files/config
         dest: "{{ ansible_env.HOME }}/.aws/config"
 
+    - name: Copy dandi plugin
+      copy:
+        src: girder-dandi.tar.gz
+        dest: /tmp/girder-dandi.tar.gz
+        force: true
+
     - name: Install Girder plugins
       pip:
         name:
@@ -38,6 +44,7 @@
           - girder-sentry
           - girder-hashsum-download
           - girder-download-statistics
+          - /tmp/girder-dandi.tar.gz
         virtualenv: "{{ girder_virtualenv }}"
         state: latest
       notify:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,6 +28,13 @@ jobs:
           name: Change key permission
           command: chmod 600 "$HOME/girder-dandi-server.pem"
       - run:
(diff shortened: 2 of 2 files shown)


ansible/playbook.yml
```diff
@@ -31,13 +31,20 @@
...
+        force: true
+
     - name: Install Girder plugins
       pip:
         name:
           - girder-oauth
           - girder-sentry
           - girder-hashsum-download
           - girder-download-statistics
+          - /tmp/girder-dandi.tar.gz
```
yarikoptic: what is this beast on someones `/tmp`? ;)
Upload to pypi is quite easy, so ideally everything should go through it (on this note -- I will probably publish `dandi-cli` 0.0.2 later today to pypi)

danlamanna: I think it depends on the cadence of the release cycle. If we want to deploy every commit of master, then deploying from pypi will likely be problematic. We've seen pypi cache the old version of a package for up to 15 minutes after a new release. The most reliable and reproducible method of deploying the latest commit of a package is to build the sdist and copy it over. Otherwise we end up in a situation where we *think* version n is deployed when version n-1 is actually deployed.

If we feel the need to distribute girder-dandi to the masses (do we?), then we should definitely setup pypi packaging; though I think using it for deployment would still cause the above problems.

FWIW we tend to use https://pypi.org/project/setuptools-scm/ so every commit is pushed as a prerelease (see https://pypi.org/project/girder/#history for example).

yarikoptic: > If we feel the need to distribute girder-dandi to the masses (do we?),

eventually -- yes, for sure!  
ATM there is no urgency though, besides may be I might like to deploy my own instance to mess around.

Specifically here -- couldn't it just be a https://github.com/dandi/XXX/archive/master.zip where XXX would point to the repo providing that module (from aforementioned path I am not quite even sure what is that piece)?  then there would be no confusion

yarikoptic: ok, chatting with @mgrauer ATM -- I think I got it cleared up a bit.  Please disregard my whining on this respect -- I will not expect this ansible playbook be reusable outside of the CI deployment cycle.

Cheers!

PR #24, opened 2019-10-10
Sync selected items from parent component

zachmullen:


zachmullen:
The upstream change has been merged, so this now works beautifully.

zachmullen:
Scratch that, I just discovered a race condition as written. Documenting it here:

1. Click a search result, our store action is dispatched. This changes the location in the DataBrowser.
2. DataBrowser internally dispatches an XHR to fetch the children of the new location.
3. Our store action dispatches its own XHR to fetch the parent folder information of the item.

The XHRs in steps 2 and 3 both call `setSelected`; the DataBrowser internal one triggers `input` when it is done, with a value of `[]` to clear the selection. The store one sets it to the item that was clicked. So whichever XHR returns last wins, but we actually want the store's answer to always win out.

A workaround solution would be to have the upstream component trigger the input immediately rather than waiting for the location change XHRs to complete. If that's not robust enough, we can address it here by adding some additional state, though that might be messy.

zachmullen:
Actually this isn't a race between two XHRs, the set location is synchronous. The problem is that the store action calls set location and set selected back to back; the set location, by virtue of two way binding, is also triggering its own set selected that is overwriting the other set selected, because it doesn't run until the next tick due to Vue's event loop behavior.

What we need is a way to set the selected state only *after* the set location action is complete. A setImmediate sort of call will likely work.

PR #27, opened 2019-10-11
Add sentry snippet, configurable via env var

zachmullen:
I added this snippet and verified that the block is getting called, but I have no actual tokens to test that it worked. When I put some dummy string as the DSN, Sentry's client code doesn't complain, but it also doesn't appear to be doing anything with the Vue errors. I may need help to verify that this actually works.


zachmullen:
Update: just force pushed, this should "just work" now.

PR #29, opened 2019-10-11
Test org membership during OAuth login

zachmullen:
Note that in order for this to work, the oauth app must be granted special access at the *org* level here: https://github.com/organizations/dandi/settings/oauth_application_policy

Once it's approved, all users of that oauth client can have their org membership tested, regardless of whether their membership status is public or private.

As of now, if the test fails, we raise an exception, which ends up simply printing some ugly text in the user's browser. We really should do an HTTP redirect instead, and take them to a human-readable page telling them what they should do to be able to log into the system. I'm leaving that as future work for now.

### DIFF
--- a/girder-dandi-archive/girder_dandi_archive/__init__.py
+++ b/girder-dandi-archive/girder_dandi_archive/__init__.py
@@ -1,9 +1,35 @@
-from girder import plugin
+import os
+import requests
+
+from girder import events, plugin
+from girder.exceptions import AccessException
 from girder.utility import search
+from girder_oauth.providers import github
+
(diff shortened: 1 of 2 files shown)


zachmullen:
BTW, can I use f-strings? Are we running on py3.6+?

mgrauer:
I don't 100% know what Python we are running, but I'm fairly sure it is the system Python3 on ubuntu 18.04, which is 3.6.8.

This Girder instance was provisioned with `ansible-role-girder`, see this [issue](https://github.com/girder/ansible-role-girder/issues/4).

PR #33, opened 2019-10-16
Remove hashsum_download plugin

mgrauer:
Remove the girder hashsum_download plugin in order to provide a performance baseline as we start to upload files.

This is a "migration" style PR that should be followed up with a removal of the state:absent section.


yarikoptic:
but it will be brought back later on, right? (I have hopes to later integrate with datalad, and being able to request by the sha512 would make implementation much leaner/easier)

PR #36, opened 2019-10-18
Allow spaces in search values

mgrauer:
I'm going to merge this myself without QA.

### DIFF
--- a/girder-dandi-archive/girder_dandi_archive/__init__.py
+++ b/girder-dandi-archive/girder_dandi_archive/__init__.py
@@ -31,6 +31,8 @@ def dandiSearchHandler(query, types, user=None, level=None, limit=0, offset=0):
 def tokenize_search_string(search_string):
     search_string = search_string.lower()
     tokens = search_string.split()
+    # use '+' as a space inside a token, e.g. nuo+li => 'nuo li'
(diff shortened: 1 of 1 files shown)


mgrauer:
@danlamanna @dorukozturk 

This PR, when merged in, but didn't actually deploy the updated plugin. The ansible deploy [job](https://circleci.com/gh/dandi/dandiarchive/tree/master) on master reported no changes, and I reran it, but still no changes. Any ideas why?

Is this me getting punished for skipping QA?

mgrauer:
This isn't urgent, I fixed it in a dirty way for now, but would like to understand why the deploy didn't work as I expected.

dorukozturk:
I think the problem is the "Build girder web client" and "Restart Girder" handlers don't get triggered because ansible thinks there is no change, when we install our plugin from pypi. I tried the "forcereinstall" option but still ansible thought the system is in an unchanged state. So, even though we install a different sha from pypi, ansible thinks the state is unchanged so the handlers don't get triggered. I think that explains why we don't see the changes because we don't restart the server.  

I did a little research and found this post:
https://serverfault.com/questions/617548/always-trigger-handler-execution-in-ansible

This seems like a hack to me but at the moment I don't know a better solution (@danlamanna do you know a better solution?).

PR #43, opened 2019-10-20
Make sure handlers are always triggered

dorukozturk:


dorukozturk:
@mgrauer Dan verbally approved this after our discussion so I am merging this.

PR #50, opened 2019-10-29
Add local dev docs

dorukozturk:
@yarikoptic I added some instructions. Can you try and let me know whether I captured everything to get up and running locally or not?

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,3 +6,79 @@ Folders in this repo:
 - Ansible: Ansible code for deploying the girder.dandiarchive.org site.
 - web: Vue.js code for holding a Girder Web Components based client deployed at gui.dandiarchive.org .
 - girder-dandi-archive: A Girder plugin for customizing the DANDI archive Girder instance.
+
+## Developing Locally
+
+### 1. Server
+
+In order to get girder up and running with the required plugins, first create a python3 virtual environment.
+
+```
+mkvirtualenv dandiarchive --python=python3
+```
+
+Install girder dandiarchive plugin in editable mode.
+
+```
+cd girder-dandi-archive
+pip install -e .
+```
+
+Build girder client.
+
+```
+girder build
+```
+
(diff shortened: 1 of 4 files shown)


danlamanna:
Even though it's pretty obvious for a python project, it might be worth adding a sentence on how to run tests via `tox`.

dorukozturk:
@mgrauer I will wait for @yarikoptic to try before I merge.

README.md
```diff
@@ -6,3 +6,79 @@ Folders in this repo:
...
+```
+
+By default girder will start at port 8080 (http://localhost:8080).
+
+1) Create an admin user, using Girder's web client. Click on the "Register" button on Girder's UI.
+The first user created automatically becomes the admin user.
+
+2) Create a filesystem assetstore (http://localhost:8080/#assetstores).
+
+3) Set CORS settings so that the vue client can talk to the Girder server (http://localhost:8080/#settings).
```
yarikoptic: Is there a reasonably easy way to automate these three steps?

Ref: https://github.com/dandi/infrastructure/issues/7

README.md
```diff
@@ -6,3 +6,79 @@ Folders in this repo:
...
+pip install -e .
+```
+
+Build girder client.
+
+```
+girder build
+```
+
+Having a mongodb database locally is required to run Girder. If that is the case you are ready to get girder up and running.
```
yarikoptic: And if I don't have any?  Is there an easy way for a throw away instance (assuming I have it installed I guess) on some non privileged port?

README.md
dorukozturk: @yarikoptic I will create a docker-compose setup, which will include those three steps soonish. It will be another PR though.

README.md
dorukozturk: Again will do docker-compose which will include mongodb.

PR #53, opened 2019-10-30
Dockerize

### LINKED ISSUE #52: Create docker-compose setup for reproducible test environment

dorukozturk:
Fixes #52

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,3 +82,30 @@ Currently there are only server side tests. In order to run the tests:
     pip install -r requirements-dev.txt
     tox
 ```
+
+## Docker
+
+Currently there is a docker-compose file which gets the necessary infrastructure up and running for dandiarchive.
+
+In order to get up and running:
+
+```
+cd docker
+docker-compose build
+docker-compose up
+```
+
+This will spin up 4 containers:
+
+1) MongoDB
+2) Girder
+3) Vue Web Client
+4) Provision
+
+Provision container creates a Girder admin user, creates a filesystem assetstore and set necessary CORS settings.
+Credentials are username: admin, password: letmein.
+
+After docker-compose up succeeds you should have:
+
+1) Girder up and running on port 8091
+2) Web client up and running on port 8092
--- a/README.md
+++ b/README.md
@@ -105,7 +105,10 @@ This will spin up 4 containers:
 Provision container creates a Girder admin user, creates a filesystem assetstore and set necessary CORS settings.
 Credentials are username: admin, password: letmein.
 
-After docker-compose up succeeds you should have:
+When the provision container finishes, it will exit. The other three containers will remain running.
+
+After docker-compose up succeeds and the provision container finishes, you should have:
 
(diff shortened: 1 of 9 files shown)


docker/provision/girder.py
```diff
@@ -0,0 +1,63 @@
...
+def setGirderSetting(key, value, gc):
+    return gc.put("/system/setting",
+                  parameters={
+                      "key": key,
+                      "value": value
+                  })
+
+
+def createInitialUser(apiUrl):
+    # no try required here, just get a 400 on failure
```
mgrauer: can you change this to "no try/except required here". It will be slightly clearer to me.

README.md
```diff
@@ -82,3 +82,30 @@ Currently there are only server side tests. In order to run the tests:
...
+```
+
+This will spin up 4 containers:
+
+1) MongoDB
+2) Girder
+3) Vue Web Client
+4) Provision
+
+Provision container creates a Girder admin user, creates a filesystem assetstore and set necessary CORS settings.
```
mgrauer: Can you add this?

"When the provision container finishes, it will exit. The other three containers will remain running."

docker/provision/girder.py
```diff
@@ -0,0 +1,63 @@
...
+        print(exc)
+
+
+def check_admin_login(apiUrl):
+    gc = girder_client.GirderClient(apiUrl=apiUrl)
+
+    try:
+        gc.authenticate(ADMIN_USER, ADMIN_PASS)
+    except girder_client.AuthenticationError:
+        print('Cannot login admin, it probably does not exist')
```
mgrauer: 'Cannot login as "admin", that user probably does not exist'

README.md
```diff
@@ -82,3 +82,30 @@ Currently there are only server side tests. In order to run the tests:
...
+Credentials are username: admin, password: letmein.
+
+After docker-compose up succeeds you should have:
```
mgrauer: "After `docker-compose up` succeeds and the provision container finishes, you should have:"

README.md
```diff
@@ -82,3 +82,30 @@ Currently there are only server side tests. In order to run the tests:
...
+
+1) Girder up and running on port 8091
+2) Web client up and running on port 8092
```
mgrauer: 3) MongoDB up and running on port (?? whatever mongo runs on)

dorukozturk:
@mgrauer Inside the mongo container, mongo runs at 27017. We don't expose it to the host machine though. Docker-compose creates a network when you do docker-compose up and containers talk through that network. If you want containers to be accessible from the host system you expose them using the "ports" section in your docker-compose file. If you do not expose, mongo will be accessible only within the network that is generated by docker. Does that make sense?

README.md
dorukozturk: We don't expose mongo so we should not state anything about ports here (I think).

mgrauer: ok, can you add

3) MongoDB container up and running, and only visible to the other docker containers

Issue #41, opened 2019-10-19
Reflect current collection/folder in URL


yarikoptic:
ATM
![image](https://user-images.githubusercontent.com/39889/67146617-f43b0d80-f25a-11e9-8db4-8eb5623a6868.png)
which 
- makes impossible to share a URL to point to the specific collection/folder
- resets to the top page upon refresh
unlike in girder
![image](https://user-images.githubusercontent.com/39889/67146623-0157fc80-f25b-11e9-8318-cf059d7dade3.png)
providing `#anchor`s to point to specific collection/folder

mgrauer:
I understand this as a request to be able to permalink to some particular resource. I've posted an [issue](https://github.com/girder/girder_web_components/issues/196) in the upstream repo.

satra:
but this could be something we do through the frontend API not girder itself. but yes in general having a direct URL could be useful, but that may require more changes based on who has access to the URL. if you think of the get link option of dropbox

PR #63, opened 2019-11-22
Add URL routing for directory and selected item

matthewma7:
Add functionality to sync URL and browser location and selection.
However, due to the current implementation of DataBrowser, multi-selected items won't have correct metadata at page loads. However, this behavior probably is acceptable at this moment. 

Resolve https://github.com/dandi/dandiarchive/issues/41

### DIFF
--- a/web/src/router.js
+++ b/web/src/router.js
@@ -7,8 +7,13 @@ Vue.use(Router);
 export default new Router({
   routes: [
     {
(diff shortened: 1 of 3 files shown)


mgrauer:
Approved, merging, thanks.

I ❤️ the netlify preview deployment.

PR #64, opened 2019-11-23
Add a script to import metadata to girder for development purposes

dorukozturk:


mgrauer:
Approved, thanks!

@yarikoptic This completes your earlier request, to have a local dev setup with some initial metadata populated.

Issue #42, opened 2019-10-19
Adjust "Rows per page" to be larger numbers (by default and overall)


yarikoptic:
ATM default is 10, and options in dropdown are 5, 10, 15, All.  IMHO for our archive it would be logical to have default e.g. 25, and have options 10, 25, 50, All or about there.  5 is useless IMHO (unless we have a plan to provide some kind of expanded view per each file/item).

PR #66, opened 2019-11-25
Change browser page size

matthewma7:
With the upgraded GWC, it's possible to customize page sizes and the default page size. 
However, Girder doesn't have good support for 'all' option at this moment. So 'all' option is removed at this moment. 

Resolve https://github.com/dandi/dandiarchive/issues/42


mgrauer:
LGTM, merging.

PR #90, opened 2020-02-22
Repoint dandiarchive home page from gui footer

mgrauer:


mgrauer:
I confirmed this fix works from the preview, merging.

PR #94, opened 2020-02-27
Add create and get dandiset endpoints

### LINKED ISSUE #80: create staging dandiset endpoint
This endpoint API is subject to change 

* find or create the staging collection, which should be publicly readable
 * create a folder in staging collection
 * add a folder metadata of `dandiset_id:$(uuid)`
 * return the folder document

### LINKED ISSUE #88: Get dandiset endpoint
Create an API endpoint which takes a dandiset ID, and some optional parameters. 
It returns (type: a girder folder document):
* The latest published dandiset, if no options were passed.
* If the option for a version was specified, return that version.
* The above option could also include the staging version.

AlmightyYakob:
Fixes #88.
Fixes #80.

This PR adds two endpoints:

1. An endpoint to create a dandiset, creating a folder with the specified name under the staging collection and assigning it an ID
2. An endpoint to get the girder folder document from a given dandiset ID

If viewing the file changes, please ignore the first commit, as it's just a formatting change.

### DIFF
--- a/girder-dandi-archive/girder_dandi_archive/rest.py
+++ b/girder-dandi-archive/girder_dandi_archive/rest.py
@@ -13,50 +13,52 @@ from .util import (
 )
 
 
-@access.user
-@describeRoute(
-    Description("Create Dandiset")
-    .param("name", "Name of the Dandiset.")
-    .param("description", "Description of the Dandiset.")
-)
-def create_dandiset(params):
-    name, description = params["name"], params["description"]
-    exists = Folder().findOne({"name": name})
-
-    if exists:
-        raise RestException("Dandiset already exists", code=409)
-
-    current = Setting().get(DANDISET_ID_COUNTER)
-    if current is None:
-        new_id_count = Setting().set(DANDISET_ID_COUNTER, 0)
-    else:
-        new_id_count = Setting().set(DANDISET_ID_COUNTER, current + 1)
-
-    padded_id = pad_dandiset_id(new_id_count["value"])
-    meta = {"name": name, "description": description, "id": padded_id}
-
-    staging = staging_collection()
-    folder = Folder().createFolder(
-        staging, name, parentType="collection", public=False,
-    )
-    folder = Folder().setMetadata(folder, {"dandiset": meta})
-    return folder
-
-
-@access.public
-@describeRoute(Description("Get Dandiset").param("id", "Dandiset ID"))
-def get_dandiset(params):
-    if not validate_dandiset_id(params["id"]):
-        raise RestException("Invalid Dandiset ID")
-
-    doc = Folder().findOne({"meta.dandiset.id": params["id"]})
-    return doc
-
-
 class DandiResource(Resource):
     def __init__(self):
         super(DandiResource, self).__init__()
 
         self.resourceName = "dandi"
-        self.route("GET", (), get_dandiset)
-        self.route("POST", (), create_dandiset)
+        self.route("GET", (), self.get_dandiset)
+        self.route("POST", (), self.create_dandiset)
+
+    @access.user
+    @describeRoute(
+        Description("Create Dandiset")
+        .param("name", "Name of the Dandiset.")
+        .param("description", "Description of the Dandiset.")
+    )
+    def create_dandiset(self, params):
+        name, description = params["name"], params["description"]
+        exists = Folder().findOne({"name": name})
+
+        if exists:
+            raise RestException("Dandiset already exists", code=409)
+
+        current = Setting().get(DANDISET_ID_COUNTER)
+        if current is None:
+            new_id_count = Setting().set(DANDISET_ID_COUNTER, 0)
+        else:
+            new_id_count = Setting().set(DANDISET_ID_COUNTER, current + 1)
+
+        padded_id = pad_dandiset_id(new_id_count["value"])
+        meta = {"name": name, "description": description, "id": padded_id}
+
+        staging = staging_collection()
+        folder = Folder().createFolder(
+            staging,
+            name,
+            parentType="collection",
+            creator=self.getCurrentUser(),
+            public=False,
+        )
+        folder = Folder().setMetadata(folder, {"dandiset": meta})
+        return folder
+
+    @access.public
+    @describeRoute(Description("Get Dandiset").param("id", "Dandiset ID"))
+    def get_dandiset(self, params):
+        if not validate_dandiset_id(params["id"]):
+            raise RestException("Invalid Dandiset ID")
+
+        doc = Folder().findOne({"meta.dandiset.id": params["id"]})
+        return doc
--- a/girder-dandi-archive/girder_dandi_archive/rest.py
+++ b/girder-dandi-archive/girder_dandi_archive/rest.py
@@ -28,7 +28,14 @@ class DandiResource(Resource):
         .param("description", "Description of the Dandiset.")
     )
     def create_dandiset(self, params):
+        if "name" not in params or "description" not in params:
+            raise RestException("Name and description required.")
+
         name, description = params["name"], params["description"]
+
+        if not name or not description:
+            raise RestException("Name and description must not be empty.")
+
         exists = Folder().findOne({"name": name})
 
         if exists:
--- a/girder-dandi-archive/girder_dandi_archive/rest.py
+++ b/girder-dandi-archive/girder_dandi_archive/rest.py
@@ -5,7 +5,12 @@ from girder.models.setting import Setting
 from girder.models.folder import Folder
 from girder.exceptions import RestException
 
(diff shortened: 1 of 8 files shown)


girder-dandi-archive/girder_dandi_archive/util.py
```diff
@@ -0,0 +1,38 @@
...
+    if len(str(doc["value"])) > DANDISET_ID_LENGTH:
+        raise ValidationException("Dandiset ID limit exceeded", "value")
+
+
+def validate_dandiset_id(dandiset_id):
+    return bool(re.match(dandiset_id_pattern, dandiset_id))
+
+
+def pad_dandiset_id(dandiset_id):
+    id_str = str(dandiset_id)
```
brianhelba: You can do all of this with just:
```
return f"{dandiset_id:0{DANDISET_ID_LENGTH}d}"
```

Also, `pad_dandiset_id` probably doesn't need to be a utility function (i.e. just inline the f-string expression where it's used), since the IDs should always be handled as strings (except when they're created).

girder-dandi-archive/girder_dandi_archive/util.py
```diff
@@ -0,0 +1,38 @@
...
+DANDISET_ID_LENGTH = 6
+DANDI_STAGING_COLLECTION_NAME = "Dandiset Staging"
+
+dandiset_id_pattern = r"^\d{6}$"
+
+
+@setting_utilities.validator(DANDISET_ID_COUNTER)
+def _validate(doc):
+    if len(str(doc["value"])) > DANDISET_ID_LENGTH:
+        raise ValidationException("Dandiset ID limit exceeded", "value")
```
brianhelba: If this happens, it's a huge problem to migrate, so ensure that you'll never have more than one million IDs issued. Ever.

girder-dandi-archive/girder_dandi_archive/rest.py
```diff
@@ -0,0 +1,64 @@
...
+        if current is None:
+            new_id_count = Setting().set(DANDISET_ID_COUNTER, 0)
+        else:
+            new_id_count = Setting().set(DANDISET_ID_COUNTER, current + 1)
+
+        padded_id = pad_dandiset_id(new_id_count["value"])
+        meta = {"name": name, "description": description, "id": padded_id}
+
+        staging = staging_collection()
+        folder = Folder().createFolder(
```
brianhelba: Is there a reason you're not just setting the `description` on the new folder?

girder-dandi-archive/girder_dandi_archive/rest.py
```diff
@@ -0,0 +1,64 @@
...
+            staging,
+            name,
+            parentType="collection",
+            creator=self.getCurrentUser(),
+            public=False,
+        )
+        folder = Folder().setMetadata(folder, {"dandiset": meta})
```
brianhelba: Why are you duplicating the `name` (and potentially `description`) into the metadata, when it already exists on the top-level folder itself?

girder-dandi-archive/girder_dandi_archive/util.py
```diff
@@ -0,0 +1,38 @@
...
+    remaining = DANDISET_ID_LENGTH - length
+
+    return "%s%s" % (remaining * "0", id_str)
+
+
+def staging_collection():
+    doc = Collection().findOne({"name": DANDI_STAGING_COLLECTION_NAME})
+
+    if doc is None:
+        doc = Collection().createCollection(DANDI_STAGING_COLLECTION_NAME, public=False)
```
brianhelba: To be clear, this (and the dandiset folders within) will be hidden to everyone, except site admins?

girder-dandi-archive/girder_dandi_archive/util.py
```diff
@@ -0,0 +1,38 @@
...
+def pad_dandiset_id(dandiset_id):
+    id_str = str(dandiset_id)
+    length = len(str(id_str))
+    remaining = DANDISET_ID_LENGTH - length
+
+    return "%s%s" % (remaining * "0", id_str)
+
+
+def staging_collection():
+    doc = Collection().findOne({"name": DANDI_STAGING_COLLECTION_NAME})
```
brianhelba: You could make this a one-line function as:
```
return Collection().createCollection(
    DANDI_STAGING_COLLECTION_NAME, 
    public=False.
    reuseExisting=True,
)
```

girder-dandi-archive/girder_dandi_archive/rest.py
```diff
@@ -0,0 +1,64 @@
...
+
+        padded_id = pad_dandiset_id(new_id_count["value"])
+        meta = {"name": name, "description": description, "id": padded_id}
+
+        staging = staging_collection()
+        folder = Folder().createFolder(
+            staging,
+            name,
+            parentType="collection",
+            creator=self.getCurrentUser(),
```
brianhelba: Since the parent collection is totally private, this `creator` may not reliably be able to view this dandiset folder (even though they have default permissions to do so). I'm not sure if that's a problem, just an observation.

girder-dandi-archive/girder_dandi_archive/rest.py
```diff
@@ -0,0 +1,64 @@
...
+        folder = Folder().setMetadata(folder, {"dandiset": meta})
+        return folder
+
+    @access.public
+    @describeRoute(Description("Get Dandiset").param("id", "Dandiset ID"))
+    def get_dandiset(self, params):
+        if not validate_dandiset_id(params["id"]):
+            raise RestException("Invalid Dandiset ID")
+
+        doc = Folder().findOne({"meta.dandiset.id": params["id"]})
```
brianhelba: This is an unindexed query! For this to scale beyond tiny sizes, you'll need to add an index to `meta.dandiset.id`.

girder-dandi-archive/girder_dandi_archive/rest.py
```diff
@@ -0,0 +1,64 @@
...
+        self.route("POST", (), self.create_dandiset)
+
+    @access.user
+    @describeRoute(
+        Description("Create Dandiset")
+        .param("name", "Name of the Dandiset.")
+        .param("description", "Description of the Dandiset.")
+    )
+    def create_dandiset(self, params):
+        name, description = params["name"], params["description"]
```
brianhelba: `@describeRoute` just updates Swagger docs. It does no validation of `params`.

In the case that a caller fails to pass `name` or `description` in the form body, the server will 500 with a `KeyError`.

brianhelba: Also, do you want to allow `name` to be empty?

girder-dandi-archive/setup.cfg
```diff
@@ -1,2 +1,6 @@
 [bdist_wheel]
 universal = 1
+
+[flake8]
```
brianhelba: `setup.cfg` is on the way out in the Python ecosystem. I'd put this config block in `tox.ini`.

girder-dandi-archive/girder_dandi_archive/rest.py
```diff
@@ -0,0 +1,64 @@
...
+        exists = Folder().findOne({"name": name})
+
+        if exists:
+            raise RestException("Dandiset already exists", code=409)
+
+        current = Setting().get(DANDISET_ID_COUNTER)
+        if current is None:
```
brianhelba: It'd be a little simpler to have:
```
current = Setting().get(DANDISET_ID_COUNTER)
if current is None:
    current = -1
new_id_count = Setting().set(DANDISET_ID_COUNTER, current + 1)
```

girder-dandi-archive/girder_dandi_archive/rest.py
AlmightyYakob: The `name` and `description` fields are just the initial required parts of a larger schema that's stored in this metadata. Because the idea is to have all metadata of a dandiset live under `meta.dandiset`, I don't think it makes sense to separate this.

girder-dandi-archive/girder_dandi_archive/rest.py
AlmightyYakob: Answered below

girder-dandi-archive/girder_dandi_archive/rest.py
AlmightyYakob: Good catch

yarikoptic:
pardon my naive question: how do I access the added endpoint? couldn't read from the diff/code

mgrauer:
Not a naive question.

You can create a Dandiset through the web api:
`https://girder.dandiarchive.org/api/v1#!/dandi/dandi_create_dandiset`
Then if you want to get it via the web api, passing in the dandiset id:
`https://girder.dandiarchive.org/api/v1#!/dandi/dandi_get_dandiset`

PR #106, opened 2020-03-02
BF: Girder folder for created dataset should be named with the id

yarikoptic:
not metadata 'name'.  Since otherwise we get something like https://girder.dandiarchive.org/#collection/5e59bb0af19e820ab6ea6c62 -- where name, later edited, also does not even correspond to the name of the folder.

### DIFF
--- a/girder-dandi-archive/girder_dandi_archive/rest.py
+++ b/girder-dandi-archive/girder_dandi_archive/rest.py
@@ -51,7 +51,7 @@ class DandiResource(Resource):
 
         staging = staging_collection()
         folder = Folder().createFolder(
-            staging, name, parentType="collection", creator=self.getCurrentUser(),
+            staging, padded_id, parentType="collection", creator=self.getCurrentUser(),
         )
         folder = Folder().setMetadata(folder, {"dandiset": meta})
         return folder


mgrauer:
This should not have been merged, it will break. Let's slow down please.

yarikoptic:
Oops, why would it break?

mgrauer:
This PR wasn't a big problem, but it was good because it surfaced some issues about our process:
 * how much time is reasonable before merging: obviously this depends on a number of factors, the more complex the PR the longer the window should be (but also try to avoid big PRs!), more time should be given if a PR will affect saved state and require a migration--we should probably have a way of marking these. At least 24 hours during the working week is a reasonable expectation for QA, though if it is important to get merged for some reason sooner, reach out to people on slack and let them know the urgency (but also, urgency leads to poor decision making).
 * code review: tests passing isn't enough to merge a PR. Sometimes there are no tests, sometimes the tests won't cover the right things. A human should read the changes, check out the code, try it locally, and think about the blast radius and the understandability and maintainability of the changes. Ideally the reviewer is someone familiar with the area in question. Again, PRs that affect saved state and require migrations of existing data should have a much higher burden of QA than something simple like a UI layout change, and should include a discussion of the migration plan. If you are reviewing and aren't sure, just say so in the PR and someone else can come help. It's not a bad thing to have multiple eyes on a PR.


Specifics about this PR:

The way we look to see if a dandiset already exists will no longer work when we change the name of the folder to the id. 

Do we want to allow multiple dandisets to exist with the same name, as long as they have different ids? If so, we can remove that search/enforcement from earlier in the changed function.

This change also requires a migration of existing state in the production site, changing the name of existing folders to be their dandiset ids, and there should be a plan for this migration. In this case it is fairly simple, but we need to think about migrations carefully in the future.

Another minor issue here is that it will be hard to see what is in a dandiset with the current UI, as we'll just see a list of folders with names like `000000`, `000001`. At a future point the more custom UI for the dandiset dashboards can display the name of the dandiset.

satra:
> Do we want to allow multiple dandisets to exist with the same name, as long as they have different ids? If so, we can remove that search/enforcement from earlier in the changed function.

yes, names can change anytime, id's cannot

> we need to think about migrations carefully in the future.

completely agree, but right now we want to move fast and get to a state that is publicly presentable and usable.

>  it will be hard to see what is in a dandiset with the current UI

i agree, but the expectation is that we do have the public dashboard and dataset landing page soon.

we should discuss processes and expectations today on the call.

PR #111, opened 2020-03-03
Remove icons

JellyWang-3:
I removed the "view item" and "open in jupyter" icons.


AlmightyYakob:
Not sure why those yarn files were there, but I removed them. I also reverted the `.env.development` file, as changes to that shouldn't be tracked.

mgrauer:
Nice work @JellyWang-3, thanks for your first DANDI PR. Thanks @AlmightyYakob for the QA.

PR #112, opened 2020-03-03
fix: client docker file needed modules installed

satra:


.gitignore
```diff
@@ -102,3 +102,6 @@ venv.bak/
 
 # mypy
 .mypy_cache/
+
+# pycharm
+**/.idea
```
satra:
```suggestion
**/.idea

```

yarikoptic:
FWIW -- seems to work for me -- build finished without crashing

mgrauer:
I approved but @yarikoptic feel free to approve anything in the Docker if it works for you. I'm not worried about data migrations for this--the Docker-compose is just a developer tool.

PR #114, opened 2020-03-03
Replaced Open in Jupyter with Open JupyterLab

JellyWang-3:


mgrauer:
@AlmightyYakob can you qa the vue?
@yarikoptic can you test that this goes to the right place, from the preview link?

web/src/views/Home.vue
```diff
@@ -31,8 +31,25 @@ import {
...
+    for: ['item'],
+    name: 'Open JupyterLab',
+    icon: 'mdi-language-python',
+    color: 'primary',
+    handler() {
+      const { value: items } = this;
+      window.open(`${JUPYTER_ROOT}`, '_blank');
+    },
+   },
+  ...keys,
```
AlmightyYakob: No need to have an extra variable, you could just do it inline:

```suggestion
  ...DefaultActionKeys.slice(1),
```

web/src/views/Home.vue
```diff
@@ -31,8 +31,25 @@ import {
   getPathFromSelected,
 } from '@/utils';
 
+// redirect to "Open JupyterLab"
+const JUPYTER_ROOT = process.env.JUPYTER_ROOT || 'http://hub.dandiarchive.org ';
+
 // removed "view item" icons
-const actionKeys = DefaultActionKeys.slice(1);
+let keys = DefaultActionKeys.slice(1);
```
AlmightyYakob: If committing the suggestion below, this would go away:
```suggestion
```

web/src/views/Home.vue
```diff
@@ -31,8 +31,25 @@ import {
   getPathFromSelected,
 } from '@/utils';
 
+// redirect to "Open JupyterLab"
+const JUPYTER_ROOT = process.env.JUPYTER_ROOT || 'http://hub.dandiarchive.org ';
```
AlmightyYakob: Where else in the code is the env var `JUPYTER_ROOT` being used? Or more generally why is this being done this way? If we're never actually injecting this environment variable, I'd say we should just hard code this for now, as that's what's essentially happening anyway if we're never setting the `JUPYTER_ROOT` environment variable.

```suggestion
const JUPYTER_ROOT = 'http://hub.dandiarchive.org';
```

web/src/views/Home.vue
```diff
@@ -31,8 +31,25 @@ import {
...
+let keys = DefaultActionKeys.slice(1);
+
+const actionKeys = [
+   {
+    for: ['item'],
+    name: 'Open JupyterLab',
+    icon: 'mdi-language-python',
+    color: 'primary',
+    handler() {
+      const { value: items } = this;
```
AlmightyYakob: I don't see this being used anywhere. Let me know if I'm missing something, but otherwise this can go away:

```suggestion
```

web/src/views/Home.vue
yarikoptic: either way is ok with me.  Eventually may be there would be a way to configure for local deployments to have own local jupyterlab to connect to.  Leaving it available might simplify some testing later on. But we could always bring it back when need arises.

yarikoptic:
> @yarikoptic can you test that this goes to the right place, from the preview link?

seems to be the right one, went there just fine.

web/src/views/Home.vue
satra:
```suggestion
const JUPYTER_ROOT = process.env.JUPYTER_ROOT || 'https://hub.dandiarchive.org ';
```

web/src/views/Home.vue
JellyWang-3: I will change it.

web/src/views/Home.vue
JellyWang-3: Thank you and I will change it to `const JUPYTER_ROOT = 'http://hub.dandiarchive.org';`

web/src/views/Home.vue
JellyWang-3: I will take it out.

web/src/views/Home.vue
JellyWang-3: Thank you for your suggestion.

mgrauer:
This PR is failing checks in Netlify. It might just be a temporary issue in Netlify, when I looked at one of the failing checks all I could see in the logs was "System error".

When I look at the deploy preview page, I do not see the Jupyter link that this PR is trying to add. It could be the Netlify error. @JellyWang-3 is this link added for you when you build locally? If so, could you add a screenshot of it to the PR?

JellyWang-3:
dandiView

The link works for me when I built locally.

mgrauer:
Thanks. I'll assume it is a netlify issue and let's merge.

PR #115, opened 2020-03-03
RF: "drafts" and "releases" will be top level collections

yarikoptic:
First two commits deal only with docker compose devel setup, but the last one renames staging into drafts in the code

### DIFF
--- a/docker/provision/populate_data.py
+++ b/docker/provision/populate_data.py
@@ -21,7 +21,12 @@ def ensure_collection(collection_name):
 
     return gc.createCollection(collection_name)
 
-staging_collection = ensure_collection("staging")
+
+for c in ("drafts", "releases"):
+    ensure_collection(c)
+
+# labs is primarily for practicing metadata search etc, to be gone later on
+# and will not be in the deployed instance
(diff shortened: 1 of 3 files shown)


yarikoptic:
I would not dare to merge, @mgrauer to do the honors? ;)

mgrauer:
I will merge, but I hope you didn't read my note of caution as a way of saying I don't want to be collaborative. If a PR has been QAed and approved, it should be OK to merge by the author (or anyone really).

My main points in that message were (1) that I hope as part of QA we will do code review, and (2) we should be extra careful whenever a PR will require a data migration.

This PR will require a data migration, btw.

yarikoptic:
> I will merge, but I hope you didn't read my note of caution as a way of saying I don't want to be collaborative.

I didn't, no worries.  Just didn't want to break anything by the last commit in the series ;)

> If a PR has been QAed and approved, it should be OK to merge by the author (or anyone really).

ok

> This PR will require a data migration, btw.

even in our deployment which IMHO yet to go through a full cleansing (of the bucket etc) and probably "let's deploy from the clean slate"?

Issue #109, opened 2020-03-02
Fix find in create dandiset endpoint


mgrauer:
First add tests for the create and get dandiset endpoints, then fix this.

mgrauer:
Change name of staging collection from `Dandiset Staging` to `staging`. This will require a migration.

mgrauer:
correction `Dandiset Staging` to `drafts`. published is `releases`. these will live in different collections.

mgrauer:
This was closed by #120 and all data migrations have been completed:

 - `drafts` collection exists
 - all previous folders in `drafts` have been deleted, so that any new ones will have a correct dandiset `identifier` key and value

Issue #105, opened 2020-03-02
Make girder metadata records "structured"


yarikoptic:
ATM, metadata editor for a dataset dumps the record straight as a metadata of that folder, see e.g. https://girder.dandiarchive.org/#collection/5e59bb0af19e820ab6ea6c62/folder/5e5c0c6c513f45fe78c07491 .  For the files uploaded by `dandi-cli` I was dumping into the first level not only metadata extracted from nwb files, but also `uploaded_*` fields and might later add additional auxiliary records such as `checksums`.  Partially that was done because the `gui.'s` search implementation ATM considers only flat structure.
I think it would be worthwhile making metadata record more structured, by not (ab)using first level, making it hard to separate different types of metadata, but adding top level sections:

- `dandi`: for the top folder of a dandiset it would be the dandiset level metadata record (mirroring `dandiset.yaml` content).  For a file - file metadata, e.g. extracted and uploaded by `dandi upload`
- `dandi-client`: anything dandi client(s) would like to communicate.  E.g. it could be the `uploaded-*` fields, which would allow to identify if file was modified by girder (by comparing mtime of the file asset, or relying on checksums). We could later extend this record with e.g. `history` for https://github.com/dandi/dandi-cli/issues/43 or any other information which is not directly data "metadata", but rather metadata to assist our technologies.

It might be that when/if we decide to implement some other solutions, like versioning, having flexibility of extending metadata records with top-level entries could be of help to assist with experimentation, and later migrations from one scheme to another (e.g. that `history` into something more generic like `dandi-versioning`)

yarikoptic:
actually we might even split `dandi` into two `dandiset` and `dandifile` to make it explicit from metadata record alone what schema to validate that entry against (for a file vs for a dandiset).

mgrauer:
I agree we should make this structured.

@AlmightyYakob and I have constructed a metadata schema for Girder folders that are dandisets, that currently looks like this, as in #120. We only have draft dandisets so far.

```
meta
    dandiset
        identifier: '000000'
        version: 'draft'
        name: 'dandiset name'
        description: 'dandiset description'
```

yarikoptic:
Fwiw I will repeat my dislike of anything but semantic (or some other sensible, eg date based?) version as a version. At least it could be based on semantic version eg `0.0.0.draft` when no version released, and otherwise be there .draft suffix attached to the last released version

satra:
eventually we can get to something like:
```
$ git describe
[...]

PR #120, opened 2020-03-04
Test dandiset endpoints

### LINKED ISSUE #107: "id" vs "identifier" - we need to converge or diverge ;)
our schema (https://github.com/dandi/dandiarchive/blob/master/web/src/assets/schema/base.json) has `identifier`, not `id`.  https://github.com/dandi/dandiarchive/blob/master/girder-dandi-archive/girder_dandi_archive/rest.py populates and uses in the route `id` not `identifier`.  That leads to having both in the metadata: see https://girder.dandiarchive.org/#collection/5e59bb0af19e820ab6ea6c62/folder/5e5c0c6c513f45fe78c07491 currently having 

```
  "id": "000001",
  "identifier": "000001",
```

Are those (`id` vs `identifier`) to be the same thing, or different?  If the same -- we should unify, probably into `id` since shorter etc (thus - adjust `schema/base.json`).  If different, e.g. `identifier` would be something like an identifier from identifiers.org, then should become `dandi:{id}/staging` (which we then change to the versioned one upon publish), but then we would need to keep both `id` and `identifier` fields in metadata (and thus schema).

attn: @satra @mgrauer

mgrauer:
Fixes #107, #109.

For #107, this PR updates all uses of `id` in dandiset metadata to `identifier`, and adds in a dandiset metadata key `version`.

This PR also builds on top of #115 by changing all uses of `staging` to `drafts`. It uses `draft` as the initial dandiset metadata value for the key `version`.

After we merge this PR we will need to do the following migrations:
 - remove the dandiset identifier counter setting
 - redo the two existing dandisets (we should just re-create them as that is safer) to give them the correct metadata structure and names

### DIFF
--- a/girder-dandi-archive/tests/test_rest.py
+++ b/girder-dandi-archive/tests/test_rest.py
@@ -1,6 +1,6 @@
 import pytest
 
-from girder_dandi_archive.util import create_drafts_collection
+from girder_dandi_archive.util import get_or_create_drafts_collection
 
 from girder.models.collection import Collection
 from girder.models.folder import Folder
@@ -8,6 +8,9 @@ from girder.models.folder import Folder
 from pytest_girder.assertions import assertStatus, assertStatusOk
 
 
+# TODO: This fixture and these tests are not well structured and hard to maintain.
+# They currently work correctly in terms of the application code that they test,
+# but they should be rewritten and not followed as an example.
 @pytest.fixture
 def draftsFolders(db):
     red_herring_collection = Collection().createCollection(
@@ -33,7 +36,7 @@ def draftsFolders(db):
         "red_herring_collection": red_herring_collection,
         "red_herring_folder": red_herring_dandiset_000000_folder,
     }
-    return_dict["drafts_collection"] = create_drafts_collection()
+    return_dict["drafts_collection"] = get_or_create_drafts_collection()
     yield return_dict
 
 
@@ -93,8 +96,7 @@ def testGetDandiset(server, draftsFolders, user, capsys):
         user=user,
         params={"identifier": "000000", "version": "draft"},
     )
-    assertStatusOk(resp)
-    assert resp.json == {}
+    assertStatus(resp, 400)
 
     # Create a Dandiset.
     resp = server.request(
--- a/girder-dandi-archive/tests/test_rest.py
+++ b/girder-dandi-archive/tests/test_rest.py
@@ -1,0 +1,127 @@
+import pytest
+
+from girder_dandi_archive.util import create_staging_collection
+
+from girder.models.collection import Collection
+from girder.models.folder import Folder
+
+from pytest_girder.assertions import assertStatus, assertStatusOk
+
+
+@pytest.fixture
+def stagingFolders(db):
+    red_herring_collection = Collection().createCollection(
+        "red_herring_collection", reuseExisting=True
+    )
+    # A folder that matches dandiset metadata, but not in staging collection.
+    red_herring_dandiset_000000_folder = Folder().createFolder(
+        parent=red_herring_collection,
+        parentType="collection",
+        name="000000",
+        public=True,
+    )
+    meta = {
+        "dandiset": {
+            "name": "red",
+            "description": "herring",
+            "identifier": "000000",
+            "version": "draft",
+        }
+    }
+    Folder().setMetadata(red_herring_dandiset_000000_folder, meta)
+    return_dict = {
+        "red_herring_collection": red_herring_collection,
+        "red_herring_folder": red_herring_dandiset_000000_folder,
+    }
+    return_dict["staging_collection"] = create_staging_collection()
+    yield return_dict
+
+
+@pytest.mark.plugin("dandi_archive")
+def testCreateDandiset(server, stagingFolders, user):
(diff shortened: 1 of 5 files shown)


mgrauer:
@AlmightyYakob Could you QA this?

One additional question for you: this failed on linting with flake8-black, but I ran black on it locally before committing. Is there a version or a config difference that I should be aware of?

girder-dandi-archive/tests/test_rest.py
```diff
@@ -0,0 +1,127 @@
...
+    test_dandiset_0_folder_meta = test_dandiset_0_folder["meta"]["dandiset"]
+
+    assert drafts_collection_id == test_dandiset_0_folder["parentId"]
+    assert "000000" == test_dandiset_0_folder["name"]
+    assert "000000" == test_dandiset_0_folder_meta["identifier"]
+    assert "test dandiset 0 name" == test_dandiset_0_folder_meta["name"]
+    assert "test dandiset 0 description" == test_dandiset_0_folder_meta["description"]
+    assert "draft" == test_dandiset_0_folder_meta["version"]
+
+    resp = server.request(
```
brianhelba: Technically, this is a separate test case, and it should live in a separate function. Given that it's trying to test stateful behavior (creation when at least one other Dandiset exists), you could initialize it by either artificially incrementing `DANDISET_IDENTIFIER_COUNTER` at the start of the case, or alternatively (and less artificial) you could make a `dandiset` fixture to ensure you start the case with another Dandiset already in the database.

Practically, I don't think you *need* to change this, but it's a good habit to make much smaller, more granular test cases that follow the pattern:
```
# set up
# execute
# assert
```
rather than large test cases which attempt to pack multiple executions into one.

girder-dandi-archive/tests/test_rest.py
```diff
@@ -0,0 +1,127 @@
...
+        path="/dandi",
+        method="GET",
+        user=user,
+        params={"identifier": "000000", "version": "draft"},
+    )
+    assertStatusOk(resp)
+    assert resp.json == {}
+
+    # Create a Dandiset.
```
brianhelba: This setup logic to create the Dandiset should live in a fixture, since it's a required prerequisite, not the actual thing being tested in this case.

girder-dandi-archive/tests/test_rest.py
```diff
@@ -0,0 +1,127 @@
...
+        params={
+            "name": "test dandiset 0 name",
+            "description": "test dandiset 0 description",
+        },
+    )
+    assertStatusOk(resp)
+    assert "000000" == resp.json["meta"]["dandiset"]["identifier"]
+
+    # Ensure we can retrieve the Dandiset.
+    resp = server.request(
```
brianhelba: Again, this is a separate test case, and should live in a separate test function.

Ideally, you'd have two functions:
```
def test_get_nonexistant_dandiset(...)
def test_get_dandiset(dandiset_id, ...)
```
The functionality being tested and the necessary prerequisites should be apparent from the function name and fixtures.

mgrauer:
@AlmightyYakob Could you sanity check this, as @brianhelba is out of the office this week?

mgrauer:
After reading through #105 more carefully, I agree that we don't have `version` settled enough as a field on the dandiset to include, so I have removed it from this PR. It's easy enough to add back in once we have come to a conclusion.

girder-dandi-archive/tox.ini
```diff
@@ -73,7 +73,13 @@ ignore =
     W503,
 
 [pytest]
-addopts = --verbose --strict --showlocals --cov-report="term"
+# We use the `--forked` option to run tests in subprocesses, which is much slower.
```
danlamanna: Add this link to the comment: https://github.com/girder/girder/issues/3206

girder-dandi-archive/tox.ini
```diff
@@ -73,7 +73,13 @@ ignore =
...
+# This is necessary due to the use of the `server` pytest fixture combined with a
```
danlamanna:
```suggestion
# See https://github.com/girder/girder/issues/3206
```

girder-dandi-archive/tests/test_rest.py
```diff
@@ -0,0 +1,115 @@
...
+    meta = {
+        "dandiset": {"name": "red", "description": "herring", "identifier": "000000"}
+    }
+    Folder().setMetadata(red_herring_dandiset_000000_folder, meta)
+    return_dict = {
+        "red_herring_collection": red_herring_collection,
+        "red_herring_folder": red_herring_dandiset_000000_folder,
+    }
+    return_dict["drafts_collection"] = get_or_create_drafts_collection()
+    yield return_dict
```
AlmightyYakob: To make the syntax simpler and following what Brian suggested:

```suggestion
    return_dict = {
        "red_herring_collection": red_herring_collection,
        "red_herring_folder": red_herring_dandiset_000000_folder,
        "drafts_collection": get_or_create_drafts_collection(),
    }
    return return_dict
```

mgrauer: I'll address this with #126, I'm doing all of the test restructuring there so that I can get this PR in sooner.

PR #122, opened 2020-03-05
Dlp fixes

### LINKED ISSUE #116: DLP UI links
- Link from Dandiset Landing Page to contents of the dandiset with some UI element
- Link from the contents view to the DLP with some UI element

### LINKED ISSUE #117: DLP tune up
Top block: name, permalink, and identifier. Remove the top right block that currently has the identifier, as we are moving identifier to the top block.

Permalink, for now is only for drafts (née staging) because we don't yet have releases.

e.g. `https://dandiarchive.org/dandiset/000001/draft`

Then middle block is as is (though we may want to update these values from something we calculate, at some point).

Then bottom block shows all other key/values. It can be displayed as it currently is, but just showing all key values of the dandiset-metadata.yaml.

Screen Shot 2020-03-03 at 6 56 24 PM

AlmightyYakob:
Fixes #117.
Fixes #116.

I've added some additional logic with the `fetchFullLocation` function. This is to allow for the button that links from the dandiset contents to the dandiset landing page to function properly, as to appear it needs to know if the folder is a valid dandiset or not.

### DIFF
--- a/web/src/utils.js
+++ b/web/src/utils.js
@@ -1,4 +1,5 @@
 const rootDandiUrl = 'dandiarchive.org';
+const dandiUrl = `https://${rootDandiUrl}`;
 
 function getLocationFromRoute(route) {
   const { _modelType, _id } = route.params;
@@ -51,6 +52,7 @@ function copyToClipboard(text) {
 
 export {
   rootDandiUrl,
+  dandiUrl,
   getLocationFromRoute,
   getPathFromLocation,
   getSelectedFromRoute,
--- a/web/src/utils.js
+++ b/web/src/utils.js
@@ -1,3 +1,5 @@
+const rootDandiUrl = 'dandiarchive.org';
+
 function getLocationFromRoute(route) {
   const { _modelType, _id } = route.params;
   if (_modelType) {
@@ -48,6 +50,10 @@ function copyToClipboard(text) {
 }
 
 export {
-  getLocationFromRoute, getPathFromLocation, getSelectedFromRoute,
-  getPathFromSelected, copyToClipboard,
+  rootDandiUrl,
+  getLocationFromRoute,
+  getPathFromLocation,
+  getSelectedFromRoute,
+  getPathFromSelected,
+  copyToClipboard,
 };
--- a/web/src/utils.js
+++ b/web/src/utils.js
@@ -1,5 +1,4 @@
-const rootDandiUrl = 'dandiarchive.org';
-const dandiUrl = `https://${rootDandiUrl}`;
+const dandiUrl = 'https://dandiarchive.org';
 
 function getLocationFromRoute(route) {
   const { _modelType, _id } = route.params;
@@ -51,7 +50,6 @@ function copyToClipboard(text) {
 }
 
 export {
-  rootDandiUrl,
   dandiUrl,
   getLocationFromRoute,
   getPathFromLocation,
--- a/web/src/views/DandiMetaViewer.vue
+++ b/web/src/views/DandiMetaViewer.vue
@@ -52,7 +52,7 @@
               <v-card-actions class="py-0">
                 <v-btn
                   icon
-                  @click="$router.push(`/collection/${selected.parentId}`)"
+                  :to="`/collection/${selected.parentId}`"
                 >
                   <v-icon>mdi-arrow-left</v-icon>
                 </v-btn>
--- a/web/src/views/DandiMetaViewer.vue
+++ b/web/src/views/DandiMetaViewer.vue
@@ -140,7 +140,7 @@ import VueJsonPretty from 'vue-json-pretty';
 import MetaEditor from '@/components/MetaEditor.vue';
 import SCHEMA from '@/assets/schema/base.json';
 import NEW_SCHEMA from '@/assets/schema/new_dandiset.json';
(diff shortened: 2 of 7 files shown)


web/src/views/DandiMetaViewer.vue
```diff
@@ -29,7 +29,27 @@
...
+                    <v-btn icon>
+                      <v-icon>mdi-content-copy</v-icon>
+                    </v-btn>
+                  </v-list-item-action>
+                  <v-spacer /> -->
+                </v-list-item>
+                <v-list-item>
+                  <v-list-item-content>
+                    <a :href="`https://dandiarchive.org/dandiset/${meta.identifier}/draft`">
+                      {{ `https://dandiarchive.org/dandiset/${meta.identifier}/draft` }}
```
zachmullen: I'm not sure how many places this base URL `https://dandiarchive.org` is hard-coded, but I'd be more comfortable if we used a variable for it.

web/src/views/DandiMetaViewer.vue
```diff
@@ -76,26 +102,22 @@
                     <v-list-item-content>{{item}}</v-list-item-content>
                   </v-list-item>
                 </template>
+                <template v-for="(item, k) in extraFields">
+                  <v-subheader :key="k">{{ k }}</v-subheader>
+                  <v-list-item :key="k">
+                    <v-list-item-content>
+                      <template v-if="['object', 'array'].includes(schema.properties[k].type)">
+                        <vue-json-pretty :data="item" highlightMouseoverNode />
```
zachmullen: Aren't props usually passed as e.g. `highlight-mouseover-node`? Or is my Vue-fu just stale?

web/src/views/DandiMetaViewer.vue
```diff
@@ -42,9 +62,15 @@
                 >
                   <v-icon>mdi-pencil</v-icon>
                 </v-btn>
+                <v-btn
+                  @click="viewContents"
```
zachmullen: In general if a button's purpose is to link to a URL, we should use `:to` rather than `@click`. This allows the user to interact with the link: preview the URL, open it in a new tab, copy the URL, etc.

web/src/views/Home.vue
```diff
@@ -35,24 +47,35 @@ import {
...
 ];
 
 export default {
   components: { GirderDataDetails, GirderFileManager },
   computed: {
+    isDandiset() {
+      return (
+        !!this.location
+        && !!this.location.meta
+        && !!this.location.meta.dandiset
```
zachmullen: To save on both punctuation and time, we can just do

```javascript
!!(this.location && this.location.meta && this.location.meta.dandiset)
```

And hold out hope for the [null propagation operator](https://ponyfoo.com/articles/null-propagation-operator).

web/src/views/Home.vue
```diff
@@ -3,13 +3,20 @@
...
+          :selectable="true"
+          :location.sync="location"
+          :upload-enabled="false"
+          :value="selected"
+          @input="setSelected"
+          :initial-items-per-page="25"
+          :items-per-page-options="[10,25,50,100,-1]"
+        >
+          <template v-slot:headerwidget v-if="isDandiset">
+            <v-btn icon color="primary" @click="viewDandiset">
```
zachmullen: Same here, use `:to` rather than `@click`

web/src/views/Home.vue
AlmightyYakob: Good point. Can't wait to be using the "Elvis" operator.

web/src/views/DandiMetaViewer.vue
AlmightyYakob: Good to know, I'll fix it!

web/src/views/DandiMetaViewer.vue
AlmightyYakob: Nope your Vue-fu is still good

web/src/utils.js
```diff
@@ -1,3 +1,5 @@
+const rootDandiUrl = 'dandiarchive.org';
```
zachmullen: While I'm sure it'll always be https, let's change this to be `https://dandiarchive.org`.

AlmightyYakob: The reason I originally did it this way was to allow for subdomains like `hub.dandiarchive.org` and `girder.dandiarchive.org`, but I've updated it.

zachmullen: If we get subdomains, I'd say those should just be entirely separate variables. This is the sort of thing that I'd normally use `process.env` for so that the application is agnostic (to whatever extent possible) to its deployed environment.

web/src/views/DandiMetaViewer.vue
```diff
@@ -29,7 +29,27 @@
...
+                    <a :href="permalink">
+                      {{ permalink }}
+                    </a>
+                  </v-list-item-content>
+                </v-list-item>
+              </v-list>
+              <v-card-actions class="py-0">
                 <v-btn
                   icon
                   @click="$router.push(`/collection/${selected.parentId}`)"
```
zachmullen: I realize this isn't part of this PR but since we are already doing this, let's go ahead and replace `@click` with `:to` here.

AlmightyYakob:
@zachmullen should be good to go

PR #123, opened 2020-03-06
Meta editor fix

### LINKED ISSUE #104: metadata editor
1. After editing the metadata, clicking on save does not save the metadata
2. The only way to return to the previous screen is to hit cancel
3. Roles should be a drop down list or pillbox with appropriate roles.
4. any orcid id should be verified with the orcid api.
5. right after i entered an author role with an orcid , the publish button came online. at present, till we have implemented validation, publish should be disabled from view.

AlmightyYakob:
Fixes #104, except for distributing the error messages. That requires a bit more work, so I''ll make a separate issue and PR for that.

### DIFF
--- a/web/src/assets/schema/base.json
+++ b/web/src/assets/schema/base.json
@@ -14,7 +14,8 @@
       "$id": "#/properties/identifier",
       "type": "string",
       "title": "Identifier",
-      "minLength": 1
+      "minLength": 1,
+      "readOnly": true
     },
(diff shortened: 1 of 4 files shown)


AlmightyYakob:
> One thing I noticed while playing with this was that it's possible to add contributors through the GUI, but no way to remove them (that I could find).

Yes, this is an issue that I still have to address.

PR #128, opened 2020-03-09
Use lockfile, yarn, --frozen-lockfile

subdavis:
Noticed a discussion on this in another thread.

It's best not to mix yarn and npm.  You should also use `--frozen-lockfile` or `--pure-lockfile` when installing into a docker image.  https://github.com/yarnpkg/yarn/issues/4147

### DIFF
--- a/docker/client.Dockerfile
+++ b/docker/client.Dockerfile
@@ -1,8 +1,9 @@
 FROM node
 
 COPY web/package.json /client/package.json
+COPY web/yarn.lock /client/yarn.lock
 WORKDIR /client
-RUN npm install
+RUN yarn install --frozen-lockfile
 COPY web /client
 RUN yarn run build --mode docker
 


docker/client.Dockerfile
```diff
@@ -1,11 +1,12 @@
 FROM node
 
 COPY web/package.json /client/package.json
+COPY web/yarn.lock /client/yarn.lock
```
AlmightyYakob: I know this isn't what the PR is addressing, but I still don't see why the build is being done this way. @satra said it was to prevent running yarn install repeatedly, however when I run `docker build` locally, unless the actual dependencies change (which requires a yarn install anyway), docker caches that step.

subdavis: Not sure if this differs from your experience, but here's what should happen:

1) when lockfile or package.json change, a full docker build happens.
2) when only code changes, docker should cache the yarn install

You're saying this was the behavior even before @satra's change?

satra:
the reason i was doing to separate steps was so that i could build things with limited bandwidth.

in the original version the first COPY statement copied the source directory over. since i was developing, the source directory was always dirty. this would lead to npm install always running and repulling the repos. 

the separation allowed npm to install only when package.json changed. allowing a faster rebuild.  on a flight that was much quicker :)

otherwise the PR LGTM.

subdavis:
^ that's what I would have expected too.

AlmightyYakob:
https://github.com/dandi/dandiarchive/blob/fa647bd44081eab517f8012d473de6e473586beb/docker/client.Dockerfile#L3-L4

Locally I replaced the above two lines with the following, which is run anyway on line 7:

https://github.com/dandi/dandiarchive/blob/fa647bd44081eab517f8012d473de6e473586beb/docker/client.Dockerfile#L7

And observed the desired behavior you described @satra, when editing source code. It cached the install unless `package.json` was changed, at which point `yarn install` was actually run.

subdavis:
That's really interesting.  With the changes above, any source code changes cause a fresh NPM install.  This is what I would expect to happen because the hash of the previous layer has changed.

This issue doesn't really matter, except that it's a mystery and I don't like mysteries.

PR #129, opened 2020-03-10
fix: add clickable dandi icon for home navigation

satra:


satra:
thanks @AlmightyYakob - i've made a quick change to go back to the root location

PR #134, opened 2020-03-11
getting API key and display above login

JellyWang-3:


mgrauer:
@subdavis would you mind QA-ing this?

@JellyWang-3 This is great progress! There are merge conflicts that need to be resolved though. Also I noticed that this displays any API key that the user has, but instead it should specifically look for the key that is named `dandicli` and display that one, or create one with that name if needed.

mgrauer:
You can see how this displays, it would be nice (if it is easy to do) to somehow add in text that says "dandicli api key", perhaps above the api key display itself, or it could be prepended as a string, e.g. "dandicli api key: <keystring>"? Otherwise it is a somewhat mysterious string.

Screen Shot 2020-03-11 at 10 26 33 PM

subdavis:
Some ideas for text that is intended to copy:

* It's always nice to put a copy button on it to that puts it on the clipboard.  
* If you put in in a textarea or text input, the user can use ctrl+a to select the whole thing.

Here's an example in another app:

Screenshot from 2020-03-12 08-57-38

web/src/components/DandiAppBar.vue
```diff
@@ -81,13 +93,38 @@
...
+        if (data[0]._modelType === 'api_key') {
+          // set the user key
+          this.user.key = data[0].key;
+          return this.user.key;
+        }
+      } else {
+        // create a key using "POST" endpoint
+        const { status2, data2 } = await this.girderRest.post(
+          '/api_key?name=dandicli&scope=%20%5B%22core.data.read%22%2C%20%20%22core.data.write%22%5D&tokenDuration=30&active=true',
+        );
```
subdavis: It's considered a best practice to factor your api queries out to either your vuex module or even an api wrapper class.  Since you're already using Vuex, you may want to factor this whole method out into something like a `fetchApiKey` action.

web/src/components/DandiAppBar.vue
```diff
@@ -81,13 +93,38 @@
...
+        if (status2 === 200) {
+          this.user.key = data2.key;
+          return this.user.key;
+        }
+      }
+    },
+  },
   computed: {
     ...mapGetters(['loggedIn', 'user']),
+    ...mapState(['girderRest']),
```
subdavis: Then you could mapActions to get your API Key and not import girderRest like this.

One of the main benefits is that when you eventually change to using a different backend, this component requires a lot less work to refactor.

web/src/components/DandiAppBar.vue
```diff
@@ -105,6 +142,20 @@ export default {
...
+        `/api_key?userId=${this.user._id}&limit=50&sort=name&sortdir=1`,
+      );
+      if (status === 200 && data[0]) {
+        // send the key id to "PUT" endpoint for updating
+        this.user.keyId = data[0]._id;
+        const updated = await this.girderRest.put(`api_key/${this.user.keyId}`);
+        this.user.key = updated.data.key;
+        return this.user.key;
+      }
+    },
```
subdavis: Ditto this method: probably refactor it to an action.

web/src/components/DandiAppBar.vue
```diff
@@ -59,12 +59,24 @@
...
       <v-list dense>
+        <v-list-item
+          @click="reloadAPIKey"
+        >
+          <v-list-item-action class="mr-2">
+            <v-btn icon>
+              <v-icon>mdi-reload</v-icon>
+            </v-btn>
+          </v-list-item-action>
+          <v-list-item-title>{{ apiKey }}</v-list-item-title>
```
subdavis: See my top-level comment about copying text.   Maybe there's some part of that you can fit in here, maybe not.

JellyWang-3:
Thanks for the suggestions and I will work on them.

JellyWang-3:
@subdavis Thanks for the great learning materials and suggestions. I am still learning about Vue and having some resources to look at is awesome.

mgrauer:
@AlmightyYakob @JellyWang-3 

Is this in a good state to get QA-ed/merged? Could one of you resolve the merge conflicts?

AlmightyYakob:
> @AlmightyYakob @JellyWang-3
> 
> Is this in a good state to get QA-ed/merged? Could one of you resolve the merge conflicts?

Yes, I believe this addresses everything mentioned above, besides the comment Brandon made about displaying the api key. I think that's something that could be fixed in a separate PR, but if you feel that it should be done here that works too. I can resolve the merge conflicts.

mgrauer:
Let’s do the copy fix later.

Jake: please resolve conflicts
Jelly: please create a follow up issue
Brandon: please QA

web/src/store.js
```diff
@@ -25,6 +29,61 @@ export default new Vuex.Store({
...
+            tokenDuration: 30,
+            active: true,
+          },
+        });
+
+        if (createStatus === 200) {
+          commit('setApiKey', key);
+        }
+      }
+    },
```
subdavis: It seems to me like these 2 actions overlap a lot could be combined into a  single "fetchApiKey" action that either gets or creates.

Maybe something to look at some other time.

web/src/store.js
```diff
@@ -25,6 +29,61 @@ export default new Vuex.Store({
...
+          params: {
+            userId: user._id,
+            limit: 50,
+            sort: 'name',
+            sortdir: 1,
+          },
+        },
+      );
+
+      const [dandiKey] = data.filter(key => key.name === 'dandicli');
```
subdavis: In the case of 4XX and 5XX errors, this line will probably error before the status error check on like 46.

Data will be undefined or null, and this array destructure will throw `TypeError: data is not iterable`.

You probably want to move like 45 into the if block.  Ideally, you also want some error handling to display to the user.   You don't have to do it right now, but it might be good to stick a TODO nearby to do error handling.

web/src/store.js
```diff
@@ -25,6 +29,61 @@ export default new Vuex.Store({
...
+      const { status, data } = await state.girderRest.get(
+        'api_key', {
+          params: {
+            userId: user._id,
+            limit: 50,
+            sort: 'name',
+            sortdir: 1,
+          },
+        },
+      );
```
subdavis: Nice use of axios params.

PR #143, opened 2020-03-13
copying API key WIP

JellyWang-3:
Copying not working and errors in console.


AlmightyYakob:
@mgrauer this should be good now

mgrauer:
Great, thanks @JellyWang-3 and @AlmightyYakob . I tested this in preview and it works.

Issue #152, opened 2020-03-15
Generated API key should have "Full Scope" to be able to "register" a new dandiset


yarikoptic:
See https://github.com/dandi/dandi-cli/issues/51

mgrauer:
To achieve this, pass in null for the `scope` param when [creating the api key](https://github.com/dandi/dandiarchive/blob/master/web/src/store.js#L76).

PR #153, opened 2020-03-16
Create dandicli apikey with full scope

mgrauer:


mgrauer:
Fixes #152.

mgrauer:
This is a small change, and I have confirmed that it works via the preview. Before testing this PR, I had deleted my `dandicli` apikey. When viewing my apikey in the preview gui, there was an apikey displayed, and when I looked at that `dandicli` apikey in the girder gui, it had full scope.

PR #159, opened 2020-03-16
Load Sentry in development

brianhelba:
If a DSN (`process.env.VUE_APP_SENTRY_DSN`) is not set, Sentry will gracefully no-op, and errors will not be sent to Sentry.

By always loading Sentry, it ensures the development environment is closer to production.

~~Also, for more useful diagnosis, enable the `attachProps` option to the Sentry + Vue integration: https://docs.sentry.io/platforms/javascript/vue/~~

### DIFF
--- a/web/src/main.js
+++ b/web/src/main.js
@@ -12,13 +12,7 @@ Vue.use(Girder);
 
 Sentry.init({
   dsn: process.env.VUE_APP_SENTRY_DSN,
-  integrations: [
-    new Integrations.Vue({ 
-      Vue, 
-      attachProps: true,
-      logErrors: true 
-    }),
-  ],
+  integrations: [new Integrations.Vue({ Vue, logErrors: true })],
 });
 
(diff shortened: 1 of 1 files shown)


web/src/main.js
```diff
@@ -10,12 +10,16 @@ import store from './store';
...
-    dsn: process.env.VUE_APP_SENTRY_DSN,
-    integrations: [new Integrations.Vue({ Vue, logErrors: true })],
-  });
-}
+Sentry.init({
+  dsn: process.env.VUE_APP_SENTRY_DSN,
+  integrations: [
+    new Integrations.Vue({ 
+      Vue, 
+      attachProps: true,
```
AlmightyYakob: FWIW, I think this defaults to `true` if not provided, but I could see wanting it to be explicit.

brianhelba: Yup, you're right, I misread the docs. I can remove, to keep it minimal.

Issue #133, opened 2020-03-11
Start with 000001 not 000000


yarikoptic:
ATM the first dataset  we create via our API would have an identifier 000000.  I think it would be better it to be 000001.

Tried via http://localhost:8091/api/v1#!/dandi/dandi_create_dandiset to receive a

<details>
<summary>response</summary> 

```json
{
  "_id": "5e68e961513b0757236c1886",
  "access": {
    "groups": [],
    "users": [
      {
        "flags": [],
        "id": "5e68df9c513b0757236c1399",
        "level": 2
      }
    ]
  },
  "baseParentId": "5e68df9c513b0757236c139e",
  "baseParentType": "collection",
  "created": "2020-03-11T13:36:33.077066+00:00",
  "creatorId": "5e68df9c513b0757236c1399",
  "description": "",
  "lowerName": "000000",
  "meta": {
    "dandiset": {
      "description": "long description",
      "identifier": "000000",
      "name": "test1"
    }
  },
  "name": "000000",
  "parentCollection": "collection",
  "parentId": "5e68df9c513b0757236c139e",
  "public": true,
  "size": 0,
  "updated": "2020-03-11T13:36:33.079169+00:00"
}
```
</details>

Issue #126, opened 2020-03-08
Restructure tests for create and get dandiset endpoints


mgrauer:
I won't have time right away to restructure the tests in response to QA for #120, so I'm breaking that restructuring out into a separate issue. This way we can get the endpoint changes in sooner without this additional work blocking, as we are pushing to have all of the beta features come together soon.

brianhelba:
@mgrauer I asked @dchiquit to work on this now, since a lot of his other WIP features will need to add tests, and new fixtures / organization will benefit those tests too.

mgrauer:
That makes me sad because I was hoping to do it someday, but this is the right choice. Thanks.

PR #162, opened 2020-03-17
Start dandiset ID counter at 1

dchiquit:
The dandiset was originally set to start at 0 when the first collection
was registered. Change that default to 1.

The name of the DANDI plugin class was `GirderPlugin` for some reason,
which conflicts with the original `GirderPlugin` superclass. Change
that to `DandiArchivePlugin` and update the `entry_point` in setup.py.

I tested this by stopping Mongo, deleting everything from the DB directory, restarting Mongo, starting Girder, then registering a dataset.

After changing the plugin name, I had to rerun `pip install -e .` to reregister setup.py so the plugin would launch.

### DIFF
--- a/girder-dandi-archive/tests/test_rest.py
+++ b/girder-dandi-archive/tests/test_rest.py
@@ -17,19 +17,19 @@ def draftsFolders(db):
         "red_herring_collection", reuseExisting=True
     )
     # A folder that matches dandiset metadata, but not in drafts collection.
-    red_herring_dandiset_000000_folder = Folder().createFolder(
+    red_herring_dandiset_000001_folder = Folder().createFolder(
         parent=red_herring_collection,
         parentType="collection",
-        name="000000",
+        name="000001",
         public=True,
     )
     meta = {
-        "dandiset": {"name": "red", "description": "herring", "identifier": "000000"}
+        "dandiset": {"name": "red", "description": "herring", "identifier": "000001"}
     }
-    Folder().setMetadata(red_herring_dandiset_000000_folder, meta)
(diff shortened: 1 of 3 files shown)


dchiquit:
Fixes #133

girder-dandi-archive/tests/test_rest.py
```diff
@@ -44,38 +44,38 @@ def testCreateDandiset(server, draftsFolders, user):
...
+    test_dandiset_1_folder = resp.json
+    test_dandiset_1_folder_meta = test_dandiset_1_folder["meta"]["dandiset"]
 
-    assert drafts_collection_id == test_dandiset_0_folder["parentId"]
-    assert "000000" == test_dandiset_0_folder["name"]
-    assert "000000" == test_dandiset_0_folder_meta["identifier"]
-    assert "test dandiset 0 name" == test_dandiset_0_folder_meta["name"]
-    assert "test dandiset 0 description" == test_dandiset_0_folder_meta["description"]
+    assert drafts_collection_id == test_dandiset_1_folder["parentId"]
+    assert "000001" == test_dandiset_1_folder["name"]
```
danlamanna: What are your thoughts on having a test that just tests the counter behavior?

dchiquit: I'd say that it's already covered by `testCreateDandiset`. Counter behavior is a part of the dandiset creation process. A counter specific test would be exactly the same as `testCreateDandiset`, but with only the ID assert.

mgrauer:
As @dchiquit will work on #126, that is an opportunity to think about restructuring or adding tests as @danlamanna suggested.

PR #170, opened 2020-03-19
#126 Restructure test_rest.py tests and add test utilities

### LINKED ISSUE #126: Restructure tests for create and get dandiset endpoints

dchiquit:
Restructure the slightly malformed test_rest.py tests into smaller, more
singleminded tests. Add new tests to cover negative code paths.

To make asserting large, complex JSON responses easier, I added a
utility, `substitute.py`, which enables "vague" assertions about objects
that do not have explicitly expected values, but do have some implicit
state. An instance of the `Timestamp` class is used to represent any
reasonably well-formatted timestamp in a JSON response. I do not think
that this utility belongs as is, but I want to generate some discussion
on it.

I've included a test file to demonstrate how `substitute.py` works. I also included both the original asserts and the new, substituty asserts in `test_rest.py` so the two styles can be compared. Any thoughts are welcome.

Fixes #126

### DIFF
--- a/girder-dandi-archive/tests/test_rest.py
+++ b/girder-dandi-archive/tests/test_rest.py
@@ -7,12 +7,16 @@ from girder.models.folder import Folder
 
 from pytest_girder.assertions import assertStatus, assertStatusOk
 
+pytestmark = pytest.mark.plugin("dandi_archive")
 
-# TODO: This fixture and these tests are not well structured and hard to maintain.
-# They currently work correctly in terms of the application code that they test,
-# but they should be rewritten and not followed as an example.
+NAME_1 = "test dandiset 1 name"
+DESCRIPTION_1 = "test dandiset 1 description"
+NAME_2 = "test dandiset 2 name"
+DESCRIPTION_2 = "test dandiset 2 description"
+
+
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


mgrauer:
@danlamanna @brianhelba could one of y'all QA?

dchiquit:
I removed the substitute code and hosted it at https://github.com/dchiquit/pytest-assert-helper until we decide what to do with it. While setting up that repository I found https://github.com/PerchSecurity/pytest-assert-utils which does some of the same things, so maybe we don't even need it.
The remaining code is the originally mandated test cleanup.

dchiquit:
@brianhelba Good ideas, I'm sneaking them into https://github.com/dandi/dandiarchive/pull/174

brianhelba:
@dchiquit If you don't mind, it'd be easier to review if you could make a separate branch and PR.

dchiquit:
Added https://github.com/dandi/dandiarchive/pull/178 to implement those changes.

PR #172, opened 2020-03-20
Add meta.dandiset.identifier index

### LINKED ISSUE #96: Add indexing to dandiset ID find
Raised by

https://github.com/dandi/dandiarchive/pull/94/files/b30fd0645be52c1010b3a08bc258e8e50cfc9693#r385823584

dchiquit:
Fixes #96 

Add a mongo index on `meta.dandiset.identifier`. When I did not include `parentId` in the index, Mongo decided to use one of the `parentId` indices instead, which is not useful. I opted for `meta.dandiset.identifier` before `parentId` so the index is still useful if we ever have an endpoint that does not specify the `parentId`.

Test steps:
1. Wipe database
2. Start server and set up user
3. Verify that the index is present
4. Create a new dandiset
5. GET /dandi with `identifier:"000001"`
6. GET /dand with `identifier:"000002"`
7. Verify that the index was used twice (`db.folder.aggregate( [ { $indexStats: { } } ] )`)


mgrauer:
I'm not following what you said here:

>When I did not include parentId in the index, Mongo decided to use one of the parentId indices instead, which is not useful. I opted for meta.dandiset.identifier before parentId so the index is still useful if we ever have an endpoint that does not specify the parentId.

There should already be a `parentId` index on folders:

https://github.com/girder/girder/blob/master/girder/models/folder.py#L28-L29

dchiquit:
There is an existing `parentId` index. For context, this is the query that `rest.py` is performing:
```
doc = Folder().findOne(
            {"parentId": drafts["_id"], "meta.dandiset.identifier": identifier}
        )
```
From my experimentation, this is how the query behaves:
* No additional index: the existing `parentId` index is used. This is not useful.
* `{"meta.dandiset.identifier":1}`: there are two usable indices, so Mongo chooses the `parentId` index for no apparent reason.
* `{"meta.dandiset.identifier":1, "parentId":1}`: this index perfectly matches the query, so Mongo will always use it.

When setting up a compound index, order matters. The index `{"meta.dandiset.identifier":1, "parentId":1}` also gives you the index `{"meta.dandiset.identifier":1}` for free. I picked that order because we already have an index on `parentId`.

mgrauer:
>compound index

Got it, that was the key term I was missing before.

girder-dandi-archive/girder_dandi_archive/__init__.py
```diff
@@ -19,6 +20,7 @@ def load(self, info):
             {"$setOnInsert": {"value": 1}},
             upsert=True,
         )
+        Folder().ensureIndex(([("meta.dandiset.identifier", 1), ("parentId", 1)], {}))
```
brianhelba:
```suggestion
        Folder().ensureIndex([("meta.dandiset.identifier", 1), ("parentId", 1)])
```
Since there are no options, there's no need to pass a tuple: https://github.com/girder/girder/blob/355e85b40bc86e37147a702aabf8338e2a13f9bb/girder/models/model_base.py#L196

dchiquit: That code seems to make the assumption that all lists/tuples are of the form `(index, options)`, and all strings are simply the name of the index. I need to create a compound index, which has to be encoded as a list/tuple. That code will misinterpret my compound index and complain that the options argument is malformed. The only way I found to avoid the misunderstanding is by specifying the empty options.

If I were doing `Folder().ensureIndex(("meta.dandiset.identifier", {}))`, you would be absolutely right.

brianhelba: Ah, I see. You're right, thanks.

girder-dandi-archive/girder_dandi_archive/__init__.py
```diff
@@ -19,6 +20,9 @@ def load(self, info):
             {"$setOnInsert": {"value": 1}},
             upsert=True,
         )
+        # Index for rest.DandiResource.get_dandiset and other endpoints that might
+        # require an index on meta.dandiset.identifier
```
brianhelba:
```suggestion
        # Index used by rest.DandiResource.get_dandiset and others.
        # This compound index is necessary, otherwise MongoDB seems to use only the
        # parentId index instead of a plain meta.dandiset.identifier index, given a
        # query for both fields.
```

I think it's most important that we communicate the *why* instead of the *what* in these sort of code comments. Any competent developer can figure out that this is an index and it's used for queries on the `meta.dandiset.identifier` field just by inspection of the code itself. Harder questions to answer from the code itself are:
1. *"Is this fragment still used? Where?"* - This could be answered by a global search, but given that the code for a dandiset query is split among two separate places (this index initialization and the query itself), this is a helpful (but not essential) comment.
2. *"Why is this a compound index? There's already an index on `parentId`, so I'll just simplify this index again..."* - This is definitely not obvious from looking at the code itself. We need to preserve the in-depth analysis that went into this programming decision, so others don't break it in the future.

dchiquit: I rewrote my comment and tried to answer the questions you posed.

Issue #103, opened 2020-03-01
schema updates


satra:
here is the biccn discussion on the schema, which should update our schema. i.e. we will need a path to update dandiset.yaml as the schema changes. so dandiset.yaml should also have schema-version in it.

https://github.com/BICCN/BICCN-Infrastructure-Draft/pull/15/files

@AlmightyYakob - please see the comment on the roles section to add a list of roles to the schema.

please add a README here: https://github.com/dandi/dandiarchive/tree/master/web/src/assets/schema 
so we know which schemas are being used for what. if some schemas are not being used, let's delete them. they can always be recovered from git history.

@yarikoptic has created a project for dandi-core that can be used as a submodule to dandiarchive. we can move common things there so that they can be used by multiple projects. schemas seem like a good thing to move there.

satra:
@AlmightyYakob - i don't think this was communicated after @mgrauer and i edited this file:

https://github.com/BICCN/BICCN-Infrastructure-Draft/blob/ce0df09b310bbcdd0a7a63cc51a0955e3cae97b8/BICCN-policies/dataset-metadata.yaml

the schema should be updated to reflect the keys/options from this file. the file is visually divided into 3 sections. the first section should all be available as part of the human metadata entry through the DLP

AlmightyYakob:
@satra in reference to this line:

https://github.com/BICCN/BICCN-Infrastructure-Draft/blob/ce0df09b310bbcdd0a7a63cc51a0955e3cae97b8/BICCN-policies/dataset-metadata.yaml#L6

We seem to have everything except "datacite", is that correct? I.e. should I add the role "Datacite" to the list of allowed roles?

satra:
@AlmightyYakob - i think @mgrauer sent you that list from the google doc, which we also curated during that call. so that should be all set.

satra:
from slack (cc: @yarikoptic @mgrauer):

- let's add an annotation key/field as a list of notes.
- each note should be a text-area field

ideally each note should be a structured key:

```
- annotation:
   - description: (text area to provide information)
      user: (logged in user)
      timestamp: timestamp of note
   - description:
...
```

cc:

AlmightyYakob:
> ```
> - annotation:
>    - description: (text area to provide information)
>       user: (logged in user)
>       timestamp: timestamp of note
>    - description:
> ...
> ```

So the user and timestamp fields would be autogenerated?

satra:
[...]

PR #173, opened 2020-03-20
Update schema

AlmightyYakob:
In response to #103.

Hopefully this covers most of it, but there's sometimes some ambiguity in the [definition](https://github.com/BICCN/BICCN-Infrastructure-Draft/blob/ce0df09b310bbcdd0a7a63cc51a0955e3cae97b8/BICCN-policies/dataset-metadata.yaml)

Please let me know if anything is missing/incorrect


web/src/assets/schema/nwb.json
```diff
@@ -0,0 +1,109 @@
...
+        "categorical": {
+          "$id": "#/properties/age/properties/categorical",
+          "type": "object",
+          "title": "Categorical"
+        }
+      }
+    },
+    "sex": {
+      "$id": "#/properties/sex",
+      "type": "string",
```
satra: this should be an array of strings

web/src/assets/schema/base.json
```diff
@@ -284,7 +258,7 @@
         }
       }
     },
-    "ethics_approvals": {
+    "ethicsApprovals": {
```
satra: should be an array of these objects.

web/src/assets/schema/base.json
```diff
@@ -376,6 +350,69 @@
       "$id": "#/properties/language",
       "type": "string",
       "title": "Language"
+    },
+    "associatedData": {
+      "$id": "#/properties/associatedData",
+      "type": "object",
```
satra: also an array of objects.

web/src/assets/schema/base.json
```diff
@@ -108,33 +127,6 @@
...
-        "affiliations": {
-          "$id": "#/properties/contact/properties/affiliations",
-          "type": "string",
-          "title": "Affiliations"
-        }
-      }
-    },
     "sponsors": {
       "$id": "#/properties/sponsors",
       "type": "object",
```
satra: sponsors should also be an array

satra:
@AlmightyYakob - this looks good. i left some comments on things that should be arrays of objects. if encoded a specific object versus an array in the yaml using the list syntax. so wherever there is a `-` it means that key should be in an array.

for example:
```
- key1:
  - key2:
     key3:
```
so `key1` is an array of objects with keys `key2` and `key3` 

the one other thing is to add to `sponsors` section an optional field called `Award Number` this would be a string.

AlmightyYakob:
@satra thanks for the feedback, I'm fixing that now. It also seems like these properties should be arrays of objects as well: 

https://github.com/BICCN/BICCN-Infrastructure-Draft/blob/ce0df09b310bbcdd0a7a63cc51a0955e3cae97b8/BICCN-policies/dataset-metadata.yaml#L16-L24

Since they're formatted the same way as the ones you just mentioned. Is that correct?

satra:
> Since they're formatted the same way as the ones you just mentioned. Is that correct?

yes

i just noticed i made a mistake with respect to that syntax for age in the nwb section. that should not be a list but either an object. fixed  here: https://github.com/BICCN/BICCN-Infrastructure-Draft/blob/3f07606f7a07584d35e6ff097ff9ecda17f5b7fc/BICCN-policies/dataset-metadata.yaml

AlmightyYakob:
@satra I made changes, hopefully now it's accurate

satra:
@AlmightyYakob - looks good now.

web/src/assets/schema/base.json
```diff
@@ -108,92 +127,52 @@
...
```
satra: sponsors should be an array.

also we should a string field to the object called `Award number`

AlmightyYakob: > sponsors should be an array.

Oops, forgot to actually update the type itself. 

> also we should a string field to the object called `Award number`

To the sponsors object (the objects contained within the sponsors array)?

satra: yes, another field to the sponsors object:

```
- sponsors:
   - name:
     ...
     award_number:
```

AlmightyYakob:
Should be good to go

satra:
This came up in the meeting today and had skipped my mind. Can we change "Associated Disease" to "Conditions studied"?

PR #174, opened 2020-03-20
Add list dandisets endpoint: GET /dandi/list

### LINKED ISSUE #144: Create an endpoint to return a list of dandisets
This would be needed to support #93 .

At the current time, this should be a list of folders in the `drafts` collection.

dchiquit:
Add list dandisets endpoint and associated tests. The endpoint is `GET /dandi/list`.

I added one tests that thoroughly validates the response when there are two dandisets, and one test for each of the paging options (sort, limit, and offset) just to ensure they're plugged in properly.

Fixes #144


girder-dandi-archive/tests/test_rest.py
```diff
@@ -225,3 +225,68 @@ def test_get_dandiset_invalid_identifier(server, drafts_folders, user, dandiset_
...
     )
     assertStatus(resp, 400)
+
+
+def test_list_dandisets(server, user, dandiset_1, dandiset_2):
+
+    resp = server.request(path="/dandi/list", method="GET", user=user)
+    assertStatusOk(resp)
+    assert len(resp.json) == 2
+    assert resp.json[0]["access"] == dandiset_1["access"]
```
mgrauer: So many asserts! If only there was a better way, as this is not readily digestible by a human. To make my complaint more concrete, there are a lot of properties to be tested that need to be individually tracked by someone who wants to maintain this test, and also there is a lot of visual noise, i.e., this line

> assert resp.json[0]["lowerName"] == dandiset_1["lowerName"]

is a lot of characters, and the key information is as compact as `lowerName`.


This isn't a criticism of your tests, more like a cry for help towards a more scalable approach. @brianhelba told me that he had already discussed some ideas around this with you, and I support investigations into this. That doesn't mean this PR needs to be held up waiting for a hero.

yarikoptic: all the fields ideally should come from some schema, and then there could be a loop which would have collected all the fields which differ, and eventually reported what differs (so seen at once, instead of rerunning upon each individual fix to discover the next one).  so I would have done something like
```python
assert len(resp.json) == 1
response = resp.json[0]
differences = {
    f: (response.get(f, None), dandiset_1.get(f, None)) 
    for f in someschema
    if response.get(f, None) != dandiset_1.get(f, None)
}
assert not differences
```

dchiquit: I added a helper method for comparing two dandisets that should make reading the tests much more painless.
While a comparison loop would work, I've learned to be wary of adding too much abstraction to tests in the name of making them shorter. When a test fails, the investigator needs to read the test failure, find the code that generated it, parse the loop, interpret the failure message according to how it is formatted here, and possibly find the original `someschema` to compare it. This is totally doable, but more difficult than just reading an assert statement.

mgrauer: absolutely agree. at the same time that i want to reduce visual noise and reader fatigue, it is valuable for tests to be explicit and test code to be simple for humans (especially those humans that aren't the original developer of the test/test harness code, as these will be the people likely to be looking at broken tests) to comprehend and debug.

girder-dandi-archive/tests/test_rest.py
```diff
@@ -10,9 +10,9 @@
 pytestmark = pytest.mark.plugin("dandi_archive")
 
 NAME_1 = "test dandiset 1 name"
-DESCRIPTION_1 = "test dandiset 1 description"
+DESCRIPTION_1 = "Zzzz!"
 NAME_2 = "test dandiset 2 name"
-DESCRIPTION_2 = "test dandiset 2 description"
+DESCRIPTION_2 = "Aaaa!"
```
brianhelba:
```suggestion
DESCRIPTION_2 = "Aaaa! This sorts first."
```

girder-dandi-archive/girder_dandi_archive/rest.py
```diff
@@ -83,3 +84,17 @@ def get_dandiset(self, params):
...
+    @autoDescribeRoute(
+        Description("List Dandisets").pagingParams(
+            defaultSort="meta.dandiset.identifier"
+        )
+    )
+    def list_dandisets(self, limit, offset, sort):
+
+        # Ensure we are only looking for drafts collection child folders.
+        drafts = get_or_create_drafts_collection()
+        return Folder().find(
```
brianhelba: There's no access control here, but I'll file an issue.

PR #175, opened 2020-03-20
Add schema to DLP and make DLP responsive

### LINKED ISSUE #163: Show information from both sections in the DLP landing page
The DLP landing page should show keys from both sections 1 (human entered) and sections 3 (from dandi organize) of the schema.

This should also show sections that are empty but required with an alert icon, so that dataset creators can see this.

AlmightyYakob:
Fixes #163.

Should be merged after #173, as it's branched off from that. Until that PR is merged, the diff here will look large, because it contains those too. The only relevant changes here are in `DandiMetaViewer.vue`.

This also makes the DLP more responsive, so it should look better on smaller pages.

### DIFF
--- a/web/src/views/DandiMetaViewer.vue
+++ b/web/src/views/DandiMetaViewer.vue
@@ -138,9 +138,11 @@ import { mapState, mapGetters } from 'vuex';
 import VueJsonPretty from 'vue-json-pretty';
 
 import MetaEditor from '@/components/MetaEditor.vue';
+import { dandiUrl } from '@/utils';
+
 import SCHEMA from '@/assets/schema/base.json';
 import NEW_SCHEMA from '@/assets/schema/new_dandiset.json';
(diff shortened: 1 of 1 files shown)


mgrauer:
@satra can you QA this from the perspective of looking at the UI, and does it show what you want?

e.g., https://deploy-preview-175--gui-dandiarchive.netlify.com/#/dandiset-meta/5e6ff4140fe0ec33e6b3d5d2

We can skip code review for now because we are close to rewriting all this.

satra:
@mgrauer - this addresses the primary thing, which is to list the sections that have information. however, showing json in sections is quite suboptimal from a presentation standpoint, so hopefully there is an intermediate state between this and the full ui redesign.

i have some things to finish by sunday, but i can take a crack at it after i do.

so i'll merge this in so things are aligned and @bendichter  can start editing some of the metadata.

PR #177, opened 2020-03-23
Reduce dandicli API key scope

### LINKED ISSUE #156: reduce dandicli apikey scopes
Reduce the dandicli apikey scopes to the minimum scope needed.

Basically revert https://github.com/dandi/dandiarchive/pull/153.

In order to do that, we need to change the [create dandiset endpoint](https://github.com/dandi/dandiarchive/blob/master/girder-dandi-archive/girder_dandi_archive/rest.py#L30) to have token write scope rather than user scope, such as this [example](https://github.com/girder/girder/blob/master/girder/api/v1/folder.py#L131).

dchiquit:
The dandicli API key currently has the default scope, which is
everything. It should use the minimum scope required, `core.data.read
core.data.write`.

This also involves specifying the correct scope in the
create_dandiset endpoint. All other endpoints are currently public.

I tested the new, reduced scope `dandicli` API key in the web UI and in the python CLI, and I was able to create new dandisets in both.

I did not add any regression tests because the `dandicli` API key isn't defined on the server side. It is instead created as necessary by the web UI. If we ever decide to initialize that API key on the server then it would definitely be appropriate to find a way to parameterize `test_rest.py` to use both the user and the `dandicli` token.

Fixes #156

### DIFF
--- a/girder-dandi-archive/tests/test_rest.py
+++ b/girder-dandi-archive/tests/test_rest.py
@@ -4,6 +4,7 @@ from girder_dandi_archive.util import get_or_create_drafts_collection
 
 from girder.models.collection import Collection
 from girder.models.folder import Folder
+from girder.models.token import Token, TokenScope
 
 from pytest_girder.assertions import assertStatus, assertStatusOk
 
@@ -16,6 +17,21 @@ DESCRIPTION_2 = "Aaaa! This sorts first."
 
 
 @pytest.fixture
+def read_write_token(user):
+    return Token().createToken(
+        user, scope=[TokenScope.DATA_READ, TokenScope.DATA_WRITE]
+    )
+
+
+@pytest.fixture(params=["user", "read_write_token"])
+def request_auth(request, user, read_write_token):
+    if request.param == "user":
+        return {"user": user}
+    elif request.param == "read_write_token":
+        return {"token": read_write_token}
+
+
+@pytest.fixture(autouse=True)
 def drafts_collection(db):
     red_herring_collection = Collection().createCollection(
         "red_herring_collection", reuseExisting=True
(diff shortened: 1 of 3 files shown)


brianhelba:
Sorry, I don't fully understand your point about testing. I agree it would be ideal to parameterize each our tests that make API calls, to do so with both user and token based auth.

However, even without that, is it possible to make a single test like `test_create_dandiset_token` where the test case itself creates a token (probably within a fixture), then runs the operation?

Also I'd be happy to discuss solutions for your idea about parameterization, as I think it will be a very worthwhile improvement for both current and future tests.

dchiquit:
My thinking was that we could parameterize the tests right now, but we'd have to specify in the test that the token is a `['core.data.read','core.data.write']` token. That token is totally nonexistent in the server, it is created and consumed strictly by the web UI. That decoupling didn't feel right to me. It would be easy to modify the token in the UI (just like Michael did) and neglect to update this specific test. I guess some documentation could fix that, but it feels clumsy. Testing this case is useful for future modifications to the token, but testing it in this way seems like it would make those future modifications more difficult.
I could be totally wrong, this is just my gut feeling.

girder-dandi-archive/tests/test_rest.py
```diff
@@ -36,24 +52,24 @@ def drafts_collection(db):
 
 
 @pytest.fixture
-def dandiset_1(server, drafts_collection, user):
```
brianhelba: Even if `drafts_collection` isn't being explicitly referenced, do we want to pre-create it (and the red herrings) as a precondition of this test case?

dchiquit: It is used automatically on all tests in this file. See line 34: https://github.com/dandi/dandiarchive/pull/177/files#diff-1275a7650980590079021de557e6bfe4R34
```
@pytest.fixture(autouse=True)
def drafts_collection(db):
```

brianhelba: Oh, I see, great.

Issue #189, opened 2020-03-23
Remove "Delete" action in item details


AlmightyYakob:
Relevant section, which appears when selecting a dandiset folder.

![Screenshot from 2020-03-23 17-18-46](https://user-images.githubusercontent.com/11370025/77364212-752a9380-6d2a-11ea-828f-aa6af06b5b75.png)

PR #192, opened 2020-03-24
Removed the "Delete" action in item details #189

JellyWang-3:


mgrauer:
Did you intend to delete the `home.vue` file?

JellyWang-3:
I did it by accident and I am trying to recover it. Sorry about that.

mgrauer:
Not a problem, just wanted to check on things because that looked a bit unusual.

You can look into reverting the commit, e.g. [this link](https://code.likeagirl.io/how-to-undo-the-last-commit-393e7db2840b), or probably better, you could reset your branch head to the previous commit and then force push (and you could look into the `--force-with-lease` option). You should expect to run into some more tricky parts of git if you have never done any of these before, so don't be surprised if it takes a while to sort out your confusion.

It isn't always a good idea to force push on a branch that you have already pushed to a remote repo, but in this case it should be ok because no one is likely to have based further work off of this branch.

JellyWang-3:
Thank you for the tips.

web/src/views/Home.vue
```diff
@@ -56,7 +56,7 @@ const actionKeys = [
       window.open(`${JUPYTER_ROOT}`, '_blank');
     },
   },
-  ...DefaultActionKeys.slice(1),
+  ...DefaultActionKeys.slice(1, 3),
```
dchiquit:
```suggestion
  DefaultActionKeys[1], // Download
  DefaultActionKeys[2], // Download (zip)
```

mgrauer:
Was this PR approved before it got merged? Looks like there might have been some review from Daniel, perhaps an out of band approval?

I want to make sure we are waiting for approval from another developer before merging.

JellyWang-3:
Sorry, my bad. I saw Daniel reviewed it so I merged. I will wait for approval before merging next time.
Should I revert and restore the branch?

mgrauer:
I think this one is fine to leave merged, thanks. Yes, let's wait for a more explicit approval, through the GitHub approval.

Issue #154, opened 2020-03-16
name/description should be textarea fields in metadata editor 


satra:
It is difficult to type a passage in description right now or a somewhat longer title.

PR #209, opened 2020-03-26
#154 Textfield to textarea

JellyWang-3:
Make name and description textarea instead of textfield

### DIFF
--- a/web/src/components/MetaNode.vue
+++ b/web/src/components/MetaNode.vue
@@ -17,36 +17,22 @@
     </v-select>
     <!-- changed v-text-field to v-textarea for name and description-->
     <!-- should I add auto-grow -->
-    <v-textarea
-      v-else-if="schema.long"
-      v-model="value"
-      dense
-      :label="schema.title"
-      :type="fieldType(schema)"
-      :class="leafClasses"
-      :readonly="schema.readOnly"
-    >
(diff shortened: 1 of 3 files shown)


brianhelba:
@AlmightyYakob As part of your review, can you verify that `yarn lint` passes? We don't have that set up in CI right now.

web/src/components/MetaNode.vue
```diff
@@ -15,20 +15,24 @@
         <v-icon v-else>mdi-circle-medium</v-icon>
       </template>
     </v-select>
-    <v-text-field
+    <!-- changed v-text-field to v-textarea for name and description-->
+    <!-- should I add auto-grow -->
```
AlmightyYakob: I think we can leave auto-grow out for now. The name/description shouldn't be long enough to warrant that happening anyway, and if it is it may take up more space than we want. If you agree then we can remove these comments.

JellyWang-3: yes I can remove the comments and push again

AlmightyYakob:
> @AlmightyYakob As part of your review, can you verify that `yarn lint` passes? We don't have that set up in CI right now.

Yep, passes linting no problem.

brianhelba:
We also need to get this into the `refactor` branch.

PR #211, opened 2020-03-26
Add stats enpoint

### LINKED ISSUE #179: Add an endpoint for high level archive statistics
- number of dandisets
- number of species
- number of users

dchiquit:
Add an endpoint that returns global dandiset stats. Fields include:

* Number of draft dandisets
* Number of published dandisets (currently always 0)
* Number of users
* Number of unique species
* Number of subjects
* Number of cells

Because of how weird the Mongo queries turned out to be I did some benchmarking and on my machine, the requests would stay <1s for up to 100,000 dandisets, which I think is reasonable for the near future.

Fixes #179

### DIFF
--- a/girder-dandi-archive/girder_dandi_archive/rest.py
+++ b/girder-dandi-archive/girder_dandi_archive/rest.py
@@ -2,9 +2,11 @@ from girder.api import access
 from girder.api.rest import Resource
 from girder.api.describe import autoDescribeRoute, describeRoute, Description
 from girder.constants import TokenScope
-from girder.models.setting import Setting
-from girder.models.folder import Folder
 from girder.exceptions import RestException
+from girder.models.collection import Collection
+from girder.models.folder import Folder
+from girder.models.setting import Setting
+from girder.models.user import User
 
 from .util import (
     DANDISET_IDENTIFIER_COUNTER,
@@ -22,6 +24,7 @@ class DandiResource(Resource):
         self.route("GET", (), self.get_dandiset)
         self.route("GET", ("list",), self.list_dandisets)
         self.route("POST", (), self.create_dandiset)
+        self.route("GET", ("stats",), self.stats)
 
     @access.user(scope=TokenScope.DATA_WRITE)
     @describeRoute(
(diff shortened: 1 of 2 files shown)


girder-dandi-archive/girder_dandi_archive/rest.py
```diff
@@ -98,3 +101,84 @@ def list_dandisets(self, limit, offset, sort):
...
+        # Here is the same benchmark but with these aggregate queries removed:
+        #    0 dandisets  |  0.004081225395202637
+        # 1000 dandisets  |  0.008220505714416505
+        # 2000 dandisets  |  0.011307811737060547
+        # 1000 more dandisets takes about 0.0035 more seconds
+
+        species_count = list(
+            Folder().collection.aggregate(
+                [
+                    {"$unwind": "$meta.dandiset.organism"},
```
brianhelba: Does adding a `$project` stage at the start make this significantly faster if there are lots of other fields on the dandiset metadata?

girder-dandi-archive/tests/rest/test_stats.py
```diff
@@ -0,0 +1,152 @@
...
+        "published_count": 0,
+        "user_count": 2,  # one admin user, one normal user
+        "species_count": 0,
+        "subject_count": 0,
+        "cell_count": 0,
+        "size": 0,
+    } == resp.json
+
+
+def test_stats_draft(server, user, dandiset_1):
```
brianhelba: Does this test need a `user`?

girder-dandi-archive/tests/rest/test_stats.py
```diff
@@ -0,0 +1,152 @@
...
+        "user_count": 2,  # one admin user, one normal user
+        "species_count": 2,
+        "subject_count": 0,
+        "cell_count": 0,
+        "size": 0,
+    } == resp.json
+
+
+def test_stats_subjects(server, user, dandiset_1):
+
```
brianhelba:
```suggestion
```

girder-dandi-archive/tests/rest/test_stats.py
```diff
@@ -0,0 +1,152 @@
...
+        "user_count": 2,  # one admin user, one normal user
+        "species_count": 0,
+        "subject_count": subject_count,
+        "cell_count": 0,
+        "size": 0,
+    } == resp.json
+
+
+def test_stats_cells(server, user, dandiset_1):
+
```
brianhelba:
```suggestion
```

girder-dandi-archive/girder_dandi_archive/rest.py
```diff
@@ -98,3 +101,84 @@ def list_dandisets(self, limit, offset, sort):
...
         )
+
+    @access.public
+    @describeRoute(Description("Global Dandiset Statistics"))
+    def stats(self, params):
+        drafts = get_or_create_drafts_collection()
+        draft_count = Collection().countFolders(drafts)
+
+        # TODO no way to publish
+        published_count = 0
```
brianhelba: Can you file an issue for this?

girder-dandi-archive/girder_dandi_archive/rest.py
dchiquit: I first wrote the queries, then did the benchmarking before doing any optimizations. Unless there are a vastly unexpected number of dandisets added in the near future, optimization now seemed premature.
`$project` doesn't work with indices according to the docs, but it's possible that it would speed up the pipeline to have less data. I can experiment if you think it's important at this stage.

girder-dandi-archive/tests/rest/test_stats.py
dchiquit: No, with caveats. I added it because dandiset_1 requires a user, with means `user_count` must be 2.
I was torn between including it to clearly signal why `user_count=2`, or excluding because it's unnecessary and leaving that 2 as a magic number. If there are any feelings one way or the other I'm happy to change it.

girder-dandi-archive/girder_dandi_archive/rest.py
dchiquit: https://github.com/dandi/dandiarchive/issues/212
That's my guess for how it would be implemented, but the whole publishing topic should probably be a much larger discussion.

girder-dandi-archive/tests/rest/test_stats.py
brianhelba: This would be great to have a brief code comment to explain.

PR #215, opened 2020-03-30
Dandiset Listing Component

### LINKED ISSUE #180: Create dandiset list component
We want to use this in the list of all dandisets, in my dandisets, and in search results page matching a query, in the subsection of search results for dandisets.

[Figma link](https://www.figma.com/proto/kEHOgqU0rcOIyCYIDlpSMV/DANDI-UI?node-id=16%3A78&scaling=min-zoom)

Screen Shot 2020-03-23 at 4 38 09 PM

AlmightyYakob:
Fixes #180.

Since I haven't actually created pages for this yet, the preview won't work for this, so I've included some screenshots below with mock data.

Screenshot from 2020-03-30 15-16-44

### DIFF
--- a/web/src/components/DandisetList.vue
+++ b/web/src/components/DandisetList.vue
@@ -1,0 +1,77 @@
+<template>
+  <v-row>
+    <v-list
+      two-line
+      subheader
+    >
+      <v-list-item
+        v-for="(item, i) in dandisets"
+        :key="i"
+        selectable
+        :to="{ name: 'dandisetLanding', params: { id: item._id } }"
+      >
+        <v-list-item-content>
+          <v-list-item-title>
+            <template v-if="item.meta.dandiset.version">
+              <v-chip
+                v-if="item.meta.dandiset.version === 'draft'"
+                small
+                color="amber lighten-3"
+                text-color="amber darken-4"
+              >
+                <b>DRAFT</b>
+              </v-chip>
+              <v-chip
+                v-else
+                small
+                color="light-blue lighten-4"
+                text-color="light-blue darken-3"
+              >
+                <b>{{ item.meta.dandiset.version }}</b>
+              </v-chip>
+            </template>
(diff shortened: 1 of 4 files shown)


AlmightyYakob:
Quick note about the implementation:

Currently the only props are `dandisets` and `dandisetDetails`. These are split up to mirror how girder returns this information, i.e. there needs to be a separate call made to get the number of folders and number of items (although we only care about the number of items atm). Please let me know if you think there's a better way to approach this, where we should be putting this burden, etc.

web/src/components/DandisetList.vue
```diff
@@ -0,0 +1,97 @@
+<template>
+  <v-row>
+    <v-list
+      two-line
+      subheader
+    >
+      <v-list-item
+        v-for="(item, i) in items"
+        :key="i"
```
brianhelba:
```suggestion
        v-for="item in items"
        :key="item._id"
```
The id is guaranteed to deterministically map to the same element after any changes to the data. The position `i` may not always deterministically yield the same element. See: https://vuejs.org/v2/api/#key

web/src/utils.js
```diff
@@ -49,11 +49,23 @@ function copyToClipboard(text) {
   return false;
 }
 
+function getDandisetContact(dandiset) {
```
brianhelba: Do you anticipate this being reused in more places than just the Dandiset landing page?

AlmightyYakob: That was my reasoning in having it in `utils.js`. For example, in the redesign of the [DLP](https://www.figma.com/proto/kEHOgqU0rcOIyCYIDlpSMV/DANDI-UI?scaling=min-zoom&node-id=52%3A656) the contact is shown.

brianhelba:
@AlmightyYakob I just fixed a merge conflict that I introduced in another PR. Should be good to merge now.

PR #216, opened 2020-03-30
Move girderRest out of store, into own file

AlmightyYakob:
This moves the `RestClient` object out of the state, and into `rest.js`, along with a couple of getters associated with it. These getters are now functions, and can be imported directly.

### DIFF
--- a/web/src/rest.js
+++ b/web/src/rest.js
@@ -1,0 +1,14 @@
+
+import { RestClient } from '@girder/components/src';
+
+const apiRoot = process.env.VUE_APP_API_ROOT || 'http://localhost:8080/api/v1';
+const girderRest = new RestClient({ apiRoot, setLocalCookie: true });
+
+const loggedIn = () => !!girderRest.user;
+const user = () => girderRest.user;
+
+export {
+  loggedIn,
+  user,
+};
+export default girderRest;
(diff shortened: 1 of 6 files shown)


AlmightyYakob:
You'll also notice that I'm placing the `loggedIn` and `user` helper functions directly into the computed props:

https://github.com/dandi/dandiarchive/blob/66c3d0c79267110642b68534b11a8af3ef5f0a35/web/src/components/DandiAppBar.vue#L224-L226

Since `computed` already expects functions, this does exactly what we want. I've tested it's reactivity and it seems to work without issue.

web/src/rest.js
```diff
@@ -0,0 +1,14 @@
+
+import { RestClient } from '@girder/components/src';
+
+const apiRoot = process.env.VUE_APP_API_ROOT || 'http://localhost:8080/api/v1';
+const girderRest = new RestClient({ apiRoot, setLocalCookie: true });
+
+const loggedIn = () => !!girderRest.user;
```
brianhelba: I think it may be helpful to have these as reactive properties within Vuex, but we can move them back to a separate module if / when appropriate.

web/src/rest.js
```diff
@@ -0,0 +1,14 @@
...
+const user = () => girderRest.user;
+
+export {
```
brianhelba: @AlmightyYakob Is there a reason you prefer this style of `export`, rather than just defining the exports inline with the declarations (e.g. `export const user = ...`)?

Issue #182, opened 2020-03-23
Implement search box component


mgrauer:
This should hook up to a search results page, using the component in #180 .

Currently this will support search of dandiset by name, possibly using folder search.

PR #218, opened 2020-03-31
Refactor app bar

brianhelba:
This requires additional styling and logic, but this can be merged now to separate some of the old code into discrete components.


brianhelba:
@AlmightyYakob Can you review this at a high level? There is certainly more work to do, but I want to merge what I have so far, to minimize future conflicts.

web/src/components/AppBar/AppBar.vue
```diff
@@ -0,0 +1,96 @@
...
+      v-if="!loggedIn"
+      :to="{ name: 'userLogin' }"
+      color="primary"
+      rounded
+    >
+      Login
+    </v-btn>
+    <UserMenu
+      v-if="loggedIn"
+    />
```
AlmightyYakob: To avoid duplicate v-ifs on the same condition:

```suggestion
    <UserMenu
      v-if="loggedIn"
    />
    <template v-else>
      <span>Want to create your own datasets?</span>
      <v-btn
        :to="{ name: 'userRegister' }"
        color="primary"
        outlined
        rounded
      >
        Create Account
      </v-btn>
      <v-btn
        :to="{ name: 'userLogin' }"
        color="primary"
        rounded
      >
        Login
      </v-btn>
    </template>
```

web/src/components/AppBar/UserMenu.vue
```diff
@@ -0,0 +1,83 @@
...
+      dense
+    >
+      <v-list-item>
+        <v-list-item-content>
+          Get Girder API Key
+        </v-list-item-content>
+        <v-list-item-action>
+          <v-icon>mdi-key</v-icon>
+        </v-list-item-action>
+      </v-list-item>
```
AlmightyYakob: Is this supposed to be hooked up to any click action currently?

web/src/App.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <v-app class="dandi-app">
-    <dandi-app-bar />
+    <AppBar />
```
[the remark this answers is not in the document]
brianhelba: So, this is purely a style issue, but I prefer keeping component names in CamelCase. When using single-file components, there's no HTML compatibility concerns, and it keeps the component name used in the `import`, the `components` property, and within the `<template>` all consistent.

web/src/components/AppBar/AppBar.vue
brianhelba: I considered that, but figured we may want to tweak the conditional visibility of some of these things (particularly the "Want to create your own datasets?" element). However, for "Create Account" and "Login", they'll always be visible under the same circumstances, so it may be easier to comprehend if we group them together in a `template` block.

I'l make that change.

web/src/components/AppBar/UserMenu.vue
brianhelba: Eventually, but that's a WIP that I'll have to fix later.

brianhelba:
> * Successful sign-in/register should redirect to another page, instead of staying on the login/register page. Maybe the home page or something
> * There's some missing spacing between elements in the app bar, between the "Welcome" button and the DANDI Image, and between elements on the right side.

Agreed, all of these things need to be fixed in future work. Since this change created and reorganized so many other components, I wanted to get it into the shared codebase ASAP.

web/src/components/SearchBox.vue
```diff
@@ -0,0 +1,52 @@
+<template>
```
brianhelba, line 1: @dchiquit I put this here in case you find it helpful, but feel free to delete as part of #182

web/src/views/CreateDandisetView/CreateDandisetView.vue
```diff
@@ -0,0 +1,72 @@
+<template>
```
brianhelba, line 1: @dchiquit I put this here in case you find it helpful, but feel free to delete as part of #188

web/src/components/AppBar/AppBar.vue
brianhelba: Fixed this, but left the "Want to create your own datasets?" out for now, because it may change and is easier to tweak in development if it's separate.

web/src/App.vue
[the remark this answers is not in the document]
brianhelba: @AlmightyYakob Let me know what you think. We should strive to be consistent across the codebase.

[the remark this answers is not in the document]
brianhelba: @AlmightyYakob Good point, I didn't even think about the grep-ing consideration. Ok, CamelCase it is.

PR #220, opened 2020-03-31
Unify editorconfig and line length rules

brianhelba:
The max line length is now always enforced at 100.


brianhelba:
This may be easier to review looking at separate commits.

brianhelba:
@AlmightyYakob @dchiquit Does this seem reasonable?

PR #225, opened 2020-03-31
fix: Updates early access message

### LINKED ISSUE #214: Update Early Access message
The [current message](https://github.com/dandi/dandiarchive/blob/82dd0c9c94f53ed2b4c0919d3f6e29d134096699/web/src/components/DandiAppBar.vue#L74-L75) says that we are in limited access mode and that public release isn't until March 2020. What should we update this to say?

satra:
Closes #214

### DIFF
--- a/web/src/components/DandiAppBar.vue
+++ b/web/src/components/DandiAppBar.vue
@@ -71,8 +71,14 @@
         </v-chip>
       </template>
       <span>
-        DANDI is currently running in limited access mode.
-        Public release isn't until March 2020.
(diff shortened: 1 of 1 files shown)


satra:
do not merge - because it's a tooltip and none of the links are clickable (or viewable).

brianhelba:
@satra Are you going to try and fix this, or should this PR be closed and this information get added somewhere else?

brianhelba:
Since this approach has complications, let's revisit a solution to apply against `refactor`, which should provide additional places to put the text. We can then discuss backporting to `master` if critical.

PR #227, opened 2020-03-31
Add Roboto font

dchiquit:
The default Vuetify font, Roboto, was not included, so the default
sans-serif system font was being used instead.

### DIFF
--- a/web/src/App.vue
+++ b/web/src/App.vue
@@ -18,6 +18,8 @@ export default {
 </script>
 
 <style scoped>
+@import url("https://fonts.googleapis.com/css?family=Roboto:100,300,400,500,700,900");
+
 .dandi-app {
   -webkit-font-smoothing: antialiased;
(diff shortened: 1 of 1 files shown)


brianhelba:
It looks like if we had enabled `vue-cli-plugin-vuetify`, this would have been configured automatically: https://github.com/vuetifyjs/vue-cli-plugins/blob/908f65d35478da2cd9df3569c2ce4c5ca1910468/packages/vue-cli-plugin-vuetify/generator/tools/fonts.js#L31

This is a little more tricky to enable in the presence of what `girder_web_components` is doing with Vuetify, so I'll file an issue.

dchiquit:
Interesting. Stylistically I would prefer putting the font in with the rest of the global styling, but that's a convincing argument to use `<link>` instead.

I will close this PR in favor of #229. That definitely feels like the better solution.

PR #232, opened 2020-04-02
Merge master into refactor

dchiquit:
Merging the work currently in `master` into the `refactor` branch. It contains the stats endpoint, some endpoint changes, and improved tests.

The only non-merge change I made as part of this PR is to add a new exception to `tox.ini`. Brian removed it from the `refactor` branch as it was unnecessary at the time, but one of the new tests from `master` requires the fixture `fsAssetStore`, which is a camelCase function argument.


girder-dandi-archive/tox.ini
```diff
@@ -70,6 +70,8 @@ ignore =
     E123
     # whitespace before ':'
     E203,
+    # lowercase argument name
+    N803,
```
brianhelba, line 74: I'd prefer we add `# noqa: N803` to the offending lines, rather than allowing a global exception for all our code.

brianhelba:
@dchiquit Also, can you re-generate this with #233 included (which was merged just now). Sorry for the hassle, but I think it will prevent future hassles down the line if we have #233 in `refactor`?

girder-dandi-archive/tests/rest/test_stats.py
```diff
@@ -125,7 +125,7 @@ def test_stats_cells(server, dandiset_1):
     } == resp.json
 
 
-def test_stats_size(server, fsAssetstore, user, dandiset_1):
+def test_stats_size(server, fsAssetstore, user, dandiset_1):  # NOQA
```
brianhelba, line 128:
```suggestion
def test_stats_size(server, fsAssetstore, user, dandiset_1):  # noqa: N803
```

Should be lowercase: http://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html#in-line-ignoring-errors
Should be as specific to a single error as possible, so other errors on that line aren't silenced.

PR #234, opened 2020-04-02
Run vue-cli-plugin-vuetify generation

### LINKED ISSUE #229: Enable vue-cli-plugin-vuetify
`vue-cli-plugin-vuetify` configures fonts and icon links automatically: https://github.com/vuetifyjs/vue-cli-plugins/blob/908f65d35478da2cd9df3569c2ce4c5ca1910468/packages/vue-cli-plugin-vuetify/generator/tools/fonts.js#L31

Icons are being embedded by `girder_web_components`, so they don't necessarily need to be included by link here.

Also, the plugin configures Babel to work with Vuetify properly: https://github.com/vuetifyjs/vue-cli-plugins/pull/58

Simply running `vue add vuetify` (or `npx -p @vue/cli vue add vuetify`) doesn't work cleanly in the presence of how `vuetify` is imported by `girder_web_components`.

brianhelba:
This is the essential result of running `npx -p @vue/cli vue add vuetify`.

Fixes #229.

### DIFF
--- a/web/public/index.html
+++ b/web/public/index.html
@@ -6,6 +6,8 @@
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <link rel="icon" href="<%= BASE_URL %>favicon.ico">
     <title>DANDI Archive</title>
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:100,300,400,500,700,900">
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@mdi/font@latest/css/materialdesignicons.min.css">
   </head>
   <body>
     <noscript>
--- a/web/vue.config.js
+++ b/web/vue.config.js
@@ -5,4 +5,7 @@ module.exports = {
   configureWebpack: {
     devtool: 'source-map',
   },
+  transpileDependencies: [
+    'vuetify',
+  ],
 };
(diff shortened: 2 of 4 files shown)


web/public/index.html
```diff
@@ -6,6 +6,8 @@
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <link rel="icon" href="<%= BASE_URL %>favicon.ico">
     <title>dandi</title>
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:100,300,400,500,700,900">
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@mdi/font@latest/css/materialdesignicons.min.css">
```
brianhelba, line 10: This is not strictly necessary right now, since the MDI are [inlined by GWC](https://github.com/girder/girder_web_components/blob/2ee63d169cd21125654871afac85ee8c14714c0b/src/components/index.js#L1). However, we may drop GWC at some point. Alternatively, we may want to consider requesting an upstream change to GWC to make inclusion of the MDI a build-time requirement, and leave the implementation choice up to the downstream builder.

web/vue.config.js
```diff
@@ -5,4 +5,7 @@ module.exports = {
   configureWebpack: {
     devtool: 'source-map',
   },
+  transpileDependencies: [
+    'vuetify',
+  ],
```
dchiquit, line 10: I could not determine what this achieves. The project was working without this code, so I guess there's some edge case we might run in to with a different project configuration?

brianhelba: https://github.com/vuetifyjs/vue-cli-plugins/pull/58

:man_shrugging:

PR #239, opened 2020-04-03
Fix text selection refactor

AlmightyYakob:
Corresponding change from #238 into refactor.

Also fixes a missing `</template>` tag that was preventing local builds from succeeding.

### DIFF
--- a/web/src/components/MetaNode.vue
+++ b/web/src/components/MetaNode.vue
@@ -176,7 +176,8 @@
         </v-list-item>
       </v-list>
     </v-menu>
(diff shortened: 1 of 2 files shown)


web/src/components/MetaNode.vue
```diff
@@ -176,7 +176,8 @@
         </v-list-item>
       </v-list>
     </v-menu>
-  </div>
+  </template>
```
brianhelba, line 179: Do you know why linting didn't catch this?

PR #242, opened 2020-04-06
Add Public Dandisets View

### LINKED ISSUE #181: Create a public dandiset list page
This page will list public dandisets and will use the component from #180

dchiquit:
Add the first draft of the public dandisets view. I anticipate the
following changes at some point:

* UX styling for the list dandisets component. It is currently not
filling the page and still needs some more work done to get all the
stats it needs.
* Comprehensive search solution. I included a placeholder `SearchField`
that looks ok, but does nothing. When search is implemented properly it
should be replaced with a general component.
* Better search option chips. We don't have the information available to
do all of Jared's recommended chips, and the UX of what I did is
probably unsatisfactory in some way. It is fully functional and
easily fixable though, so I think it's OK for an MVP.

Some screenshots:
Screenshot from 2020-04-06 09-12-20
Screenshot from 2020-04-06 09-12-36
Screenshot from 2020-04-06 09-12-55

Fixes #181

### DIFF
--- a/web/src/store/publicDandisets.js
+++ b/web/src/store/publicDandisets.js
@@ -20,7 +20,6 @@ export default {
     setSearchSettings(state, { sort }) {
       state.sort = sort;
       state.page = 1;
-      this.dispatch('publicDandisets/reload');
     },
     setDandisets(state, { dandisets, total }) {
       state.dandisets = dandisets;
@@ -29,7 +28,7 @@ export default {
   },
   actions: {
     async reload({ commit, state }) {
-      const { data: dandisets, headers } = await girderRest.get('dandi/list', {
+      const { data: dandisets, headers } = await girderRest.get('dandi', {
         params: {
           limit: DANDISETS_PER_PAGE,
           offset: (state.page - 1) * DANDISETS_PER_PAGE,
@@ -39,6 +38,10 @@ export default {
       });
       const total = headers['girder-total-count'];
       commit('setDandisets', { dandisets, total });
+    },
+    async changeSearchSettings({ commit, dispatch }, { sort }) {
+      commit('setSearchSettings', { sort });
+      dispatch('reload');
     },
   },
 };
--- a/web/src/store/publicDandisets.js
+++ b/web/src/store/publicDandisets.js
@@ -1,0 +1,44 @@
+import girderRest from '@/rest';
+
+const DANDISETS_PER_PAGE = 8;
+
+export default {
+  namespaced: true,
+  state: {
+    // search settings
+    sort: {
+      field: 'created',
+      direction: 1,
+    },
+    page: 1,
+
+    // retrieved from server
+    dandisets: [],
+    pages: 0,
+  },
+  mutations: {
+    setSearchSettings(state, { sort }) {
+      state.sort = sort;
+      state.page = 1;
+      this.dispatch('publicDandisets/reload');
+    },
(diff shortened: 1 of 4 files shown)


web/src/views/PublicDandisetsView/PublicDandisetsView.vue
```diff
@@ -1,11 +1,96 @@
...
+    <DandisetList
+      class="
+        mx-12
+        my-12"
+      :dandisets="$store.state.publicDandisets.dandisets"
+    />
+    <v-pagination
+      v-model="$store.state.publicDandisets.page"
+      :length="$store.state.publicDandisets.pages"
+      :value="$store.state.publicDandisets.page"
```
AlmightyYakob, line 41: The `v-model` on line 39 will directly modify state, which is a strict anti-pattern when it comes to vuex. If you need to model something to the state, I'd look at [computed setters](https://vuejs.org/v2/guide/computed.html#Computed-Setter). That way, you can commit a mutation if you need to change the value. If doing that, you can't mix it with `mapState`, so you'd just have to directly reference `$store.state.publicDandisets.page` in the getter.

In general, I try to avoid referencing `$store` directly in the template if I can. Even if you use a computed getter/setter for `page`, I still think it'd be worth using `mapState` on `dandisets` and `pages`, and referencing those mapped props here.

I'm curious what @brianhelba's opinion on this is though.

web/src/store/publicDandisets.js
```diff
@@ -0,0 +1,44 @@
...
+
+    // retrieved from server
+    dandisets: [],
+    pages: 0,
+  },
+  mutations: {
+    setSearchSettings(state, { sort }) {
+      state.sort = sort;
+      state.page = 1;
+      this.dispatch('publicDandisets/reload');
```
AlmightyYakob, line 23: Mutations aren't supposed to dispatch actions, only actions should, and mutations are supposed to very narrowly change properties of state. So the action of say `changeSort` would commit the mutation of setting `state.sort` and `state.page`, and then dispatch this action.

In general, I don't think you're supposed to reference `this` inside of a vuex store, I don't think what it exactly references is defined, which is why ideally in the `vuex` world you wouldn't be able to dispatch actions here, as `dispatch` isn't exposed through passed arguments.

web/src/store/publicDandisets.js
```diff
@@ -0,0 +1,44 @@
...
+    },
+    setDandisets(state, { dandisets, total }) {
+      state.dandisets = dandisets;
+      state.pages = Math.ceil(total / DANDISETS_PER_PAGE);
+    },
+  },
+  actions: {
+    async reload({ commit, state }) {
+      const { data: dandisets, headers } = await girderRest.get('dandi/list', {
```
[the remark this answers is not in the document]
dchiquit: My mistake, I merged `master` in to the `refactor` branch and forgot to restart my girder server.

web/src/views/PublicDandisetsView/PublicDandisetsView.vue
dchiquit: I fixed this, I don't know what I was thinking.

I think that is an excellent policy.

brianhelba: I also try to use (and I'm typically able to) the `map` helpers everywhere, and avoid `$store.state` entirely. It helps make it much more clear (to me at least) which parts of my store surface a given component needs.

PR #252, opened 2020-04-08
My dandisets page

### LINKED ISSUE #187: Create a my dandisets page
Uses #180 that shows the dandisets that I own.

AlmightyYakob:
Fixes #187.

This change is a bit odd, as it's basically a mirror of the Public Dandisets Page, modified to use the new included endpoint. I wanted to have a discussion here about our thoughts on combining these two things, to make sure we're all on the same page. For example, the `publicDandisets` and `myDandisets` vuex modules differ by one line, and the differences between `MyDandisetsView` and `PublicDandisetsView` is mainly in the API calls. 

One idea I had for the `vuex` side of things, is to have state for a general purpose "Dandiset Listing", which could be used by both pages, and would just use different actions to populate the state. As far as sharing the pages themselves, it seems to me that the template would remain largely the same, aside from the title changing, and that it'd mainly be the logic that changes, but again the changes between them are small. However I'm not sure how we feel about putting ourselves in a box like that, if we think that these pages are going to diverge significantly.

Another factor here, is that when it comes to the search results page, it will likely be very similar to these two pages, so that might be even more reason to unify.

Please let me know what you guys think, and any ideas you have on this.


satra:
@AlmightyYakob - another thing to keep in mind here is that mydandisets may not simply be the one you are an owner of, but ones that you are a contributor to, and may have different read/write privileges.

so as we get to more filtering options, then the component can be the same with perhaps the listing simply expressing what you are filtering by.

this is only food for thought for now.

AlmightyYakob:
@satra Good point, thanks for the input

yarikoptic:
+1 on having a dandiset listing (all, vs "my", or later may be "without releases", "with new drafts", "starred", "of collaborators", etc) be just a filter, thus reusing the same web ui component/code/etc.

web/src/utils.js
```diff
@@ -60,6 +60,30 @@ function getDandisetContact(dandiset) {
   return null;
 }
 
+const basicDandisetSortingOptions = [
```
brianhelba, line 63: Any reason you don't just `export const basicDandisetSortingOptions` here? Or more generally just export everything inline with where it's declared in this file?

web/src/views/MyDandisetsView/MyDandisetsView.vue
```diff
@@ -1,11 +1,86 @@
...
+        return this.$store.state.publicDandisets.page;
+      },
+      set(page) {
+        this.$store.state.publicDandisets.page = page;
+      },
+    },
+    ...mapState('myDandisets', ['dandisets', 'pages']),
+  },
+  created() {
+    this.reload();
```
brianhelba, line 75: To make sure I'm clear, the reason we're using Vuex at all here is to persist search settings if the user navigates to a different page and back here? And that itself is a shorter term solution, until we are able to serialize our settings within the route, correct?

web/src/views/MyDandisetsView/MyDandisetsView.vue
```diff
@@ -1,11 +1,86 @@
...
+  },
+  methods: {
+    setSort(sort) {
+      this.$store.dispatch('myDandisets/changeSearchSettings', { sort });
```
brianhelba, line 79: Any reason you're not using `mapActions` here?

web/src/views/MyDandisetsView/MyDandisetsView.vue
```diff
@@ -1,11 +1,86 @@
...
+      options: basicDandisetSortingOptions,
+    };
+  },
+  computed: {
+    page: {
+      get() {
+        return this.$store.state.publicDandisets.page;
+      },
+      set(page) {
+        this.$store.state.publicDandisets.page = page;
```
brianhelba, line 69: You shouldn't mutate Vuex-managed state outside of a mutation, right?

web/src/store/index.js
```diff
@@ -12,6 +13,7 @@ export default new Vuex.Store({
   modules: {
     girder,
     publicDandisets,
+    myDandisets,
```
brianhelba, line 16: We can certainly register the same module store times under different module names: https://vuex.vuejs.org/guide/modules.html#module-reuse

Would it be possible to merge the logic of `publicDandisets` and `myDandisets` into a single store definition, and register it twice under different module names?

web/src/utils.js
brianhelba, line 63: The default sort of the API is on `meta.dandiset.identifier`. Is there a reason that's not an option presented in the GUI?

web/src/views/MyDandisetsView/MyDandisetsView.vue
AlmightyYakob: Good catch, result of copy paste.

web/src/views/MyDandisetsView/MyDandisetsView.vue
AlmightyYakob: Correct

web/src/views/MyDandisetsView/MyDandisetsView.vue
AlmightyYakob: Nope, I'll fix it

web/src/store/index.js
AlmightyYakob: Ahh good point, yeah I'll try that

web/src/utils.js
AlmightyYakob: Things were already being exported through the object at the end of the file, so I just added to that.

AlmightyYakob: Since this is basically a duplicate of the public dandisets page, I don't have a good reason, but that could be a useful addition.

web/src/store/index.js
AlmightyYakob: f1a87ed addresses this.

dchiquit:
I just did a diff of `PublicDandisetsView.vue` and `MyDandisetsView.vue`, and there are only 5 differences: two because they are called different things, and three because of the the different vuex module.

I think it would actually make sense to make a `@/components/DandisetListingPage.vue` that takes the name of the vuex module and the desired title as props. The views would basically just be the bare minimum required to invoke that common page:
```
<template>
  <DandisetPage title="My Dandisets" module="myDandisets" />
</template>
<script>
import DandisetListingPage from '@/components/DandisetListingPage.vue';
export default {
  name: 'MyDandisetsView',
  components: { DandisetListingPage },
}
</script>
```
`dandisetListing.js` is currently handling the different API endpoint required for the different pages, so we don't even have to manage that at the view level. This also makes the search page even more trivial; all we have to do is register it, and add a new API endpoint that returns all public dandisets + my dandisets.

Any thoughts?

AlmightyYakob:
> Any thoughts?

I like the idea in principle, I agree that since those pages only differ by name, it would make sense to combine them like this. I'm willing to try this, as it would get rid of some redundancy now, but I've got the feeling that this will end up being too restrictive if we ever decide to change things. It's also good to keep in mind that we'll be removing vuex from this shortly, once we switch to tracking state with the URL.

I like the idea, but I'd like to hear @brianhelba's thoughts before moving forward with this.

dchiquit:
I guess the upside of having all the code for all the pages in one file also carries the downside of not being able to modify that code on a per-page basis. My personal assessment is that any feature we would want to add on one page would also be useful on the other pages, at least for the foreseeable future.

AlmightyYakob:
> My personal assessment is that any feature we would want to add on one page would also be useful on the other pages, at least for the foreseeable future.

Yeah I'd probably agree, but you never know. Barring any serious concerns from @brianhelba, I think we should be alright giving that approach a go.

brianhelba:
I'm ok with merging the pages now, but it's likely that the API for fetching published dandisets will be substantially different from that for draft dandisets.

@mgrauer will need to confirm this, but it's my understanding that:
* Public Dandisets: consists of only published dandisets
* My Dandisets: consists of both draft and published dandisets, both where the current user is the dandiset owner

Assuming this is the case, if we continue with merged pages, I guess we'd make Public Dandisets simply have a no-op query against the draft dandiset API?

satra:
till we get to implementing embargo, all dandisets are public. i would like to keep it that way. so public should list all public dandisets. i think the mock-up i saw from @mgrauer clarified a separation between dandisets that were draft only vs those that have been published.

once embargo is implemented, my dandisets could list dandisets that are embargoed.

dchiquit:
It sounds like the presentation layer is still going to be the same for all the pages, even if the data is coming from different endpoints. The common component can accept the vuex module as prop for now, but it could also accept an array of dandiset info for rendering. Each page could then have its own code for hitting whatever API it wants to hit.

brianhelba:
I'm fine with whatever @AlmightyYakob and @dchiquit want to go with. Everything sounds reasonable.

AlmightyYakob:
Alright, I can make the change @dchiquit suggested.

mgrauer:
I agree that 

>it's likely that the API for fetching published dandisets will be substantially different from that for draft dandisets.

and that public dandisets page listing will be a combination of published and draft dandisets as all are public (without embargo implemented). i'm not sure what value users get from seeing draft dandisets that they aren't owners of though -- @satra ?

and that for now (ignoring embargo cases), my dandisets will just be draft dandisets that i have ownership of

satra:
>  i'm not sure what value users get from seeing draft dandisets that they aren't owners of though

they could be collaborators or generally interested in playing with data, just like we play with non-released software on github.

mgrauer:
a problem is that because that data is in a draft form, it could keep changing out from under users, and they won't be able to get back to any particular set of data because there is no snapshotting in the draft. non-released software on github does have persistent commits that are both reachable and guaranteed not to change (assuming no force pushes, etc). i'm not ready to propose any particular change, just noting some dataset usability issues.

satra:
@mgrauer - indeed we would want to think about snapshotting, versioning, etc.,. but even knowing that the dataset exists and is in draft form and can be used can have scientific ramifications. people may start collaborating and pulling data together. some others may find some issues with data. so having unreleased data, i think, can be beneficial especially if well annotated as a draft.

AlmightyYakob:
@dchiquit @brianhelba I've made those changes, however I'd like to say I'm not a huge fan of what's required for that. Namely, since `mapState` and `mapActions` aren't functions of the component, but are imported from vuex, they don't have access to `this`, so we manually have to map each prop. It's a relatively small price to pay though, and since this will most likely go away when we store state in the URL, I'm okay with it.

brianhelba:
@AlmightyYakob This seems like a reasonable trade-off and approach.

web/src/components/DandisetPage.vue
```diff
@@ -0,0 +1,103 @@
...
+  </div>
+</template>
+
+<script>
+import DandisetList from '@/components/DandisetList.vue';
+import SearchField from '@/views/PublicDandisetsView/SearchField.vue';
+import { basicDandisetSortingOptions } from '@/utils';
+
+export default {
+  name: 'DandisetPage',
```
dchiquit, line 51: Thoughts on calling this `DandisetsPage` (plural)? `DandisetPage` sounds a lot like `DandisetLandingView`.

PR #254, opened 2020-04-09
137 check status

### LINKED ISSUE #137: Fix missing status check on API key actions
Raised from https://github.com/dandi/dandiarchive/pull/134#discussion_r392200076.

JellyWang-3:
Unsure about the error displaying message and line 13 in store.js
1. Should I be more specific about the error message instead of just "Request failed"
2. I set currentStatus to 400 because if I passed an invalid param, then I get an error in console log.The currentStatus remain null/undefined. But I feel hard code it to 400 is not a good practice. Is there a way to get around that?

Fixes #137


web/.env.development
```diff
@@ -1 +1 @@
-VUE_APP_API_ROOT=http://localhost:8080/api/v1
```
brianhelba, line 1: Don't touch this file in this branch.

AlmightyYakob:
Had a discussion with @JellyWang-3 offline, and gave suggestions for the current state of this PR.

brianhelba:
Fixed in `refactor` by #284. Closing.

PR #257, opened 2020-04-13
Add Create Dandiset page

### LINKED ISSUE #188: Create a create a dandiset page

dchiquit:
Update the existing Create Dandiset view and add nav bar button to
navigate to it. Visiting the page when not logged in will redirect to
the home screen.

Screenshot from 2020-04-13 10-26-59

Closes #188

### DIFF
--- a/web/src/components/AppBar/AppBar.vue
+++ b/web/src/components/AppBar/AppBar.vue
@@ -49,16 +49,24 @@
 
     <v-spacer />
 
-    <span v-if="!loggedIn" class="mr-1">
-      Want to create your own datasets?
-    </span>
-
     <template v-if="loggedIn">
+      <v-btn
+        :to="{ name: 'createDandiset' }"
+        exact
+        class="mx-3"
+        color="primary"
+        rounded
+      >
+        New Dandiset
+      </v-btn>
       <UserMenu
         v-if="loggedIn"
       />
     </template>
     <template v-else>
+      <span class="mr-1">
+        Want to create your own datasets?
+      </span>
       <v-btn
         :to="{ name: 'userRegister' }"
         class="mx-1"
--- a/web/src/components/AppBar/AppBar.vue
+++ b/web/src/components/AppBar/AppBar.vue
@@ -59,9 +59,7 @@
       >
         New Dandiset
       </v-btn>
-      <UserMenu
-        v-if="loggedIn"
-      />
+      <UserMenu />
     </template>
     <template v-else>
       <span class="mr-1">
--- a/web/src/views/CreateDandisetView/CreateDandisetView.vue
+++ b/web/src/views/CreateDandisetView/CreateDandisetView.vue
@@ -57,7 +57,7 @@ export default {
   },
   created() {
     if (!loggedIn()) {
-      this.$router.push({name: 'home'});
+      this.$router.push({ name: 'home' });
     }
   },
   methods: {
--- a/web/src/views/CreateDandisetView/CreateDandisetView.vue
+++ b/web/src/views/CreateDandisetView/CreateDandisetView.vue
@@ -56,7 +56,7 @@ export default {
       return !(this.name && this.description);
     },
   },
-  mounted() {
+  created() {
     if (!loggedIn()) {
       this.$router.push({name: 'home'});
     }
--- a/web/src/views/CreateDandisetView/CreateDandisetView.vue
+++ b/web/src/views/CreateDandisetView/CreateDandisetView.vue
@@ -40,6 +40,9 @@
 </template>
 
 <script>
+import { loggedIn } from '@/rest';
+import girderRest from '@/rest';
+
 export default {
   name: 'CreateDandisetView',
   data() {
@@ -53,16 +56,21 @@ export default {
       return !(this.name && this.description);
     },
   },
+  mounted() {
+    if (!loggedIn()) {
+      this.$router.push({name: 'home'});
+    }
+  },
   methods: {
     async register_dandiset() {
       const { name, description } = this;
-      const { status, data } = await this.girderRest.post('dandi', null, { params: { name, description } });
+      const { status, data } = await girderRest.post('dandi', null, { params: { name, description } });
 
       if (status === 200) {
         this.name = '';
         this.description = '';
         this.$router.push({
-          name: 'dandiset-metadata-viewer',
+          name: 'dandisetLanding',
           params: { id: data._id },
         });
       }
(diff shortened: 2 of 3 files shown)


web/src/components/AppBar/AppBar.vue
```diff
@@ -49,16 +49,24 @@
...
+        exact
+        class="mx-3"
+        color="primary"
+        rounded
+      >
+        New Dandiset
+      </v-btn>
       <UserMenu
         v-if="loggedIn"
       />
```
AlmightyYakob, line 64: Since this `v-if` is redundant with the wrapping template:
```suggestion
      <UserMenu />
```

web/src/views/CreateDandisetView/CreateDandisetView.vue
```diff
@@ -53,16 +56,21 @@ export default {
       return !(this.name && this.description);
     },
   },
+  mounted() {
```
AlmightyYakob, line 59: Any particular reason to use the `mounted` lifecycle hook instead of `created`?

web/src/views/CreateDandisetView/CreateDandisetView.vue
```diff
@@ -53,16 +56,21 @@ export default {
...
+    }
+  },
   methods: {
     async register_dandiset() {
       const { name, description } = this;
-      const { status, data } = await this.girderRest.post('dandi', null, { params: { name, description } });
+      const { status, data } = await girderRest.post('dandi', null, { params: { name, description } });
 
       if (status === 200) {
         this.name = '';
```
brianhelba, line 70: You can probably remove these, since the component should be destroyed when navigating to another view.

They are a holdover from when this form was embedded in a global dialog.

web/src/views/CreateDandisetView/CreateDandisetView.vue
```diff
@@ -53,16 +56,21 @@ export default {
...
     },
   },
+  mounted() {
+    if (!loggedIn()) {
+      this.$router.push({name: 'home'});
+    }
+  },
   methods: {
     async register_dandiset() {
       const { name, description } = this;
```
brianhelba, line 66: I realize you didn't write this, but this line is a little too clever for my liking. If you want to change it, feel free. If you disagree though, I don't feel strongly.

web/src/views/CreateDandisetView/CreateDandisetView.vue
```diff
@@ -40,6 +40,9 @@
 </template>
```
brianhelba, line 40: On line 34 (I can't make an inline comment there), could you change `v-slot:loader` to `#loader`? Devs may copy-paste from this form in the future, and it'd be good to have consistent usage of syntax.

web/src/views/CreateDandisetView/CreateDandisetView.vue
```diff
@@ -53,16 +56,21 @@ export default {
       return !(this.name && this.description);
     },
   },
+  mounted() {
+    if (!loggedIn()) {
+      this.$router.push({name: 'home'});
+    }
+  },
   methods: {
     async register_dandiset() {
```
brianhelba, line 65: Should be camelCase.

web/src/components/AppBar/AppBar.vue
```diff
@@ -49,16 +49,22 @@
...
-      Want to create your own datasets?
-    </span>
-
     <template v-if="loggedIn">
-      <UserMenu
-        v-if="loggedIn"
-      />
+      <v-btn
+        :to="{ name: 'createDandiset' }"
+        exact
```
brianhelba, line 55: This usage of `exact` is a smart idea. Until we have a reason to want the default regex matching, it's probably safer to always do it exactly. I'll file an issue.

web/src/views/CreateDandisetView/CreateDandisetView.vue
brianhelba: Agreed. Unless it directly relies on the DOM, I'd default to always using `created`?

web/src/views/CreateDandisetView/CreateDandisetView.vue
```diff
@@ -53,16 +56,21 @@ export default {
...
       const { name, description } = this;
-      const { status, data } = await this.girderRest.post('dandi', null, { params: { name, description } });
+      const { status, data } = await girderRest.post('dandi', null, { params: { name, description } });
 
       if (status === 200) {
```
brianhelba, line 69: Do we need this guard? I think Axios (and hence `girderRest`) will reject its promise (throwing an exception here) if there's a non-success status code.

Unless we can thoroughly recover from errors, I'd prefer not guarding against them at all, so they can bubble up, break the app in an obvious way, and be caught by Sentry.

web/src/views/CreateDandisetView/CreateDandisetView.vue
AlmightyYakob: I actually like this style, depending on the context, as if you're using the specified variables enough, it avoids the clutter of repeatedly specifying `this`. Whether it should specifically used here could be up for debate I suppose, but just wanted to give my input on that pattern in general.

web/src/views/CreateDandisetView/CreateDandisetView.vue
dchiquit: I had a vague idea of `mounted` being more "reliable", but I'm not sure where I got that idea from. `created` is the better option here, thanks

web/src/views/CreateDandisetView/CreateDandisetView.vue
brianhelba: @AlmightyYakob sounds good.

web/src/views/CreateDandisetView/CreateDandisetView.vue
```diff
@@ -40,6 +40,9 @@
 </template>
 
 <script>
+import { loggedIn } from '@/rest';
+import girderRest from '@/rest';
```
AlmightyYakob, line 44: Something the linter caught:

```suggestion
import girderRest, { loggedIn } from '@/rest';
```

web/src/views/CreateDandisetView/CreateDandisetView.vue
dchiquit: I personally prefer this pattern here. I tried removing it, and got this:
```
      const { data } = await girderRest.post('dandi', null, {
        params: {
          name: this.name,
          description: this.description,
        },
      });
```
To me the first pattern is more legible.

web/src/views/CreateDandisetView/CreateDandisetView.vue
dchiquit: I verified that this does indeed break with a non-200 response code

PR #260, opened 2020-04-13
Add user registration page

dchiquit:
This works for normal registrations, but I have not yet been able to
test it with an OAuth provider.

Waiting to put this up for review until we have netlify previews working on the refactor branch.

### DIFF
--- a/web/src/views/UserRegisterView/UserRegisterView.vue
+++ b/web/src/views/UserRegisterView/UserRegisterView.vue
@@ -1,6 +1,6 @@
 <template>
   <v-container>
-    <GirderRegister :oauthProviders="oauthProviders" />
+    <GirderRegister :oauth-providers="oauthProviders" />
   </v-container>
 </template>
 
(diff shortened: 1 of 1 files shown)


brianhelba:
@AlmightyYakob Can you please review?

web/src/views/UserRegisterView/UserRegisterView.vue
```diff
@@ -1,11 +1,40 @@
...
+  asyncComputed: {
+    async oauthProviders() {
+      return (
+        await girderRest.get('oauth/provider', {
+          params: {
+            redirect: `${window.location.href}${OauthTokenPrefix}{girderToken}${OauthTokenSuffix}`,
+            list: true,
+          },
+        })
+      ).data;
```
AlmightyYakob, line 29: I generally prefer destructuring for clarity, but I don't feel too strongly about it:
```suggestion
      const { data } = await girderRest.get('oauth/provider', {
        params: {
          redirect: `${window.location.href}${OauthTokenPrefix}{girderToken}${OauthTokenSuffix}`,
          list: true,
        },
      });

      return data;
```

dchiquit:
Changing the redirect from `${window.location.href}` to a relative path results in a redirect to `girder.dandiarchive.com/{relativePath}`. The only other solutions I can think of are to hardcode the path (icky) or to parse `window.location.href` (also icky). For now I think this is OK.

PR #264, opened 2020-04-13
Remove deploy-specific configuration from repo

brianhelba:
The settings in `.env.production` are specific to a particular deploy of the app. All production builds should not necessarily use these same settings.

### DIFF
--- a/web/.env.production
+++ b/web/.env.production
@@ -1,2 +1,0 @@
-VUE_APP_API_ROOT=https://girder.dandiarchive.org/api/v1
-VUE_APP_SENTRY_DSN=https://24675005ee044d2e9a8245833f1f8821@sentry.io/4282414


dchiquit:
Where is this configuration going to be stored now?

brianhelba:
For the https://gui.dandiarchive.org/ deployment, it's stored in the Netlify configuration (and injected via environment variables). We want to provide more documentation about how dandiarchive.org is provisioned, but that's a separate ongoing task.

However, it's a somewhat open question to me whether:
* the `dandiarchive` codebase is intended to exclusively power (and be tightly coupled to) the `*.dandiarchive.org` deployment
* the `*.dandiarchive.org` deployments are just one canonical deployment of a more general `dandiarchive` codebase

PR #265, opened 2020-04-13
Improve the development docs

brianhelba:


README.md
```diff
@@ -1,83 +1,75 @@
 # dandiarchive [![CircleCI](https://circleci.com/gh/dandi/dandiarchive/tree/master.svg?style=svg)](https://circleci.com/gh/dandi/dandiarchive/tree/master)
-Infrastructure and code for the dandiarchive
+Infrastructure and code for the DANDI Archive.
 
 Folders in this repo:
 
-- Ansible: Ansible code for deploying the girder.dandiarchive.org site.
-- web: Vue.js code for holding a Girder Web Components based client deployed at gui.dandiarchive.org .
-- girder-dandi-archive: A Girder plugin for customizing the DANDI archive Girder instance.
+- `girder-dandi-archive`: A Girder plugin which serves the DANDI Archive.
```
yarikoptic, line 6: I am not quite sure what "serves" means here.  "customizing ... Girder instance" was somewhat more accessible to me

web/vue.config.js
```diff
@@ -8,4 +8,7 @@ module.exports = {
   transpileDependencies: [
     'vuetify',
   ],
+  devServer: {
+    port: 8085,
+  }
```
yarikoptic, line 13: this is beyond "docs"... was it intended to be committed and if so - better goes into separate commit with a reason what for

README.md
mgrauer: Maybe "A Girder plugin which adds the custom endpoints for the DANDI Archive."

web/vue.config.js
brianhelba: Previously, the docs instructed developers to always run `yarn run serve --port 8085`. This internalizes that configuration, which makes the developer experience simpler and more consistent.

README.md
brianhelba: Sounds good to me.

web/vue.config.js
brianhelba: To further clarify, the only impact of this change is to simplify the commands typed by front-end developers. Indeed, for developers that really want to run the dev server on another port, the `--port` option will continue to override this.

It does not effect the way that the site is built for production.

brianhelba:
@mgrauer @yarikoptic Responded to and addressed feedback.

web/vue.config.js
yarikoptic: thanks!  I just wanted to make sure that nothing "leaked" since title (or PR and the commit where the change is mixed in) said "Improve ... docs". Such descriptions better be placed into the commit messages introducing the changes, that could help later to investigate how the change came about if comes needed.

brianhelba:
> thanks! I just wanted to make sure that nothing "leaked" since title (or PR and the commit where the change is mixed in) said "Improve ... docs". Such descriptions better be placed into the commit messages introducing the changes, that could help later to investigate how the change came about if comes needed.

This brings up a more general question about how hard we want to try to maintain Git hygene. In general, while most of our PRs tend to use branches with reasonable commit messages and squashed local development, I don't think it's something we're enforcing very strongly. We could try to be much more strict about this, but in general, I find code comments to be far more effective at explaining *why* a change was made and preventing future maintainers from unknowingly changing it.

So, I'll add a comment now to explain why we use port 8085 in the first place (whether specified via `--port` CLI command or via config file default). Thanks for catching this.

yarikoptic:
Thank you @brianhelba for raising this.  Well -- I myself isn't "perfect" in commit messages etc, but I think I am getting better since @ kyleam who works with us on DataLad spent too much on git mailing list I guess ;-) so he is our positive encouragement!
To some degree commits actually help to obtain balance in code-vs-comments ratio, since you could add just a brief reasoning in code (if needed to alert future yourself) and then provide a full nice description in the extended commit message.  Also - code changes, and often code comments (if extended) become simply outdated or even contradicting the code.  That cannot happen with git history comments since they relate to that point in time.

On the opposite -- a "dirty git history" (e.g. - a series of commits in a PR fixing up what they just broke until PR is acceptable and merge) complicates code history navigation and even bugs bisection.  Providing no reasoning for a change (in the code it might be out of place) then starts to require more dialogue with an actual human who did the change to figure out "why" ;-)
So I started to tend to use `--amend` and `rebase -i` more to bring them up into atomic and more or less well described changes.  It also pretty much forces to re-review the changes, and thus often pick up some bugs or typos. 

Altogether, to conclude, I could only repeat my twitter and dandi slack post:

I would recommend to check out review by the current (well -- pretty much "forever") git maintainer Junio: https://github.blog/2020-04-07-celebrating-15-years-of-git-an-interview-with-git-maintainer-junio-hamano/

what made strong influence on me is that  in addition to typical great features decentralized nature of git brings, he brings up the one I had recently quoted in a tweet..:  "that it allows us to completely separate the act of committing and making the result public, and I think that aspect of “distributed”-ness had the biggest impact. It helped imperfect (read: all of us humans) programmers pretend as if they were perfect".

brianhelba:
@yarikoptic Thanks for bringing all this up. I think it's an extremely worthwhile discussion, particularly as we create a codebase that needs to have long-term maintainability. Incoming Friday wall of text...

> To some degree commits actually help to obtain balance in code-vs-comments ratio, since you could add just a brief reasoning in code (if needed to alert future yourself) and then provide a full nice description in the extended commit message.

I tend to feel the opposite. :smiley: IMO, a full description is most useful right next to the code that it describes. Git is most practically useful to manage the logistics of code, rather than as an authority on the semantics of a given chunk of code. A few years ago, I think I felt differently, as Git (particularly the rationale around atomic commits) markets itself as being the solution to questions of "why is this code here?", but practically, I've found that it's very challenging to *rely* primarily on Git to provide this.

> Also - code changes, and often code comments (if extended) become simply outdated or even contradicting the code.

This can certainly happen, but it's a critical pathology that we need to avoid as developers. Regardless of whether a rationale is in an outdated code comment (which I can easily find the original commit for) or within a previous Git message, a future "code archaeologist" can review the history of changes and reconstruct the reasons for the current code, but this should be a last resort.

IMO, if we're going to hold ourselves to one high standard, it should be to have accurate and useful comments in the code as it is *now*, rather than to have a pristine history.

> On the opposite -- a "dirty git history" (e.g. - a series of commits in a PR fixing up what they just broke until PR is acceptable and merge) complicates code history navigation and even bugs bisection.

Agreed, I think commit hygiene is worth paying some attention to. However, histories will unavoidably get dirty over time. Most significantly, changes to indentation / styling made by automated tools, and [file renames](https://stackoverflow.com/questions/29468273/why-git-blame-does-not-follow-renames) are disruptive to history, but we can't be afraid to make these changes because they might break `git blame`. We don't serve the tools, the tools serve us.

> Providing no reasoning for a change (in the code it might be out of place) then starts to require more dialogue with an actual human who did the change to figure out "why" ;-)

Also agreed. I just think this reasoning is better placed in code comments, where it is visible at the time of review as well as obviously visible in the future. With a commit message, you see it at review time, but have to dig for it in the future. Also, future developers need to make a choice to dig through history (even if it's a very clean history), whereas a code comment forces future readers to be confronted with important rationale as they see the code itself.

It's been my experience that most of the time, code that looks bad is actually bad, and should be fixed. However, I've also caused a lot of regressions because I made a "quick fix" to an uncommented line of code that was "obviously wrong", but was actually doing something much more subtle. Do I need to dig through `git blame` every time I make a change, on the unlikely chance that there was actually a good rationale for some weird code? Wouldn't it be better in the weird-but-important code was just documented as such inline?

> So I started to tend to use `--amend` and `rebase -i` more to bring them up into atomic and more or less well described changes. It also pretty much forces to re-review the changes, and thus often pick up some bugs or typos.

Agreed. I do this too, and we should encourage it generally. However, one complicating factor is what to do once a code review starts. Do we squash the changes back into the original branch? This makes it very difficult to re-review the changes and figure out what was changed after the first round of feedback. Again, the tools should serve us. I think a good (though admittedly not ideal for the sake of long-term Git hygiene) solution is to fixup your branch as good as possible before pushing, but then don't rewrite history once you create a PR (which is kind of what Git itself traditionally encourages).

However, I'm less certain in my feelings about how to handle branches in response to PR feedback, and I'd really like to know about others' preferences here.

Issue #82, opened 2020-02-14
Allow copy download link from item


mgrauer:
I posted an upstream issue for this, but it might be better fixed here. Either place is fine, as long as it gets fixed.

https://github.com/girder/girder_web_components/issues/225

satra:
since this will be related to a url for an item, should we figure out where the link should sit?

the ideal link would be something like:

`dandiarchive.org/asset/{id}/`  this could return the metadata about the asset
`dandiarchive.org/asset/{id}/download` could redirect to the actual download location.

where should we implement this?

yarikoptic:
related: #274 (that we had url with selected items reflected; no more) and #275 (guide user to dandi client)

mgrauer:
I've asked @dchiquit to help with this because he got some cycles freed up.

PR #267, opened 2020-04-13
Fixes #262, remove the download link

### LINKED ISSUE #262: Remove download link from pre-refactor UI
When you click on a folder or item, don't show the download link

JellyWang-3:
Fixes #262

### DIFF
--- a/web/src/views/FileBrowser.vue
+++ b/web/src/views/FileBrowser.vue
@@ -54,8 +54,8 @@ const actionKeys = [
       window.open(`${JUPYTER_ROOT}`, '_blank');
     },
   },
-  DefaultActionKeys[1], // Download
-  DefaultActionKeys[2], // Download (zip)
+  // Download for items only
+  DefaultActionKeys[1],
 ];
 
 export default {


yarikoptic:
I thought that we decided to allow per-file downloads, just not bother with .zip's -- multifile downloads should go through `dandi` client.

yarikoptic:
@JellyWang-3: Please add `Fixes` statement into the description, so that issue number becomes a link to jump to to seek more information.  Whenever is in the title - it requires manual url entry or browsing to get there

dchiquit:
This PR does what the issue says it should, so I approved it. It sounds like there is some confusion about what is desired in terms of functionality. @mgrauer can you clarify what, if anything, should be downloadable?
@JellyWang-3 I would refrain from merging until @yarikoptic 's concerns are addressed.

JellyWang-3:
> This PR does what the issue says it should, so I approved it. It sounds like there is some confusion about what is desired in terms of functionality. @mgrauer can you clarify what, if anything, should be downloadable?
> @JellyWang-3 I would refrain from merging until @yarikoptic 's concerns are addressed.

Thanks Daniel for the PR. I am going to ask Mike for clarification because I am kind of confused.

mgrauer:
It wasn't totally clear to me, I thought we wanted to remove all download links. Here are some of the issues:

- downloading a single file (or single-file-item) is ok because it comes directly from S3
- downloading a folder with items or subfolders will create a zip file on the fly, on the Girder server, which consumes server resources, and is prone to breakage from the client's perspective (any network issues they have can break this, and the larger the set of resources, the more likely this is to happen as it will take longer to build and send the zip) and isn't restartable from a browser in the way that browsers allow for resuming an interrupted single file download
- users can download via non-browser clients, but there are many users who cannot use a CLI

It would be nice to support the from-browser-large-nested-downloads, but they are prone to breakage as currently implemented. We have `downloads` as one of the large topics to discuss on the Dandi roadmap.

@JellyWang-3 , can you look into removing the download links only from folders for now, and leave the download link for items, if this is possible?

satra:
@mgrauer - here are some thoughts. unfortunately it's a connected list of things not all directly related to this PR.

objective: we primarily want to prevent download that requires girder to the zipping at this point. since single files come from aws that should be fine.

in the previous version of the gui, when you selected multiple files in a folder, the url in the gui would change to a form with the word `select` inside. the dandi cli can download those files from that `select` url by iterating over them, so essentially directly from s3.  so the behavior of the url has changed on the gui client.

in the world of dandi cli, a user could give it:
- dataset permalink
- a girder mappable url for either a folder or a selection of items
- girder mappable url of a single item

for the single item use-case, it would also be useful for the user to retrieve an http download link, which the user cannot right now.

more generally the download button could be changed to a link that is usable with dandi cli or a direct link for an item. it could display:

```
dandi download <link> for any of the above cases
open <link> in a browser (for single items)
```

JellyWang-3:
I made a quick fix and seems to work ok locally.
I need people to take a closer look at it because I am unsure about if this is what people want.
@mgrauer could you please take a look at it? If I understand correctly, this should only allow download on the items but not the folders.
I will take out the comments later.

yarikoptic:
FWIW, I think this is inline with current "desires" to avoid any bottleneck on the server (individual files should be coming straight from S3 as @mgrauer mentioned above)

satra:
@JellyWang-3  - any chance we can make that download link to allow us to treat it as a hyperlink and "copy the link" using standard browser options . @mgrauer - this has been a todo for a while, and it seems like could be added to this pr easily instead of a new one? but i leave it as your call.

mgrauer:
@JellyWang-3 The functionality looks good to me, I tested in the preview. How about you clean up the comments and let's get this merged. Then you can do https://github.com/dandi/dandiarchive/issues/82 as a follow up task, which is the thing @satra is asking about.

yarikoptic:
> How about you clean up the comments and let's get this merged.

why to cleanup comments here?  if anything to be done is the sequence of 3 commits, some with duplicate messages, is to be squashed into a single commit with a nice description of what and why it was done (e.g. comments from above about avoiding server bottle neck).

JellyWang-3:
> > How about you clean up the comments and let's get this merged.
> 
> why to cleanup comments here? if anything to be done is the sequence of 3 commits, some with duplicate messages, is to be squashed into a single commit with a nice description of what and why it was done (e.g. comments from above about avoiding server bottle neck).

I think Mike wants me to clean up the comments in the code because I had some while I was working on it.

mgrauer:
@JellyWang-3 Can you also squash the commits together into one commit? You will have to force push your branch after doing this. It'll make for a cleaner history as @yarikoptic rightly points out, and it will be a good learning experience if you don't already know how to do it.

JellyWang-3:
> @JellyWang-3 Can you also squash the commits together into one commit? You will have to force push your branch after doing this. It'll make for a cleaner history as @yarikoptic rightly points out, and it will be a good learning experience if you don't already know how to do it.

oh. I see.

web/src/views/FileBrowser.vue
```diff
@@ -54,8 +54,8 @@ const actionKeys = [
       window.open(`${JUPYTER_ROOT}`, '_blank');
     },
   },
-  DefaultActionKeys[1], // Download
-  DefaultActionKeys[2], // Download (zip)
+  DefaultActionKeys[1],
+  // Download for items only
```
brianhelba, line 58: Nearly always, code comments should go above the code they are describing.

PR #272, opened 2020-04-14
Replace dandiset listing vuex store with url-encoded state, update sorting options

### LINKED ISSUE #245: Store Public Dandisets page filter information in the URL
The Public Dandisets page filter information is currently stored in Vuex. Moving it to be encoded in the URL parameters allows us to link directly to a specific page of results. 

All state can be removed from Vuex, the only non-filter information there is `dandisets`, which is only used on that page.

AlmightyYakob:
Closes #245.

This also pulls the sorting options out of `utils` (as it's only being used here), and changes the format a bit.

The other notable change is moving the `dandisets` data property to a computed prop. The main benefit to this is it allows for automatic updates when any dependency (`page`, `sortField`, `sortDir`) changes, and avoids having to call `this.reload()` everywhere, which semantically does the same thing.


web/src/components/DandisetsPage.vue
```diff
@@ -33,7 +41,7 @@
       class="
         mx-12
         my-12"
-      :dandisets="dandisets"
+      :dandisets="dandisets || []"
```
dchiquit, line 44: I would prefer a sensible default be provided rather than having to do this edge case handling here:
```
asyncComputed: {
  dandisets: {
    get() {
     ...
      return ...;
    },
    default: [],
  },
}
```
It's more verbose, but it keeps logic out of the template and presents a cleaner API between the script and the template.

web/src/components/DandisetsPage.vue
```diff
@@ -14,17 +14,25 @@
         Sort By:
       </div>
       <v-chip-group
-        v-model="selection"
+        :value="Number(sortOption)"
```
dchiquit, line 17:
```suggestion
        :value="sortOption"
```
Seems redundant to me

web/src/components/DandisetsPage.vue
```diff
@@ -55,48 +76,86 @@ export default {
...
   },
   computed: {
-    page: {
-      get() {
-        return this.moduleRoot.page;
-      },
-      set(page) {
-        this.changePage({ page });
-      },
+    listingUrl() {
```
dchiquit, line 95: Can you change this to a prop? I think this should be encoded in `MyDandisetsView`, `PublicDandisetsView` et al.

web/src/components/DandisetsPage.vue
```diff
@@ -55,48 +76,86 @@ export default {
...
+        this.sortOption = index;
+        this.sortDir = 1;
+      }
+
+      this.page = 1;
     },
-    reload(payload) {
-      return this.$store.dispatch(`${this.module}/reload`, payload);
+    async getTotalDandisets() {
+      const { headers: { 'girder-total-count': total } } = await girderRest.get(this.listingUrl, { params: { limit: 1 } });
```
dchiquit, line 157: I'm not a huge fan of this extra request every time the component is mounted. What do you think of something like this:
```
asyncComputed: {
  async request() {
    return await girderRest.get(listingUrl, ...);
  },
  computed: {
    dandisets() {
      return this.request.data;
    },
    totalDandisets() {
      return request.headers['girder-total-count'];
    },
  }
}
```
It feels pretty janky to have the whole request object stored like that, but I feel bad making an extra request just to get a number that is already provided on every request.

web/src/components/DandisetsPage.vue
AlmightyYakob: I agree, that's a better way to handle it. I didn't know you could do that.

web/src/components/DandisetsPage.vue
AlmightyYakob: You're right, left over from previous work.

web/src/components/DandisetsPage.vue
AlmightyYakob: I like this idea, I'll be pushing this change soon.

web/src/components/DandisetsPage.vue
AlmightyYakob: I actually don't like the idea of arbitrarily passing the URL into the component as a prop, as I feel like we should be defining our use cases here, but if you think it'll make more sense when we add the search results page I could see it being less of a hassle

web/src/components/DandisetsPage.vue
AlmightyYakob: I actually don't think this'll happen anymore, because of the other change regarding storing the request as a async computed prop, but I like that you can do this.

AlmightyYakob:
Also just a note, it looks like the app bar highlighting (i.e. highlighting `Public Dandisets` or `My Dandisets`) goes away as soon as query params are added to the URL, as I believe it's looking for an exact match.

I went looking for a solution to this, and I found a rabbit hole of proposals/issues for this (as apparently it's a known issue), which ended in [this](https://github.com/vuejs/rfcs/pull/136) PR, merged 11 hours ago as of this comment. So if we want to fix this we'll probably have to come up with a workaround.

web/src/components/DandisetsPage.vue
dchiquit: My vague architectural intuition is that DandisetsPage contains things that are in common between all the listing pages. The URL is different for different pages, so it would be defined in the listing page.

That said, it occurs to me now that this might have to be reswizzled a bit to deal with search requests, since those would involve including a search query. I'm fine with what's here, with the caveat that I might have to reevaluate it soon.

web/src/components/DandisetsPage.vue
dchiquit: In either case, I think the ` || []` can be removed safely now

web/src/components/DandisetsPage.vue
```diff
@@ -55,48 +76,91 @@ export default {
...
+          sort: sortField,
+          sortdir: sortDir,
+        },
+      });
+
+      return {
+        data,
+        headers,
+        status,
+        statusText,
```
dchiquit, line 142: Are these necessary? I don't see them used anywhere.

AlmightyYakob: They're not used anywhere right now, but I could see them being useful, particularly if we ever want to check if the request failed.

PR #278, opened 2020-04-16
Add Dandiset search page

dchiquit:
* All DandisetsPage views will now append a `search` parameter to
queries they make to their listing endpoint, gathered from the URL
parameter `search`. If not present, it is `undefined`.
* Submitting any search field will change the `search` URL parameter to
to the query, and if not already on the Dandiset search page, navigate
there.


girder-dandi-archive/girder_dandi_archive/rest.py
```diff
@@ -102,6 +105,49 @@ def list_dandisets(self, limit, offset, sort):
...
+    def search_dandisets(self, search, limit, offset, sort):
+        # Ensure we are only looking for drafts collection child folders.
+        drafts = get_or_create_drafts_collection()
+        # TODO Currently only searching identifier, name, and description of public dandisets
+        if not search:
+            # Empty search string should return all possible results
+            return Folder().find({"parentId": drafts["_id"]}, limit=limit, offset=offset, sort=sort)
+        return Folder().find(
+            {
+                "$and": [
```
brianhelba, line 123: All MongoDB query terms are implicitly `$and` together, so this shouldn't be necessary. The top level keys can be just `parentId` and `$or`.

web/src/views/HomeView/HomeView.vue
```diff
@@ -50,18 +50,20 @@
       <StatsBar />
     </v-row>
     <v-row>
-      <Search />
+      <div class="full-width grey darken-2 pa-12">
+        <DandisetSearchField :dense="false" />
+      </div>
     </v-row>
```
AlmightyYakob, line 56: This can be accomplished without extra css:
```
    <v-row no-gutters>
      <v-col
        class="grey darken-2 pa-12"
        cols="12"
      >
        <DandisetSearchField :dense="false" />
      </v-col>
    </v-row>
```

Which I've tested locally and looks the same. Obviously if you were to do this you'd also have to remove the css below.

girder-dandi-archive/girder_dandi_archive/rest.py
```diff
@@ -102,6 +105,49 @@ def list_dandisets(self, limit, offset, sort):
         drafts = get_or_create_drafts_collection()
         return Folder().find({"parentId": drafts["_id"]}, limit=limit, offset=offset, sort=sort)
 
+    @access.public
+    @autoDescribeRoute(
+        Description("Search Dandisets")
+        .param("search", "Search Query", paramType="path")
```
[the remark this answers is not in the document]
dchiquit: This is one of the officially supported values for `paramType` according to the documentation, but I'm happy to change it.

web/src/components/DandisetSearchField.vue
```diff
@@ -0,0 +1,68 @@
...
+      :value="$route.query.search"
+      label="Search Dandisets"
+      outlined
+      solo
+      hide-details
+      :dense="dense"
+      background-color="white"
+      color="black"
+      @input="updateSearch"
+      @click:append="performSearch"
```
AlmightyYakob, line 13: This is leftover from using the prop:
```suggestion
```

PR #279, opened 2020-04-17
RF: nwb.json => dandiset_metanwb.json

yarikoptic:
Initially I had missed the fact that it is not the schema we impose on top
of nwb, but rather a schema for dandiset.yaml fields which are derived from
metadata we find in nwb files.  As it is not directly the information from
nwb, I decided to call it metanwb.

I left variable `NWB_SCHEMA` as is since it seems that UI is anyways dandiset specific, but can change that one to `METANWB_SCHEMA` for clearer correspondence. Let me know

This is a follow up to previous renamings in bdf09cfed3d76a7dc73de7bcf82bdd28f984c9f2

### DIFF
--- a/web/src/assets/schema/dandiset_metanwb.json
+++ b/web/src/assets/schema/dandiset_metanwb.json
@@ -1,7 +1,7 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema",
   "type": "object",
-  "title": "NWB",
+  "title": "Dandiset data about metadata from NWB files schema",
   "required": [
     "sex",
     "organism",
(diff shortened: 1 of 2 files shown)


yarikoptic:
don't you merge `master` branch into it once in a while?  or you mean that you have some additional code relying on that path, thus merging wouldn't be enough?

brianhelba:
The differences between `master` and `refactor` are so substantial that I'm afraid this might be lost in the shuffle of a typical merge if we don't look out for it.

yarikoptic:
ah, ok... then I will keep that in mind, and try to not bother with little and not that important changes for now.

brianhelba:
@yarikoptic You can certainly still make any PRs, but it's probably better to make them against `refactor`, instead of `master` for the very near term.

brianhelba:
@yarikoptic I've actually just rewritten this branch (keeping you as the Git author) to be based on and merged into `refactor`.

PR #284, opened 2020-04-17
A refactor for API key component

### LINKED ISSUE #190: Create an API key display component
ATM this would be linked from navbar.

JellyWang-3:
Fixes #190 and clean up the code for existing API key component.

### DIFF
--- a/web/src/components/AppBar/ApiKeyItem.vue
+++ b/web/src/components/AppBar/ApiKeyItem.vue
@@ -1,9 +1,10 @@
 <template>
   <v-list-item>
     <v-list-item-action class="mr-2">
+      <!-- ".stop" prevents closing the parent v-menu when this is clicked -->
       <v-btn
         icon
-        @click="reloadApiKey"
+        @click.stop="fetch"
       >
         <v-icon>mdi-reload</v-icon>
       </v-btn>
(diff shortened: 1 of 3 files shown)


web/src/components/AppBar/ApiKeyItem.vue
```diff
@@ -12,35 +12,68 @@
...
+            limit: 50,
+            sort: 'name',
+            sortdir: 1,
+          },
+        },
+      );
+      let [dandiKey] = data.filter((key) => key.name === 'dandicli');
+      if (!dandiKey) {
+        // create a key
+        const { data2 } = await girderRest.post('api_key', null, {
```
brianhelba, line 63: Does this work? Since you're using destructuring assignment, this expects a property of `girderRest` to be called `data2` (i.e. this is the same as `const data2 = await girderRest.post(...).data2`).

You could do `const { data: data2 }` to avoid a name conflict, but `data2` is a very magic and non-descriptive variable name. If I look at that code in the future, it'd be hard to understand why it's "2" and what information that "2" is trying to deliver to the reader. What if we add more Ajax calls in the future? Will those be `data3`, `data4`, etc.? :smiley: 

Since you don't need the original `data` (from the GET request) down at this point in the code, you could just originally declare `let data`, which would allow you to reassign to the same name here.

web/src/components/AppBar/ApiKeyItem.vue
```diff
@@ -3,7 +3,7 @@
     <v-list-item-action class="mr-2">
       <v-btn
         icon
-        @click="reloadApiKey"
+        @click.stop="fetch"
```
brianhelba, line 6: I realized that the `.stop` here may be worth adding a code comment, explaining why we do this. If you only look at this one component, it's not at all obvious.

Relatedly, see: https://stackoverflow.com/a/41229144

web/src/components/AppBar/ApiKeyItem.vue
JellyWang-3: ok. Got it. Thanks. I was doing a quick fix when cli test failed and yelled about the naming of "data"

web/src/components/AppBar/ApiKeyItem.vue
```diff
@@ -12,35 +13,68 @@
...
+            limit: 50,
+            sort: 'name',
+            sortdir: 1,
+          },
+        },
+      );
+      let [dandiKey] = data.filter((key) => key.name === 'dandicli');
+      if (!dandiKey) {
+        // create a key
+        let { data } = await girderRest.post('api_key', null, {
```
dchiquit, line 64:
```suggestion
        { data } = await girderRest.post('api_key', null, {
```
This line is throwing a linting error because `data` is already declared. I think if you just remove the `let` you can reassign it safely.

web/src/components/AppBar/ApiKeyItem.vue
```diff
@@ -1,9 +1,10 @@
 <template>
   <v-list-item>
     <v-list-item-action class="mr-2">
+      <!-- stop closing the menu if API key refreshed -->
```
brianhelba, line 4: This is good, but since HTML doesn't allow us to put comments inside elements, it may not be clear whether "stop" is referring the effect or the actual `.stop` modifier (and it's both).

What do you think of:
```suggestion
      <!-- ".stop" prevents closing the parent v-menu when this is clicked -->
```

PR #285, opened 2020-04-20
Replace indexes in keys with values

### LINKED ISSUE #240: Do not use array index as ":key" value
Using an array index for `:key` inside a `v-for` defeats the purpose, since the `:key` needs to always map to the same element. After the list is shrunk or reordered, this may not hold. Instead, a unique property of each element should be used for the `:key` value.

There are several places in the code where this should be corrected. At least, this should be applied to the `refactor` branch only, so it's fixed in future development.

AlmightyYakob:
Closes #240.

There's also a lingering use of this in `MetaNode.vue`:

https://github.com/dandi/dandiarchive/blob/c7e5237fef6d0bd65726b4db39ebe46b14b7d128/web/src/components/MetaNode.vue#L95

However, perhaps when making the change to linting, or just at some point during the refactor, the template for this component has been tangled, so additional work will need to be done to fix this, along with that change.

### DIFF
--- a/web/src/views/DandisetLandingView/DandisetLandingView.vue
+++ b/web/src/views/DandisetLandingView/DandisetLandingView.vue
@@ -117,8 +117,8 @@
                 <template v-if="meta.contributors">
                   <v-subheader>Contributors</v-subheader>
                   <v-list-item
-                    v-for="(item, i) in meta.contributors"
-                    :key="i"
+                    v-for="item in meta.contributors"
+                    :key="item.orcid || `${item.name}-${item.roles}`"
                     selectable
                   >
                     <v-list-item-content>{{ item }}</v-list-item-content>
@@ -129,7 +129,7 @@
                     {{ k }}
                   </v-subheader>
                   <v-list-item
-                    :key="k"
+                    :key="`${k}-item`"
                     selectable
                   >
                     <v-list-item-content>
--- a/web/src/components/MetaEditor.vue
+++ b/web/src/components/MetaEditor.vue
@@ -15,8 +15,8 @@
             </template>
             <template v-else>
               <v-alert
-                v-for="(error, i) in errors"
-                :key="i"
+                v-for="error in errors"
+                :key="error.schemaPath"
                 dense
                 type="error"
                 text-color="white"


web/src/views/DandisetLandingView/DandisetLandingView.vue
```diff
@@ -129,7 +129,7 @@
                     {{ k }}
                   </v-subheader>
                   <v-list-item
-                    :key="k"
+                    :key="`${k}-item`"
```
brianhelba, line 132: I'm not sure this helps anything. The key is still solely determined by the position of `item` within `extraFields`, so any changes to absolute positions will confuse the re-rendering.

web/src/views/DandisetLandingView/DandisetLandingView.vue
```diff
@@ -117,8 +117,8 @@
                 <template v-if="meta.contributors">
                   <v-subheader>Contributors</v-subheader>
                   <v-list-item
-                    v-for="(item, i) in meta.contributors"
-                    :key="i"
+                    v-for="item in meta.contributors"
+                    :key="item.orcid || `${item.name}-${item.roles}`"
```
brianhelba, line 121: If a given `item` will always consistently yield the same unique result from this, it will work. I'm not familiar with the schema of `meta.contributors`, is `${item.name}-${item.roles}` alone not guaranteed to yield a unique result?

web/src/views/DandisetLandingView/DandisetLandingView.vue
AlmightyYakob: `extraFields` is an object, so shouldn't the key (`k`) for each item be sufficiently unique?

web/src/views/DandisetLandingView/DandisetLandingView.vue
AlmightyYakob: I'd say it's at least unlikely to yield the same results. Obviously using the `orcid` is the best case scenario, as those are guaranteed to be unique. However if that isn't supplied, I think it's unlikely that two people with the exact same name and list of roles will be supplied.

web/src/views/DandisetLandingView/DandisetLandingView.vue
brianhelba, line 132: One more annoying question :smiley: 

Why not keep this as-is? Adding the `-item` doesn't do anything to improve the keying.

web/src/views/DandisetLandingView/DandisetLandingView.vue
brianhelba: Sounds reasonable.

web/src/views/DandisetLandingView/DandisetLandingView.vue
AlmightyYakob: The `v-subheader` above uses `k` as a key, so keeping it as is results in duplicate keys.

PR #300, opened 2020-04-24
Add end-to-end test for register, login and logout

dchiquit:
End-to-end tests will be stored in the top level `test` directory. Tests
can be run from within `test` with a command like:

```
CLIENT_URL=http://localhost:8092 yarn run test
```

Tests will run in headless mode by default. Prepend `DEBUG=true` to
force the browser window to appear.

Currently there is only one test, `account.test.js`. It tests user
registration, logout, and login, in that order.

I included a helper called `vuetify-xpaths.js` that generates XPaths
that identify several common Vuetify elements. So far, I prefer using
this helper rather than slapping CSS classes on all the elements that
might need to be interacted with, or writing ugly XPaths directly in the
test. All the `vuetify-xpaths` helpers compose with eachother nicely to
generate complex nested locators based on rendered content. There is of
course much room for debate on the implementation.


AlmightyYakob:
I'm not sure how to feel about having the `test` folder at the top level. While technically it is an end-to-end test, it mainly tests the frontend, with the prerequisite that the backend be running (which is also a prerequisite of `yarn run serve`, as otherwise just a blank page loads). My intuition is that having the `test` folder under the `web` folder would make more sense, and would allow shared boilerplate, but if I'm wrong I'd love to hear why.

dchiquit:
That's how I initially had it set up on Medifor, but I ended up changing it to this design. The only configurations shared between `test` and `web` are about half of the `eslintConfig` in `package.json`. Everything else is different because `web` is targeting a browser build and `test` is targeting an unbundled node command. There are not even any dependencies in common. The only thing they have in common is they're both written in JS and installed with `yarn`. 

Conceptually it makes sense to have a single "frontend" bucket, but I found that having everything in the same place just made it more confusing to identify which code belonged to `test` and which to `web`. Having two different locations has made it much easier to manage what goes where, even if we do now have two different `package.json`s in the same project.

test/src/vuetify-xpaths.js
```diff
@@ -0,0 +1,64 @@
...
+      return cssClass.map(c => classAsPredicate(c)).join('');
+    }
+    if (typeof cssClass === 'string') {
+      return `[contains(@class, "${cssClass}")]`;
+    }
+    return '';
+  }).join('');
+}
+
+export function vAvatar(contents, cssClass = undefined) {
```
brianhelba, line 46: TODO: Document this library.

test/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "jest-puppeteer": "^4.4.0",
+    "puppeteer": "^3.0.0"
+  },
+  "eslintConfig": {
+    "root": true,
+    "env": {
+      "node": true
+    },
+    "extends": [
+      "eslint:recommended"
```
brianhelba, line 23: Let's use AirBnB rules, to match what we use with Vue.

test/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "test",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "test": "jest --runInBand"
```
brianhelba, line 6: Would it be helpful for a `test-debug` script here, to config the debugging envvar?

test/src/tests/account.test.js
```diff
@@ -0,0 +1,53 @@
...
+  it('login', async () => {
+    await (await page.waitForXPath(vBtn('Login'))).click();
+
+    await (await page.waitForXPath(vTextField('Username or e-mail'))).type(username);
+    await (await page.waitForXPath(vTextField('Password'))).type(password);
+
+    await (await page.waitForXPath(vBtn(['Login', vIcon('mdi-login')]))).click();
+
+    // the user avatar contains the initials and is only rendered when logged in successfully
+    await page.waitForXPath(vAvatar('MR'));
```
brianhelba, line 51: If possible, could we wrap this in some sort of `expect`..., to indicate it's an assertion?

test/src/vuetify-xpaths.js
brianhelba: We can do this in a separate issue / PR.

test/src/tests/account.test.js
dchiquit: I did you one better: I just added a new module `jest-xpaths.js` which adds some new methods to the `expect(page)` object to make this possible, and also clean up the ugly nested `await (await page.waitForXPath(...))` thing.

test/src/jest-xpaths.js
```diff
@@ -0,0 +1,19 @@
+import 'expect-puppeteer';
+
+// Some extensions to the Jest expect API to better support XPaths
+expect.extend({
+  async toContainXPath(page, xpath) {
+    return page.waitForXPath(xpath)
+      .then(() => ({ pass: true }));
```
brianhelba, line 7: Is there a reason you're not using `await` here, and just returning an object afterwards?

dchiquit: I was thinking `then/catch` would make for more elegant exception handling in the case where the XPath wasn't found, but I forgot to loop back and implement it. I tried it and a normal try/catch seems cleaner to me.

test/src/jest-xpaths.js
```diff
@@ -3,17 +3,27 @@ import 'expect-puppeteer';
 // Some extensions to the Jest expect API to better support XPaths
 expect.extend({
   async toContainXPath(page, xpath) {
-    return page.waitForXPath(xpath)
-      .then(() => ({ pass: true }));
+    try {
+      await page.waitForXPath(xpath);
+      return { pass: true }
```
brianhelba, line 8: Do we want to support `expect(page).not.toContainXPath`? If so, [the API requires](https://jestjs.io/docs/en/expect#custom-matchers-api):

> And when `pass` is true, `message` should return the error message for when `expect(x).not.yourMatcher()` fails.

I don't have a preference.

dchiquit: I was not aware of that pattern.
I think we should not support it for now. The behavior for `expect(page).not.toContainXPath` is very dependent on timeouts. If the timeout is too short, it will fail unexpectedly when the condition isn't met in time, but if it's too long, it can severely impact test runtime. Currently I think that timeout is globally configured to 30s.

I believe this will function as it currently is written, it just won't have a useful error message. If we run into a solid use case for using that pattern, it can be enabled easily, and we can tune the timeouts to something reasonable for that use case. Doing that now seems like premature optimization.

brianhelba: Makes sense.

PR #311, opened 2020-04-28
309 search usability

### LINKED ISSUE #309: Document search functionality
Right now there is no indication of what the search function does. We should include placeholder text or some text when there are no search results.

dchiquit:
* Search fields now contain the placeholder `Search Dandisets by name,
description, or identifier` to document intended usage
* Dandisets pages will now expand the search bar to fill remaining space
in the toolbar (so that the full placeholder text can be seen)
* Dandisets pages will now show a spinner while waiting for content to
load
* Dandisets pages now provide a slot for the text to display when there
are no dandisets to list
* My Dandisets will now say `You haven't created any dandisets yet`
* Search Dandisets will now provide some concise assistance if you have
no search results
* Some unrelated files were linted

fixes #309

### DIFF
--- a/web/src/components/DandisetsPage.vue
+++ b/web/src/components/DandisetsPage.vue
@@ -48,8 +48,8 @@
       >
         <v-col>
           <v-progress-circular
-            indeterminate
             v-if="!dandisets"
+            indeterminate
           />
           <slot
             v-else
--- a/web/src/components/DandisetsPage.vue
+++ b/web/src/components/DandisetsPage.vue
@@ -33,12 +33,31 @@
           </v-icon>
         </v-chip>
       </v-chip-group>
-      <DandisetSearchField />
+      <DandisetSearchField class="flex-grow-1" />
     </v-toolbar>
     <DandisetList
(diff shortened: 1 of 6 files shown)


web/src/components/DandisetsPage.vue
```diff
@@ -117,7 +136,7 @@ export default {
       return { page, sortOption, sortDir };
     },
     dandisets() {
-      return this.dandisetRequest ? this.dandisetRequest.data : [];
+      return this.dandisetRequest ? this.dandisetRequest.data : undefined;
```
AlmightyYakob, line 139: Why was this changed?

dchiquit: To enable the loading icon to work properly. `undefined` means that no request has been made yet, so it should display the spinner. `[]` means that a request has been made and there's nothing to show, so the `no-content` slot is rendered. Anything else means there's dandisets, so it renders as it used to.

AlmightyYakob: Gotcha

Issue #308, opened 2020-04-27
Evolve the Dandi CLI at the same time as the web UI refactor update


mgrauer:
Figure out what the interaction points are between the CLI and the UI, and make sure we update the CLI for those interaction points at the same time as we update the UI so that they stay in sync.

The list of interaction points should be captured [here](https://github.com/dandi/dandi-cli/blob/master/dandi/download.py#L13).

brianhelba:
https://github.com/dandi/dandi-cli/blob/master/dandi/download.py#L58

satra:
more specifically the types of URLs that were supported: 

https://github.com/dandi/dandi-cli/blob/master/dandi/download.py#L16

mgrauer:
After #303 is done, @AlmightyYakob can point @yarikoptic to the new URLs.

satra:
this should be resolved: https://github.com/dandi/dandi-cli/issues/92

mgrauer:
@AlmightyYakob can you provide a mapping of the old to new URLs? please correct me if I'm missing something.

AlmightyYakob:
I believe the URLs should stay the same, except for an an added `file-browser/` in the path. If it was:

 https://gui.dandiarchive.org/#/folder/5e5593cc1a343161ff7c5a92

Now it'd be 

 https://gui.dandiarchive.org/#/file-browser/folder/5e5593cc1a343161ff7c5a92

Which on girder would be:

https://girder.dandiarchive.org/#user/5da4b8fe51c340795cb18fd0/folder/5e5593cc1a343161ff7c5a92

Does this answer your question?

mgrauer:
@yarikoptic will you be able to make these changes? Does what @AlmightyYakob said in the previous comment give you everything you need?

AlmightyYakob:
> @yarikoptic will you be able to make these changes? Does what @AlmightyYakob said in the previous comment give you everything you need?

@yarikoptic Can we close this issue?

yarikoptic:
yes, AFAIK https://github.com/dandi/dandi-cli/pull/94 should have addressed it

PR #313, opened 2020-04-30
Switch to refactored UI

mgrauer:
As there are no longer any release blockers on the UI side, merging refactor into master should move us to the refactored UI. We want to wait for the CLI to be changed to be in sync though, which is #308.


satra:
@mgrauer - now that @yarikoptic has updated CLI, we may be able to make this switch now. are there any blockers left? 

also wdyt of dandiarchive.org pointing directly to the data portal, now that this refactor is done.

yarikoptic:
note that I have not released `dandi-cli` yet.  I also want to explicitly bring your attention to https://github.com/dandi/dandiarchive/issues/316 (not a blocker here) so we could enhance `dandi-cli` so it could alert users when a newer version needs to be used etc.

mgrauer:
>also wdyt of dandiarchive.org pointing directly to the data portal, now that this refactor is done.

i think this would break the redirection service, and so we couldn't support our dandiset permalinks.

satra:
> i think this would break the redirection service, and so we couldn't support our dandiset permalinks.

the top level redirect (https://github.com/dandi/redirector/blob/master/serve.py#L18) is separate from the permalink redirect (https://github.com/dandi/redirector/blob/master/serve.py#L33).

mgrauer:
Got it, this seems reasonable then.

You are suggesting the following?

- www.dandiarchive.org would redirect to the about static page
- dandiarchive.org would redirect to gui.dandiarchive.org, which could get to the about static page from the `about` link
- the permalink for dandisets will still go to the redirector, which will query the backend and redirect to the place in the new gui


Let's also tweak the data portal link back to the gui (I'll do this in a separate PR on the static site)

mgrauer:
>Let's also tweak the data portal link back to the gui (I'll do this in a separate PR on the static site)

I have this PR, https://github.com/dandi/dandi.github.io/pull/24, but I see it is not necessary as the `dandiset/drafts` [route will be redirected to the GUI](https://github.com/dandi/redirector/blob/master/serve.py#L27-L31).

Is there any reason to support this `dandiset/drafts` route now or can we remove that from the redirector? Especially as it isn't going to a listing of drafts and we no longer have a listing of dandisets that are only drafts.

PR #321, opened 2020-05-04
Remove old search mode

### LINKED ISSUE #315: Remove custom search mode
There was a custom search mode implemented that is no longer called by the front end. If we remove this custom search mode, we can also remove the tests for it, and remove the [`pytest-forked` package](https://github.com/dandi/dandiarchive/blob/master/girder-dandi-archive/tox.ini#L13) and the [`--forked` parameter](https://github.com/dandi/dandiarchive/blob/master/girder-dandi-archive/tox.ini#L76-L78) from the tests, which should speed up running the tests.

AlmightyYakob:
Fixes #315.


mgrauer:
This LGTM, but I'd want someone with a working dev ENV to test before merging.

@dchiquit could you QA and make sure the site still works for you locally, and that you can search?

AlmightyYakob:
> This LGTM, but I'd want someone with a working dev ENV to test before merging.

I agree Daniel should test this, I probably should have specified I also tested this locally without issue, just for transparency.

PR #322, opened 2020-05-04
Add compound indexes for dandiset name and description

### LINKED ISSUE #314: Possible performance improvements for search
While looking around at the [search endpoint](https://github.com/dandi/dandiarchive/blob/refactor/girder-dandi-archive/girder_dandi_archive/rest.py#L121-L141) in the refactor branch, it seems like we are missing a couple of indexes that could be useful. There is a [compound index](https://github.com/dandi/dandiarchive/blob/refactor/girder-dandi-archive/girder_dandi_archive/__init__.py#L24) for `(meta.dandiset.identifier, parentId)`, but not for `(meta.dandiset.name, parentId)` or `(meta.dandiset.description, parentId)`.

AlmightyYakob:
Fixes #314.

I'm not sure if this actually improves the search time, testing locally didn't seem to have much of a difference, but it's probably still good to have these extra indexes anyway. I'm not sure if there's any downsides to this.

### DIFF
--- a/girder-dandi-archive/girder_dandi_archive/__init__.py
+++ b/girder-dandi-archive/girder_dandi_archive/__init__.py
@@ -21,7 +21,9 @@ class DandiArchivePlugin(GirderPlugin):
         # Mongo is not guaranteed to use a singular index when executing the query.
         # meta.dandiset.identifier is specified first so that it is available as an
         # index for other queries that might need it.
+        Folder().ensureIndex(([("meta.dandiset.name", 1), ("parentId", 1)], {}))
         Folder().ensureIndex(([("meta.dandiset.identifier", 1), ("parentId", 1)], {}))
+        Folder().ensureIndex(([("meta.dandiset.description", 1), ("parentId", 1)], {}))
 
         Setting().collection.update(
             {"key": DANDISET_IDENTIFIER_COUNTER}, {"$setOnInsert": {"value": 1}}, upsert=True,


dchiquit:
You can actually test index usage directly: https://docs.mongodb.com/manual/tutorial/measure-index-use/
A command like this in the mongo shell will show you how many times the `folder` indices have been used:
```
db.folder.aggregate([{$indexStats:{}}, {$project:{name:1, accesses:{ops: 1}}}])
```
I did these before and after a search, and none of your indices incremented, so they definitely aren't being used. The search function currently uses a `$regex` to do partial string matching, which can't utilize indexing at all, so that makes sense.
My understanding is that there is a small constant time cost to maintaining the index whenever a collection is modified, and rebuilding indices completely becomes more costly the more records are present. For DANDI I think our data size will always be manageable, so there is effectively no real cost to having unused indices. These indices are effectively dead code though, and they will only clutter the index management process in the future when we have to try to identify what they were used for and if they are effective.

mgrauer:
@dchiquit Thanks for the detailed follow up.

I had requested this issue based on the [dandiset identifier index](https://github.com/dandi/dandiarchive/blob/master/girder-dandi-archive/girder_dandi_archive/__init__.py#L20) you added, and was thinking to cover these other two metadata sub documents with an index, but that doesn't make sense to do given what you've found.

Will that identifier index be used by our current search? If not, maybe the issue should transition to removing the identifier index rather than adding indexes!

dchiquit:
Nope, our current search is currently completely unindexed. It was a very succinct implementation that would be very easy to fix later, at the cost of being basic and potentially slow. There are ways to index for text search, but its complicated for a number of reasons and I didn't want to implement it only to throw it out when we have the comprehensive search discussion.

The existing index was put in for the get dandiset endpoint. It works perfectly there, I just tested it again for sanity.

mgrauer:
Makes sense, I'll close the issue. Sorry for the goose chase @AlmightyYakob .

PR #323, opened 2020-05-04
Replace JSON rendering with better looking component

### LINKED ISSUE #206: Don't display Dandiset metadata as json
Make the display of metadata on the Dandiset landing page nicer, with keys and nested values, instead of a json dump.

AlmightyYakob:
Fixes #206.

This involves adding a piece of metadata to the schema, `listingKey`. This is to be specified at the same level as a `required` field, and specifies which of the properties of an object should be used to represent it at a top level. I'm open to suggestions about renaming this field to whatever best suits it, but I just chose a name to get moving on it. @satra It would be good for you to confirm that the `listingKey` I specified on each property is actually what it should be. Most of the time it's `name`, but does vary.

This also adds a component, `DandisetLandingView/ListingComponent.vue`, which is recursive and is only used in the `DandisetLandingView` page.

### DIFF
--- a/web/src/views/DandisetLandingView/ListingComponent.vue
+++ b/web/src/views/DandisetLandingView/ListingComponent.vue
@@ -26,6 +26,15 @@
           </v-expansion-panel>
         </v-expansion-panels>
       </template>
+      <template v-else-if="schema.items.type === 'array'">
+        <!-- May not work. Not tested in a real scenario. -->
+        <ListingComponent
+          v-for="item in data"
+          :key="item"
+          :schema="schema.items"
+          :data="item"
+        />
+      </template>
       <template v-else>
         <b>{{ schema.title }}</b>: {{ data.join(', ') }}
       </template>
--- a/web/src/views/DandisetLandingView/ListingComponent.vue
+++ b/web/src/views/DandisetLandingView/ListingComponent.vue
@@ -5,10 +5,10 @@
         <v-expansion-panels>
           <v-expansion-panel
             v-for="item in data"
-            :key="item[schema.items.listingKey]"
+            :key="item[primaryKey]"
           >
             <!-- item is an object -->
-            <v-expansion-panel-header>{{ item[schema.items.listingKey] }}</v-expansion-panel-header>
+            <v-expansion-panel-header>{{ item[primaryKey] }}</v-expansion-panel-header>
             <v-expansion-panel-content>
               <v-list>
                 <v-list-item
@@ -63,6 +63,20 @@ export default {
       // The data at the matching level of schema
       type: [Object, Number, String, Array],
       required: true,
+    },
+  },
+  computed: {
+    primaryKey() {
+      switch (this.schema.title) {
+        case 'Access':
+          return 'status';
+        case 'Publications':
+          return 'url';
+        case 'Organism':
+          return 'species';
+        default:
+          return 'name';
+      }
     },
   },
 };
--- a/web/src/views/DandisetLandingView/ListingComponent.vue
+++ b/web/src/views/DandisetLandingView/ListingComponent.vue
@@ -1,0 +1,69 @@
+<template>
+  <div>
+    <template v-if="schema.type === 'array'">
+      <template v-if="schema.items.type === 'object'">
+        <v-expansion-panels>
+          <v-expansion-panel
+            v-for="item in data"
+            :key="item[schema.items.listingKey]"
+          >
+            <!-- item is an object -->
+            <v-expansion-panel-header>{{ item[schema.items.listingKey] }}</v-expansion-panel-header>
+            <v-expansion-panel-content>
+              <v-list>
+                <v-list-item
+                  v-for="(value, key) in item"
+                  :key="key"
+                >
+                  <!-- value's type matches that specified at schema.items.properties[key].type -->
+                  <ListingComponent
+                    :schema="schema.items.properties[key]"
+                    :data="value"
+                  />
+                </v-list-item>
+              </v-list>
+            </v-expansion-panel-content>
+          </v-expansion-panel>
+        </v-expansion-panels>
+      </template>
+      <template v-else>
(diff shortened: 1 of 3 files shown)


satra:
@AlmightyYakob - for now this is great. the `listingKeys` are fine and we can always change those in the short term.

mgrauer:
The UI side looks good, much improved!, but does this make sense to add to the schema, or should we put it in the UI code somewhere, this listing key? This relates to the MVC separation that we had talked about before. Is `listingKey` something that describes the data or describes how the data should be visualized in this specific case? I want to be careful about polluting the schema.

satra:
@mgrauer - i would be ok in the short term of adding this into the schema, but with pending schema updates to refactor the component based on the new schema. we will want to do a few things regarding customized display where contributors would look more like zenodo lists, add an ordering index to contributors, etc.,., and we will need to generate a citation. so we will likely have to create a DLP component that is quite custom and we can move some of that information out of the schema.

all of that is to say that let's keep this to improve the display and then figure out the better path forward alongside the revised schema.

AlmightyYakob:
> The UI side looks good, much improved!, but does this make sense to add to the schema, or should we put it in the UI code somewhere, this listing key? This relates to the MVC separation that we had talked about before. 

IMO it would be messy to specify this in the code, as we'd basically have to duplicate the schema for the sole purpose of specifying this data.

> Is `listingKey` something that describes the data or describes how the data should be visualized in this specific case? I want to be careful about polluting the schema.

I think it's fair to say that this describes the data, as it basically specifies the "Primary Key" of an object. I'd actually say `primaryKey` would be a better name than `listingKey`, as it better describes the semantics of it.

mgrauer:
Given that @satra is ok with this, and he can represent for the schema side of things, I withdraw that objection. I would ask a related question: how are we versioning the schema or representing the stability of the schema API? It might be fine to change/break it for now, but get some idea of API promises on the horizon.

@AlmightyYakob I get your point, that is reasonable for now too. Maybe you should change the name to `primaryKey`? Is there a way to make it a required property/key, given that you want to call it a primary key?

My conceptual QA is done, and this generally passes, but we still need a code QA.

satra:
> how are we versioning the schema or representing the stability of the schema API

none at the moment. hence the need to actually fix some of those things before we move forward. i am waiting on some convergence of things from the biccn side as well and then will release a versioned data model.

AlmightyYakob:
> @AlmightyYakob I get your point, that is reasonable for now too. Maybe you should change the name to `primaryKey`? Is there a way to make it a required property/key, given that you want to call it a primary key?

Yeah I can change the name, although I'm not sure what can be done to enforce that this key exists on sub-schemas with an "object" type. Since this field isn't a property itself, but is just extra metadata, the JSON Schema validator ignores it because it's not part of the JSON Schema spec. I think this would be best left as an addtional task. Should I make an issue for it?

mgrauer:
Make an additional task, let's not hold up this PR with it. I know we will want to do some DLP refactoring and specializing, so we can think about this issue as a part of that work.

web/src/views/DandisetLandingView/ListingComponent.vue
```diff
@@ -0,0 +1,69 @@
...
+                    :data="value"
+                  />
+                </v-list-item>
+              </v-list>
+            </v-expansion-panel-content>
+          </v-expansion-panel>
+        </v-expansion-panels>
+      </template>
+      <template v-else>
+        <b>{{ schema.title }}</b>: {{ data.join(', ') }}
```
dchiquit, line 30: I assume an array of arrays doesn't exist in the schema?

AlmightyYakob: It seems like `ethicsApprovals` might, so I'll add the case. It shouldn't be too hard.

AlmightyYakob:
> I too am not fond of adding presentational data to the schema. `listingKey` is exclusively used in the `ListingComponent`, which makes it easy to define in a smaller, isolated scope rather than a global schema definition.
> 
> Rather than redefining the entire schema, I recommend including a new property `field` on `ListingComponent` so it knows when it's rendering an `organism` vs. a `language` vs. a `sponsors`, etc. Then you can just define a computed property like so:
> 
> ```
>   listingKey() {
>     switch(this.field) {
>       case 'access':
>         return 'status';
>       case 'publications':
>         return 'url';
>       case 'organism':
>         return 'species';
>       default:
>         return 'name';
>     }
>   }
> ```
> 
> This doesn't seem that gross to me. It even solves the problem of `listingKey` not being defined on sub-schema metadata. If `name` is not sufficient for those cases, you can just add that case next to the top level fields (assuming no collisions). It also won't require a schema change if we want to change how the frontend is rendered in the future.
> 
> I don't feel good about modifying the schema for the benefit of the presentational layer as more than a temporary bandaid. If that's all it is I'm OK with it, but I think my solution is cleaner.

I like this idea, I'll try that. The only caveat is that if we update the schema we might have to update this computed prop as well, but I'm okay with that.

AlmightyYakob:
Also like you said @dchiquit, this would become a problem if there's field name collisions (e.g. two properties `access` that have different primary keys, and occur at different places in the schema). @satra Do you think this is likely to happen?

web/src/views/DandisetLandingView/ListingComponent.vue
AlmightyYakob: I couldn't find any examples of an array within an array, so I just added a placeholder.

satra:
> this would become a problem if there's field name collisions (e.g. two properties access that have different primary keys, and occur at different places in the schema). @satra Do you think this is likely to happen?

@AlmightyYakob - i very much doubt that this will ever happen.

mgrauer:
Great, this looks a lot nicer than what was there before and the separation of concerns seems reasonable. Nice going @AlmightyYakob and @dchiquit .

PR #324, opened 2020-05-04
E2E test fixes and improvements

brianhelba:


test/src/tests/account.test.js
```diff
@@ -8,19 +7,18 @@ import {
...
-  const email = `${username}@kitware.com`;
+  const email = `${username}@dandi.test`;
   const password = 'password'; // Top secret
 
+  beforeAll(async () => {
+    await page.goto(CLIENT_URL);
+  });
+
   it('register', async () => {
+    expect.assertions(9);
```
dchiquit, line 20: My understanding is that this statement is for testing expects in callbacks. The `await`s effectively de-asynchronize these tests, so `expect.assertions` doesn't provide any value that I can see. Does the jest eslint plugin require it/generate it automatically? I would rather disable that for this project.

brianhelba: Agreed.

PR #327, opened 2020-05-05
Create a separate jest-puppeteer-vuetify package

brianhelba:


test/jest-puppeteer-vuetify/src/jest-xpaths.js
```diff
@@ -1,7 +1,7 @@
 import 'expect-puppeteer';
```
dchiquit, line 1: I think you can actually remove the dependency on `expect-puppeteer`. It adds some puppeteer bindings to the jest expect API, so you can do stuff like `await expect(page).toClick(aButton);`. We never actually use expect anywhere in this module, even though `jest-xpaths` adds bindings that are stylistically similar.

I think having a peer dependency on `expect-puppeteer` is sufficient.

test/jest-puppeteer-vuetify/src/index.js
```diff
@@ -0,0 +1,4 @@
+// Import to register with expect-puppeteer as a side effect
+import './jest-xpaths';
```
dchiquit, line 2: I'm not sure about how best to organize these. `jest-xpaths` (perhaps ill-named) adds some puppeteer bindings that make XPaths easier. `vuetify-xpaths` generates XPaths for vuetify components. The two clearly synergize and for projects that are set up the way ours is, they will both be needed all the time.

However, both could be useful on their own. `vuetify-xpaths` would be equally useful for use with Selenium, but it wouldn't require `jest-xpaths` at all. `jest-xpaths` would be hypothetically useful for any Puppeteer+Jest tests that wanted to use XPaths cleanly (although that seems less realistic to me).

For the scope of this project, I think this organizational scheme is fine. If we want other projects to start using it though, we will eventually have to consider this. It may be better to split them up now, or we could just cross that bridge when we come to it.

brianhelba: Agreed.

In the event we do split them, I think `jest-xpaths` would likely depend on `vuetify-xpaths`.

PR #328, opened 2020-05-06
Add more end-to-end tests

dchiquit:
Add the following tests:
* registerDandiset - Tests creating a new dandiset
* stats - Tests that stats increment correctly
* search - Tests the basic search functionality

Also add a new flavor of helper, pages. Pages are stored in
`test/src/pages/`. The contain methods that interact with a certain view
in the app. For example, `homePage` contains the methods `getStat` and
`search`, as those are the only two actions available on the home page.
The methods in `searchPage` will need to be fleshed out as we add more
tests for search filtering/paging.


test/src/pages/searchPage.js
```diff
@@ -0,0 +1,32 @@
...
+} from '../vuetify-xpaths';
+
+
+/**
+ * Searches for dandisets using the search bar on the search page
+ *
+ * @param {String} query
+ */
+export async function search(query) {
+  await expect(page).toFillXPath(vToolbar() + vTextField(), query);
```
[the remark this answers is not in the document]
AlmightyYakob: I'm not too familiar with XPaths, how are these different?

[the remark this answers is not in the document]
dchiquit: This is all documented in the README which did not exist at the time I left this comment.
```
<v-toolbar>
  <v-text-field />
</v-toolbar>
```
The first XPath will locate the <v-toolbar> element, while the second will locate the `v-text-field> element.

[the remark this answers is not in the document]
AlmightyYakob: That makes sense, thank you for the explanation.

test/src/tests/search.test.js
```diff
@@ -0,0 +1,51 @@
...
+    await homePage.search(name);
+
+    // there should only be one result, the dandiset we just created
+    const results = await searchPage.getSearchResults();
+    expect(results).toHaveLength(1);
+    expect(results[0].name).toStrictEqual(name);
+  });
+
+  it('does not find a dandiset that does not exist', async () => {
+    // search for a nonexistent name
```
brianhelba, line 35: This case (and several others) does not start with a page navigation, making it dependent on the success of the previous test. I think these sort of inter-test dependencies should be avoided if possible, so test failures are easier to diagnose.

test/src/tests/search.test.js
```diff
@@ -0,0 +1,51 @@
...
+describe('dandiset search page', () => {
+  const name = `test${uniqueId()}`;
+  const description = `Description ${uniqueId()}`;
+
+  beforeAll(async () => {
+    await page.goto(CLIENT_URL);
+    await registerNewUser();
+
+    // create a new dandiset that will appear in search results
+    await registerDandiset(name, description);
```
brianhelba, line 20: I worry that this shared setup is starting to tread close to the inter-test shared state that [the `jest/no-hooks` ESLint rule](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-hooks.md) is trying to prevent. In this case, I think it's acceptable, since none of the test cases mutate the server state, but I wonder what you think in general.

test/src/tests/stats.test.js
```diff
@@ -0,0 +1,55 @@
...
+  CLIENT_URL,
+  registerNewUser,
+  registerDandiset,
+  uniqueId,
+} from '../util';
+import * as homePage from '../pages/homePage';
+
+describe('home page stats', () => {
+  beforeAll(async () => {
+    await Promise.all([
```
brianhelba, line 11: Would it be helpful to have a helper like
```
async function loadPage(pageUrl) {
  return Promise.all([
    // TODO: Handle the case where CLIENT_URL ends with a "/"
    page.goto(`{CLIENT_URL}/#{pageUrl}`),
    // wait until there are no active network connections for 500ms
    page.waitForNavigation({ waitUntil: 'networkidle0' }),
  ]);
}
```

Right here, it could be used as simply `await loadPage('/')`.

test/src/util.js
```diff
@@ -14,23 +15,26 @@ export function uniqueId() {
...
-  await (await page.waitForXPath(vBtn('Create Account'))).click();
+  await expect(page).toClickXPath(vBtn('Create Account'));
 
-  await (await page.waitForXPath(vTextField('Username'))).type(username);
-  await (await page.waitForXPath(vTextField('Email'))).type(email);
-  await (await page.waitForXPath(vTextField('First Name'))).type('Mister');
-  await (await page.waitForXPath(vTextField('Last Name'))).type('Roboto');
-  await (await page.waitForXPath(vTextField('Password'))).type(password);
-  await (await page.waitForXPath(vTextField('Retype password'))).type(password);
+  await expect(page).toFillXPath(vTextField('Username'), username);
```
brianhelba, line 27: I thought the rule was to not use assertions / `expect` within these helpers, and instead have separate cases to directly test these functionalities?

test/src/pages/homePage.js
```diff
@@ -0,0 +1,22 @@
+import { vIcon, vTextField } from 'jest-puppeteer-vuetify';
+
+/**
+ * Gets the current value of the given stat from the home page
+ *
+ * @param {String} stat the stat to check
+ */
+export async function getStat(stat) {
```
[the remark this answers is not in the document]
dchiquit: This is an old habit from maintaining a much larger suite of end to end tests on a much larger app. For larger projects it's definitely useful to keep that separation of duties consistent so you always know where to find helper methods without digging through the entire test catalog. 
In the case of the home page, it's probably overkill. This is probably the only test we will ever need that involves the home page. For the search page, I think there is definite value in having `searchPage` defined now, as there are several more search tests that would be useful.

test/src/tests/search.test.js
dchiquit: I think if there is a place to fudge this rule, it would be browser tests. Test setup is generally done through the UI, which is good, but also can be very slow. Having to to do lengthy setup steps for lots of really granular assertions can severely impact run time.
In this case, I think you're right, I need to refactor these to avoid intertest dependencies. Jest almost seems to encourage that pattern, but I should have known better.

test/src/tests/stats.test.js
dchiquit: Quite possibly. I really don't like the `Promise.all` syntax that's required to effectively use `waitForNavigation`, so a helper would make that less painful. However I'm still not 100% sold that this `networkidle0` thing is a good idea, I just wanted to try it out since puppeteer offered it.
I think a helper would still be useful though, I will add one.

test/src/util.js
dchiquit: I personally find the `expect(page).toFillXPath(...)` syntax to be cleaner and more legible than the `await (await page.waitForXPath(...)).type(...)` syntax. In this case, it has the exact behavior I am looking for: put some words into a text field, or else throw an error if it doesn't work for some reason.
I agree it's misleading to have that `expect` statement in a helper method that isn't testing anything. When I read this it's pretty clear to me that `toFillXPath` is being used to modify the page, which I think is excusable. Requiring that all code not directly in a test method use none of the `jest-xpath` helpers sounds reasonable on paper, but obnoxious and arbitrary in practice.

PR #331, opened 2020-05-07
Pass route to login view for return

### LINKED ISSUE #306: Redirect to original page after login or register

AlmightyYakob:
Fixes #306.


AlmightyYakob:
@dchiquit 7c65258 should fix that issue, you should be able to test once the preview is live

PR #338, opened 2020-05-13
Dandiset permissions

### LINKED ISSUE #335: dandiset permissions endpoints
Provide a backend for #334.

This is a simplified permission model: users are either owners or not of dandisets. The original creator is the owner, can add and remove owners.

Create endpoints that support this: 
- list owners, given a dandiset identifier
- remove owners, given a dandiset identifier and list of owners
- add owners, given a dandiset identifier and list of owners

This doesn't have to map to three endpoints exactly, as long as this functionality is supported and the UI in #334 can call the right endpoints.

AlmightyYakob:
Fixes #335.

Draft PR because I'm still trying to get tests working for this, but am running into some errors. Once I get those figured out I'll push those changes.

### DIFF
--- a/girder-dandi-archive/girder_dandi_archive/rest.py
+++ b/girder-dandi-archive/girder_dandi_archive/rest.py
@@ -152,9 +152,8 @@ class DandiResource(Resource):
 
         # Only work with admin level users, removing non-admins from the ACL
         user_id_to_level = {
-            str(user["id"]): user["level"]
-            for user in Folder().getFullAccessList(dandiset)["users"]
-            if user["level"] == AccessType.ADMIN
+            str(user["id"]): AcessType.ADMIN
+            for user in get_dandiset_owners(dandiset)
         }
 
         for owner in owners:
--- a/girder-dandi-archive/girder_dandi_archive/rest.py
+++ b/girder-dandi-archive/girder_dandi_archive/rest.py
@@ -4,7 +4,7 @@ from girder.api import access
 from girder.api.describe import autoDescribeRoute, describeRoute, Description
 from girder.api.rest import Resource
 from girder.constants import AccessType, TokenScope
-from girder.exceptions import RestException
+from girder.exceptions import RestException, ValidationException
 from girder.models.collection import Collection
 from girder.models.folder import Folder
 from girder.models.setting import Setting
@@ -17,6 +17,7 @@ from .util import (
     find_dandiset_by_identifier,
     get_dandiset_owners,
     get_or_create_drafts_collection,
+    validate_user,
 )
 
 
@@ -118,6 +119,9 @@ class DandiResource(Resource):
         }
 
         for owner in owners:
+            if not validate_user(owner):
+                raise ValidationException("All owners must be valid user objects.")
+
             user_id_to_level[owner["_id"]] = AccessType.ADMIN
 
         final_users = [
@@ -155,6 +159,9 @@ class DandiResource(Resource):
         }
 
         for owner in owners:
+            if not validate_user(owner):
+                raise ValidationException("All owners must be valid user objects.")
+
             # Prevents a KeyError if owner isn't an existing owner
             user_id_to_level.pop(owner["_id"], None)
 
--- a/girder-dandi-archive/girder_dandi_archive/rest.py
+++ b/girder-dandi-archive/girder_dandi_archive/rest.py
@@ -3,18 +3,21 @@ import re
 from girder.api import access
 from girder.api.describe import autoDescribeRoute, describeRoute, Description
 from girder.api.rest import Resource
-from girder.constants import TokenScope
+from girder.constants import AccessType, TokenScope
 from girder.exceptions import RestException
 from girder.models.collection import Collection
 from girder.models.folder import Folder
 from girder.models.setting import Setting
 from girder.models.user import User
+from girder.models.group import Group
 
 from .util import (
     DANDISET_IDENTIFIER_COUNTER,
     DANDISET_IDENTIFIER_LENGTH,
+    find_dandiset_by_identifier,
     get_or_create_drafts_collection,
     dandiset_identifier,
+    get_or_create_dandiset_group,
 )
 
 
@@ -24,11 +27,9 @@ class DandiResource(Resource):
 
         self.resourceName = "dandi"
         self.route("GET", (":identifier",), self.get_dandiset)
-
         self.route("GET", (":identifier", "owners"), self.get_dandiset_owners)
         self.route("PUT", (":identifier", "owners"), self.add_dandiset_owners)
         self.route("DELETE", (":identifier", "owners"), self.remove_dandiset_owners)
-
         self.route("GET", ("user",), self.get_user_dandisets)
         self.route("GET", ("search",), self.search_dandisets)
         self.route("GET", (), self.list_dandisets)
@@ -91,51 +92,66 @@ class DandiResource(Resource):
     )
     @dandiset_identifier
     def get_dandiset_owners(self, identifier, params):
-        # Ensure we are only looking for drafts collection child folders.
-        # drafts = get_or_create_drafts_collection()
-        # doc = Folder().findOne({"parentId": drafts["_id"], "meta.dandiset.identifier": identifier})
-        # if not doc:
-        #     raise RestException("No such dandiset found.")
-        # return doc
-
-        # TODO:
-        pass
+        return Folder().getFullAccessList(find_dandiset_by_identifier(identifier))["users"]
 
     @access.user
-    @describeRoute(
-        Description("Add Dandiset Owners").param(
-            "identifier", "Dandiset Identifier", paramType="path"
+    @autoDescribeRoute(
+        Description("Add Dandiset Owners")
+        .param("identifier", "Dandiset Identifier", paramType="path")
+        .jsonParam(
+            "owners",
+            "A JSON list of girder users to add as owners.",
+            paramType="body",
+            requireArray=True,
         )
     )
     @dandiset_identifier
-    def add_dandiset_owners(self, identifier, params):
-        # # Ensure we are only looking for drafts collection child folders.
-        # drafts = get_or_create_drafts_collection()
-        # doc = Folder().findOne({"parentId": drafts["_id"], "meta.dandiset.identifier": identifier})
-        # if not doc:
-        #     raise RestException("No such dandiset found.")
-        # return doc
+    def add_dandiset_owners(self, identifier, owners, params):
+        dandiset = find_dandiset_by_identifier(identifier)
+        Folder().requireAccess(dandiset, user=self.getCurrentUser(), level=AccessType.ADMIN)
 
-        # TODO:
-        pass
+        # Make sure the list doesn't contain duplicates
+        current_owners = [
+            (str(user["id"]), user["level"])
+            for user in Folder().getFullAccessList(dandiset)["users"]
+        ]
+        new_owners = [(user["_id"], AccessType.ADMIN) for user in owners]
+        final_owners = [
+            {"id": user_id, "level": level}
+            for user_id, level in set([*current_owners, *new_owners])
+        ]
+
+        doc = Folder().setAccessList(dandiset, {"users": final_owners}, save=True, recurse=True)
+        return Folder().getFullAccessList(doc)
 
     @access.user
-    @describeRoute(
-        Description("Remove Dandiset Owners").param(
-            "identifier", "Dandiset Identifier", paramType="path"
+    @autoDescribeRoute(
+        Description("Remove Dandiset Owners")
+        .param("identifier", "Dandiset Identifier", paramType="path")
+        .jsonParam(
+            "owners",
+            "A JSON list of girder users to remove from owners.",
+            paramType="body",
+            requireArray=True,
         )
     )
     @dandiset_identifier
-    def remove_dandiset_owners(self, identifier, params):
-        # # Ensure we are only looking for drafts collection child folders.
-        # drafts = get_or_create_drafts_collection()
-        # doc = Folder().findOne({"parentId": drafts["_id"], "meta.dandiset.identifier": identifier})
-        # if not doc:
-        #     raise RestException("No such dandiset found.")
-        # return doc
+    def remove_dandiset_owners(self, identifier, owners, params):
+        dandiset = find_dandiset_by_identifier(identifier)
+        Folder().requireAccess(dandiset, user=self.getCurrentUser(), level=AccessType.ADMIN)
 
-        # TODO:
-        pass
+        current_owners = {
+            (str(user["id"]), user["level"])
+            for user in Folder().getFullAccessList(dandiset)["users"]
+        }
+        owners_to_delete = {(user["_id"], AccessType.ADMIN) for user in owners}
+        final_owners = [
+            {"id": user_id, "level": level}
+            for user_id, level in (current_owners - owners_to_delete)
+        ]
+
+        doc = Folder().setAccessList(dandiset, {"users": final_owners}, save=True, recurse=True)
+        return Folder().getFullAccessList(doc)
 
     @access.user
     @autoDescribeRoute(
--- a/girder-dandi-archive/girder_dandi_archive/rest.py
+++ b/girder-dandi-archive/girder_dandi_archive/rest.py
@@ -124,7 +124,11 @@ class DandiResource(Resource):
             {"id": user_id, "level": level} for user_id, level in user_id_to_level.items()
         ]
 
-        doc = Folder().setAccessList(dandiset, {"users": final_users}, save=True, recurse=True)
+        # Assumes there is at least one admin
+        admin = next(User().getAdmins())
+        doc = Folder().setAccessList(
+            dandiset, {"users": final_users}, save=True, recurse=True, user=admin
+        )
         return get_dandiset_owners(doc)
 
     @access.user
@@ -158,7 +162,11 @@ class DandiResource(Resource):
             {"id": user_id, "level": level} for user_id, level in user_id_to_level.items()
         ]
 
-        doc = Folder().setAccessList(dandiset, {"users": final_users}, save=True, recurse=True)
+        # Assumes there is at least one admin
+        admin = next(User().getAdmins())
+        doc = Folder().setAccessList(
+            dandiset, {"users": final_users}, save=True, recurse=True, user=admin
+        )
         return get_dandiset_owners(doc)
 
     @access.user
--- a/girder-dandi-archive/girder_dandi_archive/rest.py
+++ b/girder-dandi-archive/girder_dandi_archive/rest.py
@@ -14,7 +14,7 @@ from .util import (
     DANDISET_IDENTIFIER_COUNTER,
     DANDISET_IDENTIFIER_LENGTH,
     get_or_create_drafts_collection,
-    validate_dandiset_identifier,
+    dandiset_identifier,
 )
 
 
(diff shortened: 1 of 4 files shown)


mgrauer:
A few issues here for consideration.

- girder already has a notion of owner, which is tied to the `AccessType.ADMIN` (I think that's the right variable name), if you use `AccessType.WRITE` for owner, that can be confusing. i'd recommend using `AccessType.ADMIN` here. But also this is specifying the user's access to the group, you haven't yet given the group access to the folder (see next point). So really what I'm saying is that the group or user should have ADMIN access on the folder.
- the dandisets are nested folders in girder. i don't see here where the group is given ownership of the dandiset folder. also, when the group is attached to the folder, it needs to have the permissions attached to the folder recursively (so that all subfolders are also owned by the group). you'll want to add a group to owning the folder at the time the dandiset is created (a different endpoint), unless you don't use groups at all.
- you don't have to use groups, could just use individual user permissions added and removed from the folder. using groups adds indirection and also requires additional machinery (code such as the functions you've added). i think groups are a positive when managing resource access by a human, i don't see them as valuable when the groups are automatically created per dandiset and have people added to them, as in this case. i'd be curious about other opinions here.

AlmightyYakob:
> * you don't have to use groups, could just use individual user permissions added and removed from the folder. using groups adds indirection and also requires additional machinery (code such as the functions you've added). i think groups are a positive when managing resource access by a human, i don't see them as valuable when the groups are automatically created per dandiset and have people added to them, as in this case. i'd be curious about other opinions here.

Ahh, that makes more sense, I wasn't aware you could update the access of a folder this way. I'll switch away from using groups in favor of this. This should also fix the other 2 issue you brought up, so long as I have a way to specify an admin level of access when updating the ACL of a folder.

girder-dandi-archive/girder_dandi_archive/util.py
```diff
@@ -23,3 +25,40 @@ def validate_dandiset_identifier(dandiset_identifier):
...
     return Collection().createCollection(DANDI_DRAFTS_COLLECTION_NAME, reuseExisting=True)
+
+
+def find_dandiset_by_identifier(identifier):
+    """Use a unique identifier to find a dandiset in the dandiset drafts collection."""
+    drafts = get_or_create_drafts_collection()
+    return Folder().findOne({"parentId": drafts["_id"], "meta.dandiset.identifier": identifier})
+
+
+def find_dandiset(query):
```
mgrauer, line 36: why is this added? doesn't seem like it is used. also, what form should `query` take?

girder-dandi-archive/girder_dandi_archive/rest.py
```diff
@@ -69,21 +73,84 @@ def create_dandiset(self, params):
...
+    @dandiset_identifier
+    def remove_dandiset_owners(self, identifier, owners, params):
+        dandiset = find_dandiset_by_identifier(identifier)
+        Folder().requireAccess(dandiset, user=self.getCurrentUser(), level=AccessType.ADMIN)
+
+        current_owners = {
+            (str(user["id"]), user["level"])
+            for user in Folder().getFullAccessList(dandiset)["users"]
+        }
+        owners_to_delete = {(user["_id"], AccessType.ADMIN) for user in owners}
```
mgrauer, line 145: should you do some validation of passed in list of users here and in other endpoints?

girder-dandi-archive/girder_dandi_archive/util.py
AlmightyYakob: This is a general helper function which may be useful to add into several places in `rest.py`, but I haven't done that yet as I wanted consensus  from people. I encountered a weird bug from not including the `{"parentId": drafts["_id"], ...}` portion of a mongo query. This would just abstract that out, and `query` would be any valid mongo query.

girder-dandi-archive/girder_dandi_archive/rest.py
```diff
@@ -69,21 +73,84 @@ def create_dandiset(self, params):
...
+            requireArray=True,
+        )
+    )
+    @dandiset_identifier
+    def add_dandiset_owners(self, identifier, owners, params):
+        dandiset = find_dandiset_by_identifier(identifier)
+        Folder().requireAccess(dandiset, user=self.getCurrentUser(), level=AccessType.ADMIN)
+
+        # Make sure the list doesn't contain duplicates
+        current_owners = [
```
mgrauer, line 112: you aren't checking access levels in current owners, here or in other endpoints. could that have undesired effects?

it definitely means that `final_owners` could be mislabeled as it might include non-owners, which would be confusing.

girder-dandi-archive/girder_dandi_archive/rest.py
```diff
@@ -69,21 +73,84 @@ def create_dandiset(self, params):
...
 
+    @access.public
+    @describeRoute(
+        Description("Get Dandiset Owners").param(
+            "identifier", "Dandiset Identifier", paramType="path"
+        )
+    )
+    @dandiset_identifier
+    def get_dandiset_owners(self, identifier, params):
+        return Folder().getFullAccessList(find_dandiset_by_identifier(identifier))["users"]
```
mgrauer, line 93: you should only return those with `ADMIN` access

girder-dandi-archive/girder_dandi_archive/rest.py
AlmightyYakob: I had a similar thought, but I'm not sure what I'd filter it on?

girder-dandi-archive/girder_dandi_archive/rest.py
```diff
@@ -69,21 +73,84 @@ def create_dandiset(self, params):
...
+
+        # Make sure the list doesn't contain duplicates
+        current_owners = [
+            (str(user["id"]), user["level"])
+            for user in Folder().getFullAccessList(dandiset)["users"]
+        ]
+        new_owners = [(user["_id"], AccessType.ADMIN) for user in owners]
+        final_owners = [
+            {"id": user_id, "level": level}
+            for user_id, level in set([*current_owners, *new_owners])
```
dchiquit, line 120: If a user already has non-admin permissions on the folder and is added again with this method, that user will appear twice in the owners list with two different permissions levels.

Why not just use `user["_id"]` alone for `current_owners` and `new_owners` and do `final_owners = [{"id": user_id, "level": AccessType.ADMIN}]? This will upgrade any existing users to ADMIN, but that's currently what we expect 

If you're trying to preserve user permission levels, I would build a dict `{user_id: access_level}` of the current owners, add the new owners, then map that back to the required form of `[{"id":..., "level":...}]`. I think you did something similar in `remove_dandiset_owners`.

girder-dandi-archive/girder_dandi_archive/rest.py
AlmightyYakob: Good catch, addressed in 753cce0

girder-dandi-archive/tests/rest/test_dandiset_ownership.py
```diff
@@ -0,0 +1,88 @@
...
+def test_initial_dandiset_owners(server, admin_created_dandiset):
+    """Assert that the list of owners isn't zero.
+
+    The group members should always contain at least the dandiset creator.
+    """
+    identifier = admin_created_dandiset["meta"]["dandiset"]["identifier"]
+    resp = server.request(path=f"/dandi/{identifier}/owners", method="GET")
+
+    assertStatusOk(resp)
+    assert len(resp.json) == 1
```
dchiquit, line 52: I think it's worth asserting something about the one user, even if it's just an ID.

girder-dandi-archive/tests/rest/test_dandiset_ownership.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def test_add_dandiset_owners(server, admin, user, admin_created_dandiset):
+    """Assert that adding a user to the list of owners produces the expected result."""
```
dchiquit, line 56: I don't like to assume that the response from an operation means that it changed the things it said it did. I would like to see an attempt from the user to use the folder without permission that fails with a 403, then that same attempt succeeding after being granted permission.

girder-dandi-archive/girder_dandi_archive/rest.py
AlmightyYakob: Since the call to `setAccessList` will replace entirely what's currently in the ACL, we have to include all users in the ACL in `final_owners`, in order to preserve any existing users, regardless of their access level.

AlmightyYakob: However I agree, the names are a bit misleading in that regard, so I can change them.

girder-dandi-archive/tests/rest/test_dandiset_ownership.py
```diff
@@ -0,0 +1,88 @@
...
+            "name": "admin_created_dandiset",
+            "description": "Admin created this test dandiset.",
+        },
+    )
+    assertStatusOk(resp)
+    return resp.json
+
+
+@pytest.fixture
+def multi_owner_dandiset(server, admin, user, admin_created_dandiset):
```
dchiquit, line 26: It's not an issue as this currently stands, but using the `multi_owner_dandiset` fixture will corrupt the `admin_created_dandiset` fixture. I can't picture a single test that would need both, but it feels like a trap. Considering this fixture is only used in one test, it might be better to remove this fixture and just add the permissions in the test? I don't feel very strongly.

girder-dandi-archive/tests/rest/test_dandiset_ownership.py
```diff
@@ -0,0 +1,88 @@
...
+        path=f"/dandi/{identifier}/owners",
+        method="DELETE",
+        body=request_body,
+        user=admin,
+        type="application/json",
+    )
+
+    users = resp.json["users"]
+    assert len(users) == 1
+    assert not [new_user for new_user in users if new_user["id"] == str(user["_id"])]
```
dchiquit, line 88: Same comment about asserting that the deleted user actually has the correct permissions by making some requests with the user.

AlmightyYakob:
This PR is definitely still a work in progress, as I'm addressing your concerns @mgrauer I'm finding some more edge cases to keep track of. I'll mark this as ready for review when I think I've addressed everything.

girder-dandi-archive/tests/rest/test_dandiset_ownership.py
AlmightyYakob: The main reason I did that is because the remove test basically contained the entirety of the add test, and I didn't like that pattern. 

I was curious about this, as I had thought that fixtures were created new and independently for each test, so I tried testing it. What I did was replace the `admin_created_dandiset` argument in `test_initial_dandiset_owners` with `multi_owner_dandiset`, and then ran the tests. The tests run in the order of:

1. `test_initial_dandiset_owners`
2. `test_add_dandiset_owners`
3. `test_remove_dandiset_owners`

I didn't replace the fixture argument of `admin_created_dandiset` in `test_add_dandiset_owners`, so if it were to corrupt the `admin_created_dandiset` fixture, in that test it should show two users as being owners on the dandiset, instead of just 1. However, when I did this, only the 1 user showed.

Furthermore if I just printed the drafts collection in each test, I get different ID each time. I know this is a lot more than you asked for, but from looking around online and testing myself, it seems like mutation of the fixtures won't actually occur, so I'd be fine leaving it as is. If there's something here I'm missing please lmk.

girder-dandi-archive/girder_dandi_archive/rest.py
AlmightyYakob: However if you mean validating that the passed in user objects themselves are valid user objects then I agree, that's something I'm going to add.

girder-dandi-archive/tests/rest/test_dandiset_ownership.py
dchiquit: Sorry, I should have been more clear. The clobbering will occur when you have a test that requests both the `multi_owner_dandiset` fixture and the `admin_created_dandiset` at the same time, e.g.
```
def broken_test(admin_created_dandiset, multi_owner_dandiset):
  assert admin_created_dandiset["_id"] != multi_owner_dandiset["_id"] # fails
```
Both fixtures refer to the same dandiset, which now has multiple owners. The way it's written now will work perfectly for any reasonable tests that I can think of. I'm not 100% happy with the way this is written, but all the alternatives I can think of have flaws of similar magnitude. Up to you.

AlmightyYakob: Ahh I see what you mean. In that case I agree they should probably be independent. It will result in some duplicate code but I'm okay with that. I'll make that change.

AlmightyYakob:
Had a discussion with @mgrauer, we decided that since we're going for a simplified permission model, users will either be owners of a dandiset, or have no permissions. That means that any operations done by these endpoints should be assumed to remove all existing permissions lower than `AccessType.ADMIN`.

girder-dandi-archive/girder_dandi_archive/rest.py
AlmightyYakob: 94b1e48

girder-dandi-archive/tests/rest/test_dandiset_ownership.py
```diff
@@ -0,0 +1,234 @@
...
+        path=f"/dandi/{identifier}/owners",
+        method="DELETE",
+        body=request_body,
+        user=admin,
+        type="application/json",
+    )
+
+    users = resp.json
+    assert len(users) == 1
+    assert not [new_user for new_user in users if new_user["id"] == str(user["_id"])]
```
dchiquit, line 141:
```suggestion
    assert resp.json[0]["id"] == str(admin["_id"])
```
I personally find list comprehensions difficult to parse. I think if we've already asserted that there is exactly one user, it is more legible to assert that that one user is the admin.

girder-dandi-archive/tests/rest/test_dandiset_ownership.py
```diff
@@ -0,0 +1,234 @@
...
+            path=f"/folder/{multi_owner_dandiset['_id']}/access", method="GET", user=user
+        ),
+        403,
+    )
+
+
+def test_user_removed_from_owners(server, user, user_2, admin, admin_created_dandiset):
+    """Test that an existing user without admin permissions has none after another user is added."""
+    access_param = json.dumps(
+        {"users": [{"id": user["_id"], "level": AccessType.WRITE}]}, default=str
```
dchiquit, line 155:
```suggestion
@pytest.mark.parametrize("accessLevel", [AccessType.READ, AccessType.WRITE])
def test_user_removed_from_owners(server, user, user_2, admin, admin_created_dandiset, accessLevel):
    """Test that an existing user without admin permissions has none after another user is added."""
    access_param = json.dumps(
        {"users": [{"id": user["_id"], "level": accessLevel}]}, default=str
```
Write one get one free
Same deal for the next test

girder-dandi-archive/girder_dandi_archive/rest.py
```diff
@@ -69,21 +75,107 @@ def create_dandiset(self, params):
...
+    def add_dandiset_owners(self, identifier, owners, params):
+        dandiset = find_dandiset_by_identifier(identifier)
+        Folder().requireAccess(dandiset, user=self.getCurrentUser(), level=AccessType.ADMIN)
+
+        # Make sure the list doesn't contain duplicates
+        # Only work with admin level users, removing non-admins from the ACL
+        user_id_to_level = {
+            str(user["id"]): user["level"]
+            for user in Folder().getFullAccessList(dandiset)["users"]
+            if user["level"] == AccessType.ADMIN
```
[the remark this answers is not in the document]
AlmightyYakob: Good suggestion. As far as the difference between those two keys, `"id"` is just how the ACL of a folder returns the id of a user, instead of `"_id"` normally found elsewhere.

girder-dandi-archive/girder_dandi_archive/rest.py
```diff
@@ -69,21 +75,107 @@ def create_dandiset(self, params):
...
+        }
+
+        for owner in owners:
+            if not validate_user(owner):
+                raise ValidationException("All owners must be valid user objects.")
+
+            # Prevents a KeyError if owner isn't an existing owner
+            user_id_to_level.pop(owner["_id"], None)
```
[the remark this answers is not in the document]
AlmightyYakob: I originally had that, but I realized that will cause a `KeyError` if any of the users passed into the endpoint aren't already owners on the dandiset. We could handle this case and throw an error, but I'm in favor of keeping this endpoints idempotent, which this currently is.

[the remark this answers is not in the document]
dchiquit: I had not thought of that, good point

mgrauer:
Can you wait to merge until I've had a chance to review?

AlmightyYakob:
> Can you wait to merge until I've had a chance to review?

Sure thing

girder-dandi-archive/girder_dandi_archive/util.py
```diff
@@ -23,3 +28,46 @@ def validate_dandiset_identifier(dandiset_identifier):
...
     return Collection().createCollection(DANDI_DRAFTS_COLLECTION_NAME, reuseExisting=True)
+
+
+def find_dandiset_by_identifier(identifier):
+    """Use a unique identifier to find a dandiset in the dandiset drafts collection."""
+    drafts = get_or_create_drafts_collection()
+    return Folder().findOne({"parentId": drafts["_id"], "meta.dandiset.identifier": identifier})
+
+
+def dandiset_find(query, **kwargs):
```
mgrauer, line 39: can you remove this? it isn't called anywhere and it isn't clear to me how to call it or why it is needed from reading the code (I'm not saying it is not useful, but it is opaque to me when to call it and how, and that's from reading it in context of the PR that adds it. in the future someone coming in will not know how to use it, but will be afraid to remove it). i think it just adds cruft to the codebase and violates YAGNI.

girder-dandi-archive/girder_dandi_archive/rest.py
```diff
@@ -69,21 +75,103 @@ def create_dandiset(self, params):
...
+            requireArray=True,
+        )
+    )
+    @dandiset_identifier
+    def add_dandiset_owners(self, identifier, owners, params):
+        dandiset = find_dandiset_by_identifier(identifier)
+        Folder().requireAccess(dandiset, user=self.getCurrentUser(), level=AccessType.ADMIN)
+
+        # Make sure the list doesn't contain duplicates
+        # Only work with admin level users, removing non-admins from the ACL
```
mgrauer, line 114: This is surprising enough that I think it deserves more explanation. Perhaps:

# Only work with admin level users, removing non-admins from the ACL/
# This is done intentionally as a simplifying assumption of providing this API.

girder-dandi-archive/girder_dandi_archive/rest.py
```diff
@@ -69,21 +75,103 @@ def create_dandiset(self, params):
...
+            paramType="body",
+            requireArray=True,
+        )
+    )
+    @dandiset_identifier
+    def remove_dandiset_owners(self, identifier, owners, params):
+        dandiset = find_dandiset_by_identifier(identifier)
+        Folder().requireAccess(dandiset, user=self.getCurrentUser(), level=AccessType.ADMIN)
+
+        # Only work with admin level users, removing non-admins from the ACL
```
mgrauer, line 152: add same comment as above, for extra explanation

mgrauer:
ok, a couple minor requests and then could be merged. i don't need to review further.

girder-dandi-archive/tests/rest/test_dandiset_ownership.py
```diff
@@ -0,0 +1,231 @@
...
+
+from girder.constants import AccessType
+from girder.models.user import User
+from pytest_girder.assertions import assertStatus, assertStatusOk
+
+pytestmark = pytest.mark.plugin("dandi_archive")
+
+
+@pytest.fixture
+def user_2(server):
```
brianhelba, line 13: Would it be helpful to move some of these fixtures to `conftest.py` in case they're reused in other test files, or is that not likely?

AlmightyYakob: b631ba6

PR #343, opened 2020-05-15
Fetch version info from publish API

### LINKED ISSUE #342: Fetch data from the publish prototype to the GUI
Part of #340 
Hook up the publish API into the GUI so that it is possible to fetch the current version data. Displaying it cleanly is out of scope.

dchiquit:
Define a new axios client for using the publish API.

Define a new `publish` store for storing stuff from the publish API,
much as we have a `girder` store to store stuff from girder.

Display the published versions on the DandisetLandingView. This display
code is temporary to prove that the underlying code works as intended.

Fixes #342


web/src/rest.js
```diff
@@ -1,14 +1,19 @@
 
+import axios from 'axios';
 import { RestClient } from '@girder/components/src';
 
 const apiRoot = process.env.VUE_APP_API_ROOT || 'http://localhost:8080/api/v1';
+const publishApiRoot = process.env.PUBLISH_APP_API_ROOT || 'http://localhost:8000/api';
```
brianhelba, line 6: The `VUE_APP_` prefix is magic, and is necessary for the value to be injected: https://cli.vuejs.org/guide/mode-and-env.html#environment-variables

Perhaps `VUE_APP_PUBLISH_API_ROOT`?

brianhelba, line 6: I think it'd be better to to define a typical development default of the publish API root environment variable in `.env.development`, and omit the default value here (likewise with `apiRoot`). For production builds, I don't think we want this fallback default, as we want the app to fail obviously if the environment was not configured.

web/src/views/DandisetLandingView/DandisetLandingView.vue
```diff
@@ -270,6 +277,10 @@ export default {
     ...mapState('girder', {
       selected: (state) => state.currentDandiset,
     }),
+    ...mapState('publish', {
+      mostRecentDandiset: (state) => state.mostRecentDandiset,
+      versions: (state) => state.versions,
+    }),
```
brianhelba, line 283: This can be simplified: https://vuex.vuejs.org/guide/state.html#the-mapstate-helper

```suggestion
    ...mapState('publish', [
      'mostRecentDandiset',
      'versions',
    ]),
```

web/src/store/publish.js
```diff
@@ -0,0 +1,28 @@
...
+  mutations: {
+    setVersions(state, versions) {
+      state.versions = versions;
+    },
+  },
+  actions: {
+    async fetchPublishedVersions({ commit }, identifier) {
+      if (identifier) {
+        const response = await publishRest.get(`dandisets/${identifier}/versions/`)
+          .catch((error) => {
```
brianhelba, line 17: I think using `try...catch` exclusively with `await` is cleaner in most cases, including here.

web/src/store/publish.js
```diff
@@ -0,0 +1,28 @@
+import { publishRest } from '@/rest';
+
+export default {
+  namespaced: true,
+  state: {
+    versions: null,
```
brianhelba, line 6: I don't think any of this should be in Vuex at all. Since every different navigation of the `DandisetLandingView` is going to be overwriting all of the state here, it's better to keep it within the `DandisetLandingView` itself, rather than deal with the extra complexity and potential desynchronization bugs of having it in a singleton Vuex module.

web/src/views/DandisetLandingView/DandisetLandingView.vue
```diff
@@ -301,6 +312,10 @@ export default {
         }
       },
     },
+    // TODO move this somewhere more sensical
+    async meta(val) {
+      this.$store.dispatch('publish/fetchPublishedVersions', val.identifier);
```
brianhelba, line 317: How about putting this at the end of the `id` watcher, which initializes this component on creation or navigation?

web/src/views/DandisetLandingView/DandisetLandingView.vue
```diff
@@ -270,6 +277,10 @@ export default {
     ...mapState('girder', {
       selected: (state) => state.currentDandiset,
     }),
+    ...mapState('publish', {
+      mostRecentDandiset: (state) => state.mostRecentDandiset,
```
AlmightyYakob, line 281: I don't see where this exists in the store, and it doesn't look like it's used anywhere else in this component either. Why is this here?

web/src/store/publish.js
AlmightyYakob: I agree, and if `DandisetLandingView` is the only place using this it makes sense. I know we plan on showing version info in the main dandiset listing component ([figma link](https://www.figma.com/proto/kEHOgqU0rcOIyCYIDlpSMV/DANDI-UI?scaling=min-zoom&node-id=16%3A78)), but that seems orthogonal to this, and depends on the specifics of how we want display that info.

web/src/views/DandisetLandingView/DandisetLandingView.vue
dchiquit: I thought it might be useful, but decided it was too much for this review. I forgot to remove it here, my b

web/src/rest.js
```diff
@@ -3,7 +3,7 @@ import axios from 'axios';
 import { RestClient } from '@girder/components/src';
 
 const apiRoot = process.env.VUE_APP_API_ROOT || 'http://localhost:8080/api/v1';
-const publishApiRoot = process.env.PUBLISH_APP_API_ROOT || 'http://localhost:8000/api';
```
brianhelba: Can you remove this default too, just for consistency?

PR #346, opened 2020-05-18
Add use of util functions

AlmightyYakob:
The main benefit of this is removing the need to specify `{"parentId": drafts["_id"]}` everywhere, which cleans up the code as well as offering an abstracted way of finding dandisets by their identifier (via `find_dandiset_by_identifier`).

### DIFF
--- a/girder-dandi-archive/girder_dandi_archive/rest.py
+++ b/girder-dandi-archive/girder_dandi_archive/rest.py
@@ -11,6 +11,7 @@ from girder.models.setting import Setting
 from girder.models.user import User
 
 from .util import (
+    dandiset_find,
     dandiset_identifier,
     DANDISET_IDENTIFIER_COUNTER,
     DANDISET_IDENTIFIER_LENGTH,
@@ -78,8 +79,7 @@ class DandiResource(Resource):
     @dandiset_identifier
     def get_dandiset(self, identifier, params):
         # Ensure we are only looking for drafts collection child folders.
-        drafts = get_or_create_drafts_collection()
-        doc = Folder().findOne({"parentId": drafts["_id"], "meta.dandiset.identifier": identifier})
+        doc = find_dandiset_by_identifier(identifier)
         if not doc:
             raise RestException("No such dandiset found.")
         return doc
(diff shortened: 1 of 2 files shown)


girder-dandi-archive/girder_dandi_archive/rest.py
```diff
@@ -203,15 +198,14 @@ def list_dandisets(self, limit, offset, sort):
     )
     def search_dandisets(self, search, limit, offset, sort):
         # Ensure we are only looking for drafts collection child folders.
-        drafts = get_or_create_drafts_collection()
         # TODO Currently only searching identifier, name, description, and contributor name
```
mgrauer, line 201: what is the TODO here exactly?

mgrauer:
Let's wait to merge so we can discuss.

My general thoughts are that (1) this is good, the more we can centralize the implementation details of the way drafts work the better (2) the endpoints generally read more cleanly now (3) what should we do with all those comments in the endpoints about drafts folders?--they are now disconnected from the fact that we are searching in the drafts folders because the search is in a utility function and it no longer makes sense to describe the implementation details (see 1 above), but these comments are noting something important about drafts that potentially should be preserved (4) i'm not a huge fan of the kwargs style signatures, but maybe this is the right approach here.

girder-dandi-archive/girder_dandi_archive/rest.py
dchiquit: That's me, basically it's `search better`. I wanted to leave a note describing the current behavior so whoever gets to do the full search function knows what's up.

AlmightyYakob:
> I think this is a great idea. The function name `find_dandiset` reads a little better to me, but this is fine as is.

I originally thought that too, but to me, `find_dandiset` implies that a single dandiset will be returned, when actually it's just calling `find` behind the scenes, so I thought `dandiset_find` would convey that it's the `find` operation, but scoped to dandisets.

girder-dandi-archive/girder_dandi_archive/rest.py
mgrauer: Could we remove this comment in that case? There isn't something actionable here and this suggests to me when reading the code that something is to be done.

PR #348, opened 2020-05-19
Add most recent published version to lists pages

### LINKED ISSUE #347: Display draft/version on dandiset list pages
Before displaying a list of dandisets, the publish API should be queried to get the current published versions for each dandiset. The latest version should be displayed, or the text `DRAFT` if nothing has been published yet.

dchiquit:
Before displaying a DandisetList, query the publish API to get the most
recently published dandiset. If it exists, display it instead. If it
doesn't exist, display the girder draft as before.

Add some convenience methods to the `publishRest` object.

Fixes #347 

Screenshot from 2020-05-19 12-44-08

### DIFF
--- a/web/src/components/DandisetsPage.vue
+++ b/web/src/components/DandisetsPage.vue
@@ -139,15 +139,9 @@ export default {
       if (this.girderDandisetRequest === null || this.publishedVersions === null) {
         return null;
       }
-      const dandisets = [];
-      for (let i = 0; i < this.totalDandisets; i += 1) {
-        if (this.publishedVersions[i]) {
-          dandisets.push(this.publishedVersions[i]);
-        } else {
-          dandisets.push(this.girderDandisetRequest.data[i]);
-        }
-      }
-      return dandisets;
+      return this.girderDandisetRequest.data.map(
+        (dandiset, i) => (this.publishedVersions[i] || dandiset),
+      );
     },
     totalDandisets() {
       return this.girderDandisetRequest ? this.girderDandisetRequest.headers['girder-total-count'] : 0;
--- a/web/src/components/DandisetsPage.vue
+++ b/web/src/components/DandisetsPage.vue
@@ -68,7 +68,7 @@
 <script>
 import DandisetList from '@/components/DandisetList.vue';
 import DandisetSearchField from '@/components/DandisetSearchField.vue';
-import { girderRest } from '@/rest';
+import { girderRest, publishRest } from '@/rest';
 
 const DANDISETS_PER_PAGE = 8;
 
@@ -136,14 +136,25 @@ export default {
       return { page, sortOption, sortDir };
     },
     dandisets() {
-      return this.dandisetRequest ? this.dandisetRequest.data : undefined;
+      if (this.girderDandisetRequest === null || this.publishedVersions === null) {
+        return null;
+      }
+      const dandisets = [];
+      for (let i = 0; i < this.totalDandisets; i += 1) {
+        if (this.publishedVersions[i]) {
+          dandisets.push(this.publishedVersions[i]);
+        } else {
+          dandisets.push(this.girderDandisetRequest.data[i]);
+        }
+      }
+      return dandisets;
     },
     totalDandisets() {
-      return this.dandisetRequest ? this.dandisetRequest.headers['girder-total-count'] : 0;
+      return this.girderDandisetRequest ? this.girderDandisetRequest.headers['girder-total-count'] : 0;
     },
   },
   asyncComputed: {
-    async dandisetRequest() {
+    async girderDandisetRequest() {
       const {
         listingUrl, page, sortField, sortDir,
         $route: {
@@ -168,6 +179,16 @@ export default {
       return {
         data, headers, status, statusText,
       };
+    },
+    async publishedVersions() {
+      if (!this.girderDandisetRequest) {
+        return null;
+      }
+      return Promise.all(this.girderDandisetRequest.data.map(
+        async (dandiset) => publishRest.mostRecentVersion(
+          dandiset.meta.dandiset.identifier, dandiset._id,
+        ),
+      ));
     },
   },
   watch: {
--- a/web/src/rest.js
+++ b/web/src/rest.js
@@ -5,8 +5,60 @@ import { RestClient } from '@girder/components/src';
 const apiRoot = process.env.VUE_APP_API_ROOT;
 const publishApiRoot = process.env.VUE_APP_PUBLISH_API_ROOT;
 
+// TODO remove the girderId if we can eliminate the girder ID from the URL
+function girderize(publishedDandiset, girderId) {
+  const { // eslint-disable-next-line camelcase
+    created, updated, dandi_id, version, metadata,
+  } = publishedDandiset;
+  return {
+    _id: girderId,
+    created,
+    updated,
+    version,
+    name: dandi_id,
+    lowerName: dandi_id,
+    meta: metadata,
+  };
+}
+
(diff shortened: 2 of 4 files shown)


web/src/rest.js
```diff
@@ -5,9 +5,61 @@ import { RestClient } from '@girder/components/src';
 const apiRoot = process.env.VUE_APP_API_ROOT;
 const publishApiRoot = process.env.VUE_APP_PUBLISH_API_ROOT;
 
+// TODO remove the girderId if we can eliminate the girder ID from the URL
```
dchiquit, line 8: Our frontend is still very girder-centric, so sometimes there is no choice but to accommodate that. This method translates the django schema from the publish API into something that can be easily rendered by the current frontend.
The girder object ID is not provided from the publish API, and it is required as both a unique key for rendering the list and to generate the route to navigate to when clicking on one of the dandisets. I pass it in to this helper, but I would much rather not have to do that. That's a problem for later though.

AlmightyYakob: I agree, and think it'd be better if we didn't rely on the girder ID in the route or here. However, that would necessitate a URL change, which may conflict with the redirector. I agree we should deal with this in a later issue.

web/src/rest.js
```diff
@@ -5,9 +5,61 @@ import { RestClient } from '@girder/components/src';
...
+    name: dandi_id,
+    lowerName: dandi_id,
+    meta: metadata,
+  };
+}
+
 const girderRest = new RestClient({ apiRoot, setLocalCookie: true });
 const publishRest = axios.create({ baseURL: publishApiRoot });
 
+Object.assign(publishRest, {
```
AlmightyYakob, line 27: I personally tend to stray away from extending `axios` objects, and towards normal functions like shown in `loggedIn()` and `user()`. Not saying using `Object.assign` is the wrong approach here, but I'm just curious what the benefits are? Is it just for the shorthand?

mgrauer:
>Also, does this mean the expected behavior is to only show a dandiset as a draft if it has no published versions yet? Technically, dandisets between versions are also drafts, correct? 

I'll add in a little to clarify in case there is still confusion, and what I'm explaining is the common main line workflow (not describing corner cases like someone deleting the draft dandiset after creating publications), based on the current implementation, which has no embargoed datasets.

- There is a draft dandiset (in implementation this is a Girder folder), that is first created, and exists always.
- Dandiset owners put data in the draft Dandiset, and edit the data and metadata. At some point they want to publish.
- Upon publication, a copy of the exact data and metadata in the draft folder is copied to a separate place, made immutable, read-only (to the Dandiset owner) and public.
- For now, no change happens to the data in the Dandiset draft upon publication. A dataset owner could come along and change the data in the draft folder, or leave the data as is for as long as they like. I believe this is the Draft Dandiset that @AlmightyYakob is referring to with "Technically, dandisets between versions are also drafts, correct? "
- At any point in the future, a Dandiset owner could come along and publish the data again. Which would perform publication against whatever is in the current Dandiset draft, with a new version. (Note that we haven't yet defined what will happen in the corner case where someone wants to publish a Dandiset that is the exact same as the most recent published version)

I still think there is cognitive friction around the term `Draft`. Perhaps `Staging` would be better, or perhaps `Draft of Next Published Version`.

AlmightyYakob:
> * (Note that we haven't yet defined what will happen in the corner case where someone wants to publish a Dandiset that is the exact same as the most recent published version)

I think this is where something like a "dirty" flag that @brianhelba mentioned would come into play. Once the publish succeeds, the dirty flag is set to false, and any publishes while that flag is still false  are either no-ops or are idempotent in regards to the previous publish. Once any data in the dandiset is changed, the dirty flag is set to true, and the cycle repeats.

web/src/components/DandisetsPage.vue
```diff
@@ -169,6 +180,16 @@ export default {
...
       };
     },
+    async publishedVersions() {
+      if (!this.girderDandisetRequest) {
+        return null;
+      }
+      return Promise.all(this.girderDandisetRequest.data.map(
+        async (dandiset) => publishRest.mostRecentVersion(
+          dandiset.meta.dandiset.identifier, dandiset._id,
+        ),
```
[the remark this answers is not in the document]
dchiquit: I agree it would be cleaner, but if we didn't need the girder ID, I think the destructuring would be unnecessary. I'd rather not modify the signature and overall design of the function to canonicalize a temporary hack. The more we look at this and cringe, the sooner we fix the underlying problem.

web/src/rest.js
dchiquit: I copied the pattern from the implementation of girder's `RestClient`. I don't really like how decontextualized the raw functions are:
```
import { versions }  from '@/rest';
... lots of code ...
const myVersions = await versions(identifier); // ??? where did this come from? why async?
```
vs.
```
import { publishRest } from `@/rest`;
... lots of code ...
const myVersions = await publishRest.versions(identifier); // ah yes, an API call
```
 Especially now that we have two different client objects, I think it's helpful to have the namespacing provided by having the functions belong to their respective clients.

I admit I'm a little biased because I've used Java a lot and this is how the Java implementation would be set up.
`Object.assign` is gross, but I like the more explicit API surface it provides. Are there benefits to just exporting the raw functions I'm not thinking of?

web/src/components/DandisetsPage.vue
```diff
@@ -136,14 +136,25 @@ export default {
...
+      }
+      const dandisets = [];
+      for (let i = 0; i < this.totalDandisets; i += 1) {
+        if (this.publishedVersions[i]) {
+          dandisets.push(this.publishedVersions[i]);
+        } else {
+          dandisets.push(this.girderDandisetRequest.data[i]);
+        }
+      }
+      return dandisets;
```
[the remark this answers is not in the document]
dchiquit: TBH I think maps/functional chains are pretty neato, but they're always kinda painful to read. The extra index parameter is cool, but makes it harder to parse. I think that's more a personal weakness of mine.
I definitely try to avoid `for (i=0;i<x;i++)` loops in JS, so I'm doing this, just thought I'd leave my 2¢

dchiquit:
Just to summarize the current behavior:

My Dandisets/Public Dandisets/Search Dandisets:
- Use the old girder endpoints to fetch the dandisets to display.
- For every dandiset in the list, query the publish API to get the most recently published dandiset.
- If present, display it with the most recent version tag. If a draft has changed a title or whatever since the last publish, it will not show up in the listing.
- If 404, display the girder data with the `DRAFT` tag.

Dandiset Landing Page:
 - Query the publish API to get the list of published versions and display them on the page.
 - The whole page is getting revamped very soon, so I didn't put much effort into making it look good.

satra:
Shall we schedule a call for discussing this together with how to reduce duplication? and include that someone may download a dandiset trim it/expand it for their scientific purpose and post it as a new dandiset.

dchiquit:
@satra @mgrauer I think the deduplication discussion is still premature. The publish app is still a prototype for now, so it makes sense to use the simplest, most naive implementation.

We have a discussion planned for this Thursday to talk about publish, I think it's worth bringing up desired behavior then.

mgrauer:
>Shall we schedule a call for discussing this together with how to reduce duplication? and include that someone may download a dandiset trim it/expand it for their scientific purpose and post it as a new dandiset.

For sure--I'm trying to set up a call today to talk through some of the publication workflow and requirements. And then we will have our call tomorrow to talk over schema changes and API (including publish API). 

I think the deduplication work should be seen as a behind the scenes optimization that we add in a later pass after working out the initial publication workflow--but it's important to have this in mind as part of designing the API surface so we can add this in without breaking the API (or breaking it intentionally and knowing to let people know that this could happen).

PR #350, opened 2020-05-19
Fix login redirect bug

### LINKED ISSUE #349: Fix login not redirecting to a dandiset
When a user that is not logged in views a dandiset, then clicks `Log in`, the return URL does not record the dandiset they were viewing. Logging in will bounce them back to a blank dandiset page.

dchiquit:
When a user that is not logged in views a dandiset, then clicks Log in, the return URL does not record the dandiset they were viewing. Logging in will bounce them back to a blank dandiset page.
Fixes #349

### DIFF
--- a/web/src/components/AppBar/AppBar.vue
+++ b/web/src/components/AppBar/AppBar.vue
@@ -1,7 +1,5 @@
 <template>
-  <v-app-bar
-    app
-  >
+  <v-app-bar app>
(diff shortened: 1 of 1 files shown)


dchiquit:
Already fixed in https://github.com/dandi/dandiarchive/commit/0db3a003625fe9e018ddb300833b4380a3a70f10

PR #351, opened 2020-05-20
Update DLP

AlmightyYakob:
I apologize for the diff, as I split up `DandisetLandingView.vue` to keep things cleaner.

I'm not sure if we want to change what main data is displayed, I started with the Figma mockups, but then just added the rest of the data like was in the old DLP. I'm not sure if we want to show this all, or what we want to show.


AlmightyYakob:
@mgrauer Your opinion on the new look would be good

web/src/views/DandisetLandingView/DandisetDetails.vue
```diff
@@ -0,0 +1,176 @@
...
+                <template v-slot:activator="{ on }">
+                  <v-icon
+                    small
+                    color="grey darken-1"
+                    v-on="on"
+                  >
+                    mdi-help-circle
+                  </v-icon>
+                </template>
+                <span>This is a version of your dandiset that contains uncommitted changes.</span>
```
dchiquit, line 72:
```suggestion
                <span>This is a version of your dandiset that contains unpublished changes.</span>
```

web/src/views/DandisetLandingView/DandisetLandingView.vue
```diff
@@ -8,142 +8,51 @@
...
+          @click="detailsPanel = !detailsPanel"
+        >
+          <v-icon color="white">
+            <template v-if="detailsPanel">
+              mdi-chevron-right
+            </template>
+            <template v-else>
+              mdi-chevron-left
+            </template>
+          </v-icon>
```
dchiquit, line 36: This `>` is situated right next to the search box and looks exactly like a "submit the search query" button. The figma makes it a little clearer that it's an expand/collapse button, but not a whole lot.
I don't see an obvious way to change it, but the way it is now is very counterintuitive.

dchiquit: You also cleverly hide the details view completely at lower widths, but the chevron is still visible. That should also be hidden if the thing it expands/collapses is never shown.

web/src/views/DandisetLandingView/DandisetLandingView.vue
```diff
@@ -8,142 +8,51 @@
...
+        </v-btn>
+      </v-toolbar>
+      <v-container
+        fluid
+        class="grey lighten-4"
+      >
         <v-row>
+          <v-col>
+            <DandisetData
```
dchiquit, line 45: It's unclear to me reading this what `DandisetData` is vs. `DandisetDetails`. It looks like `DandisetData` is the main content, and `DandisetDetails` is the collapsible side bar. Perhaps they could be named more clearly?

web/src/views/DandisetLandingView/DandisetDetails.vue
```diff
@@ -0,0 +1,176 @@
...
+          </v-timeline-item>
+        </v-timeline>
+      </v-row> -->
+    </template>
+    <template v-else>
+      <v-progress-linear
+        top
+        indeterminate
+      />
+    </template>
```
dchiquit, line 117: I couldn't get this to render by slowing down my connection. If the dandiset hasn't loaded yet, I just see a wimpy little skeleton of a card. Are you sure this is working?

web/src/views/DandisetLandingView/DandisetDetails.vue
```diff
@@ -0,0 +1,176 @@
...
+    }),
+  },
+  methods: {
+    formatDateTime(datetimeStr) {
+      const datetime = moment(datetimeStr);
+      const date = datetime.format('LL');
+      const time = datetime.format('hh:mm A');
+
+      return `${date} at ${time}`;
+    },
```
dchiquit, line 173: This smells like it belongs in `utils.js`, but since it's more about formatting for presentation, I can see it belonging with a specific component.

web/src/views/DandisetLandingView/DandisetDetails.vue
```diff
@@ -0,0 +1,176 @@
...
+      rowClasses: 'my-1',
+      labelClasses: 'mx-2 text--secondary',
+      itemClasses: 'font-weight-medium',
+    };
+  },
+  computed: {
+    computedSize() {
+      if (!this.currentDandiset || !this.currentDandiset.size) return null;
+      return filesize(this.currentDandiset.size);
+    },
```
dchiquit, line 140: I don't see this used anywhere. IIRC we're waiting to figure out exactly how we should be calculating file size in girder? This could conceivably just be removed until we decide to tackle that problem.

web/src/views/DandisetLandingView/DandisetLandingView.vue
AlmightyYakob: It's actually not hidden, it's just pushed below. I'm not sure what behavior we want here exactly, i.e. where to show this data or if to show it at all at lower widths

web/src/views/DandisetLandingView/DandisetDetails.vue
AlmightyYakob: Good catch, that was kind of left over. I'll push a better implementation of that.

web/src/views/DandisetLandingView/DandisetDetails.vue
AlmightyYakob: Since it's currently specific to this, I say we leave it here, but if in the future there's a need for it we can easily move it to `utils.js`.

web/src/views/DandisetLandingView/DandisetDetails.vue
AlmightyYakob: > IIRC we're waiting to figure out exactly how we should be calculating file size in girder?

Yes, however I believe I've discussed this recently with @mgrauer and this should be addressed soon, so I think it's fine to leave in. However I don't feel too strongly about it, I could go either way.

web/src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -0,0 +1,146 @@
...
+        </v-card-title>
+      </v-row>
+      <v-row class="mx-1 px-4 font-weight-light">
+        {{ meta.description }}
+      </v-row>
+
+      <template v-for="(field, key) in extraFields">
+        <v-divider
+          :key="`${key}-divider`"
+          class="mt-4"
```
dchiquit, line 55: This divider works well for maintaining spacing between list items, but the very last item doesn't have a divider, so it doesn't have a margin between it and the bottom border. It's purely cosmetic, but it looks very abreviated.

web/src/views/DandisetLandingView/DandisetLandingView.vue
dchiquit: I think making it fill the width and putting it at the top would be reasonably intuitive. The open/close button is still confusing though.

AlmightyYakob: > I think making it fill the width and putting it at the top would be reasonably intuitive.

This has been addressed in 4bd2df5, I encourage you to try it out. I'm not sure what to do about the open/close button, but I think it should be left as a follow-up issue, so we can get this merged.

web/src/views/DandisetLandingView/DandisetDetails.vue
AlmightyYakob: fd12c94

web/src/views/DandisetLandingView/DandisetLandingView.vue
dchiquit: How about up/down chevrons instead of left/right? After toying with it they seem to convey the same thing without being as suggestive of a "Search" button.
I do think the button needs some more work, though. This is just a bandaid.

web/src/views/DandisetLandingView/DandisetDetails.vue
```diff
@@ -0,0 +1,168 @@
...
+import moment from 'moment';
+
+
+export default {
+  name: 'DandisetDetails',
+  data() {
+    return {
+      rowClasses: 'my-1',
+      labelClasses: 'mx-2 text--secondary',
+      itemClasses: 'font-weight-medium',
```
dchiquit, line 129: These work and are fine, but they seem just a tad smelly to me. Having presentational code like this in the component logic feels like violating some separation of concerns. I do see the benefit of not repeating these classes in every row definition and how that makes it easier to change later.
The only alternative I could think of to keep this tidy would be to make another component specifically for data rows. Do you think that's overkill?
To be clear I'm willing to accept this, I'm just thinking of alternatives.

AlmightyYakob: I think having a separate component just for this is a bit overkill, I see your point though. Would scoping these fields under a `style` field make it a bit cleaner?

web/src/views/DandisetLandingView/DandisetLandingView.vue
AlmightyYakob: cc015b7

web/src/views/DandisetLandingView/DandisetDetails.vue
dchiquit: Not really, but if it makes you feel better go for it ¯\_(ツ)_/¯

AlmightyYakob: Alright, well since it doesn't look like we have a solution to this right now, I think we can merge this and deal with that later. Also you dropped this  \

PR #353, opened 2020-05-22
Make tests run independently

### LINKED ISSUE #344: Ensure browser tests are set up independently
I wrote a few tests that share some state between `it`s. Jest supports this, but it is generally a bad practice and this should be refactored out.

```
I think if there is a place to fudge this rule, it would be browser tests. Test setup is generally done through the UI, which is good, but also can be very slow. Having to to do lengthy setup steps for lots of really granular assertions can severely impact run time.
In this case, I think you're right, I need to refactor these to avoid intertest dependencies. Jest almost seems to encourage that pattern, but I should have known better.
```
_Originally posted by @dchiquit in https://github.com/dandi/dandiarchive/pull/328_

dchiquit:
I made some tests dependent on the browser state left behind by previous tests. This makes tests run faster as they don't have to set everything up every time, but also poisons the rest of the tests in a file when the first test in the chain fails.
This PR starts all tests already at the home page with a fresh state. Tests must now initialize themselves.
Fixes #344


test/src/tests/account.test.js
```diff
@@ -5,18 +5,14 @@ import {
...
-
-  beforeAll(async () => {
-    await page.goto(CLIENT_URL);
-  });
-
   it('registers a new user', async () => {
+    const username = `user${uniqueId()}`;
+    const email = `${username}@dandi.test`;
+    const password = 'password'; // Top secret
+
```
AlmightyYakob, line 15: It seems like much of this test overlaps with the `registerNewUser` utility function. Any reason not be using that here?

test/jest.setup.js
```diff
@@ -1,6 +1,26 @@
...
+
+// this is not redundant
+// it allows tests to be logged in already before doing setup in their own beforeAlls
+// eslint-disable-next-line jest/require-top-level-describe
+beforeAll(async () => {
+  await Promise.all([
+    page.goto(CLIENT_URL),
+    page.waitForNavigation({ waitUntil: 'networkidle0' }),
+  ]);
+});
```
AlmightyYakob, line 17: By logged in here, do you mean simply navigated to the page with everything loaded? I don't see any logging in going on here.

test/jest.setup.js
```diff
@@ -1,6 +1,26 @@
...
+
+// eslint-disable-next-line jest/require-top-level-describe
+beforeEach(async () => {
+  await jestPuppeteer.resetBrowser();
+  await Promise.all([
+    page.goto(CLIENT_URL),
+    page.waitForNavigation({ waitUntil: 'networkidle0' }),
+  ]);
+});
```
AlmightyYakob, line 26: I'm confused as to the reason the above `beforeAll` is needed if this is also here? Wouldn't this cover all of the necessary functionality of the `beforeAll`?

test/jest.setup.js
dchiquit: I mistyped, I meant `already on the webpage`. I will refactor that docblock

test/jest.setup.js
dchiquit: This is interestingly arcane, and I intended the documentation on the `beforeAll` to explain it. I will rework that more thoroughly.
Basically jest runs `before` and `after` functions in this order:
* global `beforeAll`, like this one
* `beforeAll` in the test file
* global `beforeEach`
* `beforeEach` in the test file
* the test code
* `afterEach` and `afterAll`, in reverse of the `before` functions
Tests should be able to assume that before each test method, the browser is a clean slate and is already sitting on the home page. The problem is that some tests also might want to set stuff up in a `beforeAll` (like creating a single dandiset so it can be searched for in a search test). According to the above run order, we also need a global `beforeAll` to start the test in the home page. Otherwise test writers will need to remember that `beforeEach` and `beforeAll` have different initial conditions, which is dumb.

test/src/tests/account.test.js
dchiquit: This is a testing methodology thing. My thinking is that helpers and utilities should be treated like black boxes which magically change the app state. Tests need to explicit about what they're testing, so they should never be using helpers or utilities to perform the actions being tested. Helpers should only be used to set up or tear down test data.
In this case, the helper and the test code are basically the same, but that won't necessarily be the case for other helpers, and it's possible that the helper will be changed in the future. Helpers should be free to take whatever shortcuts they can to speed up test set up, and no tests should break if a helper changes (assuming the helper still does what it is advertised to do).
The code duplication is sad, but IMHO it's necessary.

brianhelba: Agreed!

test/jest.setup.js
brianhelba: Makes sense. This should definitely be a code comment.

PR #359, opened 2020-05-28
Clean up and document jest-puppeteer-vuetify dependencies

brianhelba:


brianhelba:
@dchiquit I ended up completely re-working the approach here. I previously failed to realize that `page` was a function parameter provided explicitly by the dispatch to the expect functions.

So, only `jest` is strictly required, and we can explicitly depend on its globals as an import. See the code comment for a more detailed discussion of this. Ultimately, I think this will be a better approach for an independent package.

Could you take a look again?

PR #373, opened 2020-06-03
Add dandiset stats

### LINKED ISSUE #372: Add per-dandiset stats endpoint and frontend use
Following up from this comment: https://github.com/dandi/dandiarchive/issues/38#issuecomment-637828542

This endpoint would return the following about a specific dandiset:

- number of folders
- number of items
- number of bytes
- ~~number of items with 0 files~~

AlmightyYakob:
Fixes #372

### DIFF
--- a/web/src/views/DandisetLandingView/DandisetDetails.vue
+++ b/web/src/views/DandisetLandingView/DandisetDetails.vue
@@ -188,10 +188,10 @@ export default {
     },
     formattedSize() {
       const { stats } = this;
-      if (!stats) return undefined;
-
-      const a = filesize(stats.bytes);
-      return a;
+      if (!stats) {
+        return undefined;
+      }
+      return filesize(stats.bytes);
     },
     ...mapState('girder', {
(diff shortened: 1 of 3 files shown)


web/src/views/DandisetLandingView/DandisetDetails.vue
```diff
@@ -152,10 +186,24 @@ export default {
 
       return null;
     },
+    formattedSize() {
+      const { stats } = this;
+      if (!stats) return undefined;
+
+      const a = filesize(stats.bytes);
+      return a;
```
dchiquit, line 194:
```suggestion
      if (!stats) {
        return undefined;
      }
      return filesize(stats.bytes);
```
Java style guides almost universally ban the one line if statement, so I habitually avoid whenever possible. It's not terribly important to me, but I see no reason for the `a`.

AlmightyYakob: Whoops, that was left over from me quickly testing it. Good catch.

PR #376, opened 2020-06-03
NF: rudimentary .travis.yml to check that docker-compose build/up/down works

yarikoptic:
As long as we have this passing - would already be worth merging.

Someone alternatively could better workout github actions or circle-ci since I believe (although didn't check in travis recently) they would allow for jobs dependencies so we could build docker-compose as a common step, and then fan out into testing client etc.

Meanwhile after this one is done -- we would need to add running dandi-cli tests against this local instance for download/upload cycle

For some reason travis says ATM

This is not an active repository
You don't have sufficient rights to enable this repo on Travis.
Please contact the admin to enable it or to receive admin rights yourself.

so I (or someone else) might need to dig out where permissions are due for travis.  In my account I see travis CI for Open  Source enabled for dandi -- so I do not quite get what else it wants


yarikoptic:
@satra - may be your magic touch could make travis behave?

jwodder:
I have converted the workflow from Travis to GitHub Actions and have added a test to check that the "provision" container succeeds.  This test is currently passing when run on the branch but failing when merged into master due to the still-inexplicable non-installation of `girder_dandi_archive` that I mentioned in #494.

yarikoptic:
I have restarted the tests after the fix was merged - green now for GitHub workflows. Circle isn't happy but it is unrelated afaik.

IMHO it is worth merging.  dandi cli tests on top in a follow up PR

jwodder:
> I have restarted the tests after the fix was merged

What fix?  There are no new commits to dandiarchive's master.

yarikoptic:
Have I got confused by the fix in https://github.com/dandi/dandi-publish/pull/144 (too many pieces without integration testing)? But the point is that it is green now, and should stay green! ;-)

waxlamp:
Closing since the docker container infrastructure will no longer be hosted in this repo.

yarikoptic:
yeah, I guess we would eventually want some dedicated "dandi-standalone" repo with docker-compose which would bring all components together. `dandi-api` would have only the one for itself.

PR #392, opened 2020-06-10
Upgrade web client dependencies

brianhelba:


AlmightyYakob:
Although I do see this warning when serving it locally, regarding dependencies of `moment`.

```
WARNING  Compiled with 1 warnings                                                             6:48:09 PM

warning  in ./node_modules/moment/locale/en-SG.js

There are multiple modules with names that only differ in casing.
This can lead to unexpected behavior when compiling on a filesystem with other case-semantic.
Use equal casing. Compare these module identifiers:
* /home/jake/projects/dandiarchive/web/node_modules/moment/locale/en-SG.js
    Used by 2 module(s), i. e.
    /home/jake/projects/dandiarchive/web/node_modules/moment/locale sync /^\.\/.*$/
* /home/jake/projects/dandiarchive/web/node_modules/moment/locale/en-sg.js
    Used by 2 module(s), i. e.
    /home/jake/projects/dandiarchive/web/node_modules/moment/locale sync /^\.\/.*$/
```

However there's not really much we can do about this, and I don't think this should affect us.

PR #394, opened 2020-06-11
Exclude superfluous Moment locales

brianhelba:
This reduces the final bundle size by ~15%.


AlmightyYakob:
> Seems to work for me

To clarify, took a look at the two places this is used (`DandisetDetails.vue` and `DandisetList.vue`), and both seem to behave as expected.

PR #397, opened 2020-06-12
Display stats under item title

### LINKED ISSUE #396: stats feature
@AlmightyYakob - love the stats feature.

here just suggesting a potential change. in the public datasets dashboard page, could the files and size be added to the subtitle line. right after the identifier?  (i think that will also help with the wrapping on smaller devices)

image

AlmightyYakob:
Fixes #396 

Additional changes:
- Change the ordering of subtitle items
- Change the size icon from `server` to `database`

### DIFF
--- a/web/src/components/DandisetList.vue
+++ b/web/src/components/DandisetList.vue
@@ -44,45 +44,25 @@
               Contact <b>{{ getDandisetContact(item) }}</b>
               ·
               Updated on <b>{{ formatDate(item.updated) }}</b>
-              <template v-if="$vuetify.breakpoint.mdAndDown">
-                ·
-                <v-icon
-                  small
-                  class="pb-1"
-                >
-                  mdi-file
-                </v-icon>
-                {{ dandisetStats[i].items }}
-                ·
-                <v-icon
-                  small
-                  class="pb-1"
-                >
-                  mdi-database
-                </v-icon>
-                {{ filesize(dandisetStats[i].bytes) }}
-              </template>
+              ·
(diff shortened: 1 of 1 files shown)


satra:
looks great. thank you @AlmightyYakob

satra:
@dchiquit - where are you seeing units being wrapped around? this is i think what's fixed in this PR?

AlmightyYakob:
It occurs at very specific breakpoints, and was already there before this change. I wasn't exactly sure how to fix it, I'm not too good with small styling details like that. If we think this is a blocker I can try to fix it before merging.

mgrauer:
I much prefer the style where the stats are displayed along with the other metadata as the dandiset subtitle, I think it is easier to see which stats go with which dandiset.

Screen Shot 2020-06-13 at 11 34 31 AM

With the separated stats, I have a hard time tracking which stats go with which dandiset, and I find it jarring that the stats move between UI elements at different breakpoints

Screen Shot 2020-06-13 at 11 41 16 AM



So I propose removing the breakpoint and only having the stats appear in the dandiset subtitle. This should also address @dchiquit 's concern.

AlmightyYakob:
I agree with @mgrauer's comment. As long as it's okay with @satra, I can remove the right side stats and just display them on the subtitle

satra:
@AlmightyYakob - ha! difference of field of view. i didn't even go that far out - that's why i could not understand what @dchiquit was saying. yes, no need to float those stats out. they can just be in the subtitle.

PR #398, opened 2020-06-12
Add a locking service to girder

dchiquit:
Add two endpoints, `dandi/:identifier/lock` and
`dandi/:identifier/unlock`. These will lock or unlock dandisets for the
current user. When unlocked, a dandiset can be modified by anyone with
permission. When locked, a dandiset can be modified only by the person
who owns the lock.

Locks are enforced using event listeners on folder save and assetstore
upload events. Modifying a dandiset's metadata or uploading new files to
it will fail if another user has locked the dandiset.

The current lock status is stored in a property on the root folder
model. We may eventually want to add an endpoint to return the current
lock status so it can be displayed in the web client.

### DIFF
--- a/girder-dandi-archive/girder_dandi_archive/__init__.py
+++ b/girder-dandi-archive/girder_dandi_archive/__init__.py
@@ -1,10 +1,12 @@
 # -*- coding: utf-8 -*-
+from girder import events
 from girder.models.folder import Folder
 from girder.models.setting import Setting
 from girder.plugin import GirderPlugin
 from girder.settings import SettingKey
 from girder_user_quota.settings import PluginSettings as UserQuotaPluginSettings
 
+from . import locking
 from .rest import DandiResource
 from .util import DANDISET_IDENTIFIER_COUNTER
 
@@ -28,3 +30,14 @@ class DandiArchivePlugin(GirderPlugin):
         Setting().set(UserQuotaPluginSettings.DEFAULT_USER_QUOTA, 0)
 
         info["apiRoot"].dandi = DandiResource()
+
+        events.bind("model.folder.save", "folder_save_listener", locking.folder_save_listener)
+        # TODO item save
+        # TODO file save
+        events.bind(
+            "model.upload.assetstore",
+            "upload_assetstore_listener",
+            locking.upload_assetstore_listener,
+        )
+        # TODO file delete
+        # TODO file move
--- a/girder-dandi-archive/girder_dandi_archive/__init__.py
+++ b/girder-dandi-archive/girder_dandi_archive/__init__.py
@@ -32,12 +32,13 @@ class DandiArchivePlugin(GirderPlugin):
         info["apiRoot"].dandi = DandiResource()
 
         events.bind("model.folder.save", "folder_save_listener", locking.folder_save_listener)
-        # TODO item save
-        # TODO file save
+        events.bind("model.item.save", "item_save_listener", locking.item_save_listener)
+        events.bind("model.item.remove", "item_remove_listener", locking.item_remove_listener)
         events.bind(
             "model.upload.assetstore",
             "upload_assetstore_listener",
             locking.upload_assetstore_listener,
         )
-        # TODO file delete
+        # TODO file save
+        # TODO file remove
         # TODO file move
--- a/girder-dandi-archive/girder_dandi_archive/locking.py
+++ b/girder-dandi-archive/girder_dandi_archive/locking.py
@@ -13,7 +13,7 @@ def _get_lock(identifier):
     If there is no lock, None is returned.
     """
     dandiset_folder = find_dandiset_by_identifier(identifier)
-    if "locked" not in dandiset_folder:
+    if dandiset_folder is None or "locked" not in dandiset_folder:
         return None
     return str(dandiset_folder["locked"])
 
--- a/girder-dandi-archive/girder_dandi_archive/locking.py
+++ b/girder-dandi-archive/girder_dandi_archive/locking.py
@@ -124,6 +124,18 @@ def folder_save_listener(event):
         pass
 
 
+def _get_dandiset_for_file(resource):
+    """Get the root dandiset folder which the given file belongs to."""
+    # TODO assuming that the file being uploaded is in a folder
+    subfolder = Folder().findOne(resource["folderId"])
+    # Dandisets are always top level folders.
+    # Ascend the file tree until the subfolder is the root collection.
+    # After the loop completes, subfolder will be the dandiset containing the resource.
+    while subfolder["parentCollection"] != "collection":
+        subfolder = Folder().findOne(subfolder["parentId"])
+    return subfolder
+
+
 def upload_assetstore_listener(event):
     """
     Listen to upload assetstore events to enforce locking.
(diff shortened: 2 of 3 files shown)


girder-dandi-archive/girder_dandi_archive/rest.py
```diff
@@ -326,3 +331,35 @@ def stats(self, params):
...
+        Description("The owner of the lock on a Dandiset").param(
+            "identifier", "Dandiset Identifier", paramType="path"
+        )
+    )
+    @dandiset_identifier
+    def get_dandiset_lock_owner(self, identifier, params):
+        owner = locking.get_lock_owner(identifier)
+        if owner is None:
+            return
+        print("lock owned by", owner)
```
AlmightyYakob, line 364: I'd prefer if the return was explicit here, and the print statement removed:

```suggestion
            return None
```

girder-dandi-archive/girder_dandi_archive/__init__.py
```diff
@@ -28,3 +30,14 @@ def load(self, info):
...
+        events.bind("model.folder.save", "folder_save_listener", locking.folder_save_listener)
+        # TODO item save
+        # TODO file save
+        events.bind(
+            "model.upload.assetstore",
+            "upload_assetstore_listener",
+            locking.upload_assetstore_listener,
+        )
+        # TODO file delete
+        # TODO file move
```
AlmightyYakob, line 43: Are these TODOs still being worked on in this PR? And if not or if you plan on adding them in a later PR, we should make an issue for them.

girder-dandi-archive/girder_dandi_archive/rest.py
```diff
@@ -42,6 +43,10 @@ def __init__(self):
         self.route("POST", (), self.create_dandiset)
         self.route("POST", (":identifier",), self.publish_dandiset)
         self.route("GET", ("stats",), self.stats)
+        self.route("POST", (":identifier", "lock"), self.lock_dandiset)
+        self.route("POST", (":identifier", "unlock"), self.unlock_dandiset)
+        self.route("GET", (":identifier", "lock", "owner"), self.get_dandiset_lock_owner)
+        # TODO add a way for admins to remove dead locks
```
AlmightyYakob, line 49: Same thing as mentioned above with this TODO

girder-dandi-archive/girder_dandi_archive/locking.py
```diff
@@ -0,0 +1,143 @@
...
+    """
+    Listen to upload assetstore events to enforce locking.
+
+    These events occur before files are uploaded.
+    If the root dandiset folder has been locked by someone other than the current user,
+    an exception is thrown.
+    """
+    thing = event.info["resource"]
+    while thing["parentCollection"] != "collection":
+        thing = Folder().findOne(thing["parentId"])
```
AlmightyYakob, line 137: I think some comments here would be useful, to describe what's going on to someone looking at it for the first time.

yarikoptic:
I wondered if you anyone had tried `dandi upload` for a locked dataset -- how would it fail (I hope it would ;))?

girder-dandi-archive/girder_dandi_archive/locking.py
```diff
@@ -0,0 +1,143 @@
...
+    try:
+        identifier = thing["meta"]["dandiset"]["identifier"]
+        require_access(identifier, getCurrentUser())
+    except KeyError:
+        # This event is on a non-dandiset folder, no locking is required
+        pass
```
yarikoptic, line 143: I am not girder API savvy and generally paranoid!  All the `["key"]` (instead of e.g. `.get()`) and then just silently not locking upon `KeyError` suggest that this code would simply stop locking anything, without any log message, whenever some assumptions would change (e.g., "meta" starts to carry "dadaset" instead of "dandiset", or "identifier" becomes "id").  Why not at least to log some debug message or even error out so developers could see something going wrong or unexpected?  would code request lock on a folder which is not a dandiset -- shouldn't it inform about that?

girder-dandi-archive/girder_dandi_archive/rest.py
yarikoptic: and who sees those `print`s?

girder-dandi-archive/girder_dandi_archive/__init__.py
dchiquit: The ticket was for the bare minimum required for a locking solution, so I deliberately didn't implement these things yet. As I understand it this is a prototype that we don't necessarily want to invest too much time and effort into, so as long as it works well enough we don't need to guarantee perfect operation in all conditions.

girder-dandi-archive/girder_dandi_archive/rest.py
dchiquit: This is leftover debug code, I will remove.

dchiquit:
@yarikoptic This is what happens when you try to upload when the lock is held by someone else:
```
$ dandi upload -i local-girder-only -e force
2020-06-16 11:17:12,547 [    INFO] Found 54 files to consider
PATH                                         SIZE     ERRORS UPLOAD STATUS                        MESSAGE                                                  
dandiset.yaml                                1.1 kB     0           ERROR                         Dandiset 000006 is currently locked by Daniel Chiquito   
sub-anm369964/sub-anm369964_ses-20170322.nwb 5.0 MB     0           ERROR                         Dandiset 000006 is currently locked by Daniel Chiquito   
sub-anm369964/sub-anm369964_ses-20170320.nwb 7.7 MB     0           ERROR                         Dandiset 000006 is currently locked by Daniel Chiquito   
...
sub-anm369963/sub-anm369963_ses-20170306.nwb 6.7 MB     0           ERROR                         Dandiset 000006 is currently locked by Daniel Chiquito   
sub-anm369963/sub-anm369963_ses-20170310.nwb 604.8 kB   0           ERROR                         Dandiset 000006 is currently locked by Daniel Chiquito   
Summary:                                     144.1 MB               54 ERROR                      54 Dandiset 000006 is currently locked by Daniel Chiquito
```
If the user doing the upload has the lock, the upload works fine.

mgrauer:
@yarikoptic to discuss this at a higher level, we wanted to get something basic in place, and then engage with you to start adding API calls to the CLI to use the locking system, and that would flush out bugs and clarify the usage. That should happen in a follow on PR.

girder-dandi-archive/girder_dandi_archive/locking.py
dchiquit: This was copied from the `folder_save_listener` method. It makes sense there, but it is not needed here. I have removed it.

dchiquit:
I did find some bugs where the dandi-cli would delete an item so it could reupload it, then encounter the lock when trying to upload, resulting in the item being deleted even though it was locked. There are now handlers for item save and item remove.

Hopefully though, these checks should be redundant safety measures once the CLI acquires locks before modifying a dandiset. If all clients correctly acquire locks before modifying dandisets, these errors should never be thrown. It's still useful that they exist, they just shouldn't be encountered often and shouldn't be invested too heavily in.

girder-dandi-archive/girder_dandi_archive/locking.py
```diff
@@ -124,6 +124,18 @@ def folder_save_listener(event):
         pass
 
 
+def _get_dandiset_for_file(resource):
```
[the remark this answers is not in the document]
dchiquit: I think I meant `item`, I will rename it

yarikoptic:
Thank you @dchiquit for trying dandi-cli and accounting for encountered issues!

PR #400, opened 2020-06-17
Display lock status on the DLP

dchiquit:
If a dandiset is locked, the `Edit Metadata` and `Publish` buttons are
now disabled. If the lock owner has the email address
`publish@dandiarchive.org`, the tooltip is `A publish is currently in
progress`. Otherwise, the tooltip is `This dandiset is currently locked
by John Doe`, with the propert first/last name substituted. The buttons
are disabled by default until the lock status can be fetched.


dchiquit:
I don't think we would need to know the lock owner anywhere but here, so I don't know what that argument would be

PR #403, opened 2020-06-18
Add initial content browser for published dandisets

AlmightyYakob:
Notable changes:
- This replaces the use of a Girder id in the URL, with the dandiset identifier.
- FileBrowser.vue is now split into two components, one for published dandisets and one for drafts.

Some open questions:

- Should we remove the `..` navigation? Breadcrumbs technically facilitates its functionality, but I personally still like having it. Not sure how others feel.
- Does the breadcrumb area look alright?
- Would it makes sense to have a version switcher at the top right of the browser? It may be useful to some people, however I'm not sure if it's a good idea to give the user that much control in this specific section of the site.

### DIFF
--- a/web/src/components/DandisetList.vue
+++ b/web/src/components/DandisetList.vue
@@ -7,7 +7,7 @@
       v-for="(item, i) in items"
       :key="item._id"
       selectable
-      :to="{ name: 'dandisetLanding', params: { id: item._id, origin } }"
+      :to="{ name: 'dandisetLanding', params: { identifier: item.meta.dandiset.identifier, origin } }"
     >
       <v-row
         no-gutters
--- a/web/src/components/DandisetList.vue
+++ b/web/src/components/DandisetList.vue
@@ -7,7 +7,10 @@
       v-for="(item, i) in items"
       :key="item._id"
       selectable
-      :to="{ name: 'dandisetLanding', params: { identifier: item.meta.dandiset.identifier, origin } }"
+      :to="{
+        name: 'dandisetLanding',
+        params: { identifier: item.meta.dandiset.identifier, origin }
+      }"
     >
       <v-row
         no-gutters
--- a/web/src/router.js
+++ b/web/src/router.js
@@ -21,11 +21,6 @@ export default new Router({
       component: HomeView,
     },
     {
-      path: '/file-browser/:_modelType?/:_id?',
-      name: 'file-browser',
-      component: FileBrowser,
-    },
-    {
       path: '/user/login',
       name: 'userLogin',
       component: UserLoginView,
@@ -56,7 +51,13 @@ export default new Router({
       component: CreateDandisetView,
     },
     {
-      path: '/dandiset/:id/:version?',
+      path: '/dandiset/:identifier/:version/files',
+      name: 'fileBrowser',
+      props: true,
+      component: FileBrowser,
+    },
+    {
+      path: '/dandiset/:identifier/:version?',
       name: 'dandisetLanding',
       props: true,
       component: DandisetLandingView,
--- a/web/src/store/dandiset.js
+++ b/web/src/store/dandiset.js
@@ -1,4 +1,5 @@
 import { girderRest, publishRest } from '@/rest';
+import { draftVersion } from '@/utils';
 
 export default {
   namespaced: true,
@@ -7,6 +8,11 @@ export default {
     girderDandiset: null,
     loading: false,
     owners: null,
+  },
+  getters: {
+    version(state) {
+      return state.publishDandiset ? state.publishDandiset.version : draftVersion;
+    },
   },
   mutations: {
     setGirderDandiset(state, dandiset) {
@@ -28,10 +34,10 @@ export default {
 
       state.loading = false;
     },
-    async fetchGirderDandiset({ state, commit }, { girderId }) {
+    async fetchGirderDandiset({ state, commit }, { identifier }) {
       state.loading = true;
 
-      const { data } = await girderRest.get(`folder/${girderId}`);
+      const { data } = await girderRest.get(`dandi/${identifier}`);
       commit('setGirderDandiset', data);
 
       state.loading = false;
--- a/web/src/utils.js
+++ b/web/src/utils.js
@@ -2,8 +2,8 @@ const dandiUrl = 'https://dandiarchive.org';
 const dandiAboutUrl = 'https://dandiarchive.org/about';
 
 function getLocationFromRoute(route) {
-  const { _modelType, _id } = route.params;
-  if (_modelType) {
+  const { _modelType, _id } = route.query;
+  if (_modelType && _id) {
     return { _modelType, _id };
   }
   return null;
--- a/web/src/views/CreateDandisetView/CreateDandisetView.vue
+++ b/web/src/views/CreateDandisetView/CreateDandisetView.vue
@@ -70,9 +70,11 @@ export default {
         },
       });
 
+
+      const { identifier } = data.meta.dandiset;
       this.$router.push({
         name: 'dandisetLanding',
-        params: { id: data._id },
+        params: { identifier },
       });
     },
   },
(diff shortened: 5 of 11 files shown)


mgrauer:
>Should we remove the .. navigation? Breadcrumbs technically facilitates its functionality, but I personally still like having it. Not sure how others feel.

Leave it in for now.

>This replaces the use of a Girder id in the URL, with the dandiset identifier.

Does this change the URL for draft dandisets? 

>Would it makes sense to have a version switcher at the top right of the browser? It may be useful to some people, however I'm not sure if it's a good idea to give the user that much control in this specific section of the site.

Don't do this for now, because the additional complexity. I expect there will be many corner cases here.


Also, I don't remember from when you showed this to me, when we are content browsing in a published dandiset, do we get to see the dandiset identifier, name, and version somewhere?

AlmightyYakob:
> > Should we remove the .. navigation? Breadcrumbs technically facilitates its functionality, but I personally still like having it. Not sure how others feel.
> 
> Leave it in for now.

:+1: 

> 
> > This replaces the use of a Girder id in the URL, with the dandiset identifier.
> 
> Does this change the URL for draft dandisets?

Yes, for all dandisets. Old url was:
```
https://gui.dandiarchive.org/#/dandiset/<girder_id>
```

new one would be:
```
https://gui.dandiarchive.org/#/dandiset/<dandi_identifier>
```

> > Would it makes sense to have a version switcher at the top right of the browser? It may be useful to some people, however I'm not sure if it's a good idea to give the user that much control in this specific section of the site.
> 
> Don't do this for now, because the additional complexity. I expect there will be many corner cases here.

:+1: 

> Also, I don't remember from when you showed this to me, when we are content browsing in a published dandiset, do we get to see the dandiset identifier, name, and version somewhere?

Nope, that's not in there currently. However it should be easy to add, if it's a blocker.

mgrauer:
@AlmightyYakob do you include draft or version as suffixes to `https://gui.dandiarchive.org/#/dandiset/<dandi_identifier>` ? Could you provide examples of how they look if so?

AlmightyYakob:
> @AlmightyYakob do you include draft or version as suffixes to `https://gui.dandiarchive.org/#/dandiset/<dandi_identifier>` ? Could you provide examples of how they look if so?

Yes, I believe this was done in #395. So the draft of a dandiset would be something like:
```
https://gui.dandiarchive.org/#/dandiset/000001/draft
```

whereas a versioned example would be:
```
https://gui.dandiarchive.org/#/dandiset/000001/0.200605.1956
```

Navigation to an unspecified version like so:
```
https://gui.dandiarchive.org/#/dandiset/000001
```

Will redirect you to the draft version shown above.

mgrauer:
Given the new URL routes, we'll want to coordinate with @yarikoptic for the CLI.

satra:
> Navigation to an unspecified version like so:
> `https://gui.dandiarchive.org/#/dandiset/000001`
> Will redirect you to the draft version shown above.

can this redirect you to the latest published version if available, and if not to the draft version?

AlmightyYakob:
> > Navigation to an unspecified version like so:
> > `https://gui.dandiarchive.org/#/dandiset/000001`
> > Will redirect you to the draft version shown above.
> 
> can this redirect you to the latest published version if available, and if not to the draft version?

Yes, that'll be easy to add.

mgrauer:
As a follow up, we cannot guarantee long term stability of our urls on the client, and these should not be depended on as if they are a stable API.

So I'd like to work with @yarikoptic to update the CLI to move towards the URLs such as these in the near term:

- `https://gui.dandiarchive.org/#/dandiset/000001/0.200605.1956`
- `https://gui.dandiarchive.org/#/dandiset/000001/draft`

and away from girder internal ID based URLs.  We will want to do this before we go live with the publish features, or else Yarik can tell us it's ok to go live without updating the CLI.

What would be even better (either near term or longer term) would be to have a contract between the server and the CLI such that we have CLI commands like  `download dandiset 000001 draft` or `download dandiset 000001 0.200605.1956`, and the UI could expose the CLI command to run, then the user could copy and run that command and the CLI would not have to consume URLs. That way the URLs could change and the CLI won't have to be updated. This could potentially be a good hackathon item.

yarikoptic:
> * `https://gui.dandiarchive.org/#/dandiset/000001/0.200605.1956`
> * `https://gui.dandiarchive.org/#/dandiset/000001/draft`

And `#` in those really a "fragment", so processed by client side (JS?)?  I think that getting away from girder IDs is the step in the right direction.  But I still believe that some bright future will come when I will see a simple `https://dandiarchive.org/dandiset/000001/0.200605.1956` where there is "pure information" (server is `dandiarchive.org`,  dandiset `000001`, and version `0.200605.1956`) and no implementation specifics anyhow in the url. So, again, proposed urls are a step in the right direction as to me ;-)

yarikoptic:
> download dandiset 000001 0.200605.1956

we could work something like that out.  But once again -- I would like to clear "cut the url from browser into client and download that thing", instead of more ad-hoc (although possibly more stable) shortcuts. And that is why having clear descriptive URLs would be of help here, and parallel existing other systems (e.g. github etc)

mgrauer:
> So, again, proposed urls are a step in the right direction as to me ;-)

>- https://gui.dandiarchive.org/#/dandiset/000001/0.200605.1956
>- https://gui.dandiarchive.org/#/dandiset/000001/draft

To double check for near term actions, @yarikoptic, if we move the production website to have these URLs that is ok for the CLI, or do you need to do some work first? Should we coordinate the release of the production website with you, or is it ok to release whenever we are ready with the website updates?

satra:
@mgrauer - this also has impact on the redirector as well. will you send a PR to that as well to coincide with this. and could we take this as an opportunity to fold the redirector into the main archive and deploy from there as well?

yarikoptic:
yeap... I guess I need to fire up a docker-compose on this branch now and tune up -cli for that?  or should I wait for testing to turn green first? Also it might be good time to introduce development instance so we could have PRs merged, deployed (into a dedicated instance), and then -cli  tuned up (formalized in #407)

mgrauer:
>will you send a PR to that as well to coincide with this.

done

>and could we take this as an opportunity to fold the redirector into the main archive and deploy from there as well?

that is more scope than i'm ready for right now with all of the moving parts going on with publish--i'd want to take some time to think about the right way to arrange this. what would be the goal of moving the redirector? and can you remind me who is using the permalinks currently?

@yarikoptic 

>I guess I need to fire up a docker-compose on this branch now and tune up -cli for that? 

yes, i don't think you need to wait for these tests to pass in order to start this work. 

i agree with the desire for a staging instance, but that is too much to manage right now with all of the other moving parts. when publish settles down i think we can reconvene and work through some of these ideas. we will soon have the girder production backend updated for lock, and you can test against that.

you might be able to test against netlify preview, though i don't understand exactly what testing you need to do. my assumption is that you only need to read/parse URLs from the GUI and update those URLs you are reading in, you shouldn't be changing any of the API endpoints that you call to upload files or metadata. here is a preview URL example: `https://deploy-preview-403--gui-dandiarchive-org.netlify.app/#/dandiset/000004/draft`

satra:
> what would be the goal of moving the redirector? 

the redirector is our current entrypoint for all permalinks and identifiers.org so it's part of the same infrastructure and it internally responds in relation to how it will redirect to gui.dandiarchive.org. so it's as coupled as the cli, and should all be tested together whenever changes are made on gui. 

> and can you remind me who is using the permalinks currently?

- the cli
- any user who is putting it in publications either as the dandi url or the identifiers.org url

AlmightyYakob:
> > Navigation to an unspecified version like so:
> > `https://gui.dandiarchive.org/#/dandiset/000001`
> > Will redirect you to the draft version shown above.
> 
> can this redirect you to the latest published version if available, and if not to the draft version?

@satra It seems this would actually be better suited address in a follow up PR. I've made issue #410 for tracking it.

PR #404, opened 2020-06-18
Fix dandiset navigation

AlmightyYakob:
Depends on #403.

Fixes a bug I found where clicking on a dandiset from the dandiset listing page would display the version of that dandiset that was previously selected (instead of starting with draft) due to `publishDandiset` not being set to `null` after leaving the DLP.

### DIFF
--- a/web/src/components/DandisetList.vue
+++ b/web/src/components/DandisetList.vue
@@ -7,10 +7,7 @@
       v-for="(item, i) in items"
       :key="item._id"
       selectable
-      :to="{
-        name: 'dandisetLanding',
-        params: { identifier: item.meta.dandiset.identifier, origin }
-      }"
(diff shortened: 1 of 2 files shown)


mgrauer:
@AlmightyYakob Can you fix up this PR and then we can try to get it merged?

AlmightyYakob:
> @AlmightyYakob Can you fix up this PR and then we can try to get it merged?

This should be good to go now.

mgrauer:
@dchiquit Can you review when you get a chance?

AlmightyYakob:
@dchiquit discovered that this bug doesn't seem to be occurring anymore, and when #410 is addressed these changes will be moot anyway, so I'm closing this PR.

PR #406, opened 2020-06-22
Add backend changes from publish branch

dchiquit:


mgrauer:
@dchiquit My assumptions are that this has the backend for sync and lock, and those have both already been QA-ed separately and merged into the publish branch. If this is all true, then this PR should have also passed QA already and be ok to merge, and in that case I'll approve it.

PR #409, opened 2020-06-24
Add typescript to jest-puppeteer-vuetify

dchiquit:


test/jest-puppeteer-vuetify/package.json
```diff
@@ -1,9 +1,16 @@
...
-  "main": "src/index.js",
+  "description": "Test utilities to make locating Vuetify components easier",
+  "main": "lib/index.js",
+  "scripts": {
+    "build": "tsc"
+  },
   "dependencies": {
-    "@jest/globals": "*"
+    "@jest/globals": "*",
+    "@types/puppeteer": "^3.0.0",
```
brianhelba, line 11: Do you foresee consumers also being written in TS? If so, they'd need `@types/puppeteer`, so it should be in `dependencies`. If consumers will just import the transpiled JS version of this, then `@types/puppeteer` should be a `devDependency` (as [endorsed by `DefinitelyTyped`](https://github.com/DefinitelyTyped/DefinitelyTyped#npm))

test/jest-puppeteer-vuetify/src/vuetify-xpaths.ts
```diff
@@ -0,0 +1,174 @@
+type XPath = string;
+type Predicate = string;
+type Strings = string | Strings[] | undefined;
```
brianhelba, line 3:
```suggestion
type Strings = string | Strings[];
```

With `compilerOptions.strictNullChecks` (which you've enabled via `compilerOptions.strict`), use of the `?.` (which you seem to be using in all of the `Strings`-typed parameters) will automatically include `undefined`.

test/jest-puppeteer-vuetify/package.json
dchiquit: I think it be unnecessarily prescriptive to require TS, so I will move this to a devDependency. Generally I think TS is less useful in test code, as there are much fewer moving parts, and the code being written is generally very linear and easy to read.

test/jest-puppeteer-vuetify/src/vuetify-xpaths.ts
dchiquit: I copied this from some online resource that mixed the two. If `?=` is the preferred standard syntax, I will use it.

dchiquit: I turned on all the strictest things because I couldn't think of a reason not to, especially when starting a fresh project like this. In this case, I don't mind having both the `type Strings = ... | undefined` and the `?:`, because methods using the spread syntax apparently do not work with the `?:` syntax:
```
function classAsPredicate(...classes: Strings[]): Predicate {...}
```
Even if it's technically optional, I would rather have that `undefined` typed in to `Strings`.

dchiquit:
I've migrated jest-puppeteer-vuetify to https://github.com/dchiquit/jest-puppeteer-vuetify, so this review is no longer meaningful. Closing this.

PR #411, opened 2020-06-25
Default to latest published version when version not specified

### LINKED ISSUE #410: Default to latest published version when no version specified, if possible
Follow up from https://github.com/dandi/dandiarchive/pull/403#issuecomment-647758130

If no version specified, and there are no published versions, default back to draft.

AlmightyYakob:
Fixes #410.

Since this feature required changing how/when specific fetches/state changes are being made, much of the change here is to keep things organized and easy to follow.

### DIFF
--- a/web/src/store/dandiset.js
+++ b/web/src/store/dandiset.js
@@ -38,6 +38,8 @@ export default {
       state.loading = false;
     },
     async initializeDandisets({ state, dispatch }, { identifier, version }) {
+      await dispatch('uninitializeDandisets');
+
       dispatch('fetchGirderDandiset', { identifier });
       dispatch('fetchOwners', identifier);
 
@@ -45,12 +47,8 @@ export default {
       await dispatch('fetchDandisetVersions', { identifier });
 
       if (dandisetHasVersion(state.versions, version)) {
+        // Version is a valid dandiset version, load that version
         dispatch('fetchPublishDandiset', { identifier, version });
-      } else if (!version) {
-        if (state.versions.length) {
-          const { version: mostRecentVersion } = state.versions[0];
-          dispatch('fetchPublishDandiset', { identifier, version: mostRecentVersion });
-        }
       }
     },
     async fetchDandisetVersions({ state, commit }, { identifier }) {
--- a/web/src/store/dandiset.js
+++ b/web/src/store/dandiset.js
@@ -1,12 +1,13 @@
 import { girderRest, publishRest } from '@/rest';
-import { draftVersion } from '@/utils';
+import { draftVersion, isPublishedVersion } from '@/utils';
 
 export default {
   namespaced: true,
   state: {
     publishDandiset: null,
     girderDandiset: null,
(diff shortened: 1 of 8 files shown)


web/src/views/DandisetLandingView/DandisetLandingView.vue
```diff
@@ -152,30 +152,63 @@ export default {
...
   },
   methods: {
+    navigateToLatestVersion() {
+      // Set default version to most recent if this dandiset has versions
+      // Otherwise set it to draft
+      let version = draftVersion;
+      const { dandisetVersions: versions } = this;
+
+      if (versions && versions.length) {
+        if (dandisetHasVersion(versions, this.version)) { return; }
```
dchiquit, line 186: This makes me think that this method doesn't actually `navigateToLatestVersion`. It seems like it only navigates to latest version if the current version isn't valid; if `this.version` is not falsy, then I would assume this method would generally do nothing. Should it be renamed?

web/src/store/dandiset.js
```diff
@@ -21,16 +22,58 @@ export default {
...
+      await dispatch('fetchDandisetVersions', { identifier });
+
+      if (dandisetHasVersion(state.versions, version)) {
+        dispatch('fetchPublishDandiset', { identifier, version });
+      } else if (!version) {
+        if (state.versions.length) {
+          const { version: mostRecentVersion } = state.versions[0];
+          dispatch('fetchPublishDandiset', { identifier, version: mostRecentVersion });
+        }
+      }
```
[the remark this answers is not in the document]
AlmightyYakob: Good catch, I'll add it.

web/src/views/DandisetLandingView/DandisetLandingView.vue
AlmightyYakob: > if this.version is not falsy, then I would assume this method would generally do nothing

This method only does nothing if `this.version` is contained in `versions`. You're right though, the point of this method isn't to strictly navigate to the latest version, it's to default the version to something known (`draft` or the latest version), if the specified version isn't a known version. I can change the name.

web/src/store/dandiset.js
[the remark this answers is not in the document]
AlmightyYakob: Addressed in 7251558.

I realized that I didn't like the logic I had around setting a default in the store anyway. To me, the `initializeDandisets` action should apply strictly to the `identifier` and `version` passed in. If the version happens to be `draft` or an invalid version, then we simply leave `publishDandiset` as `null`, indicating that the specified version doesn't have a published dandiset. That commit adds that change, as well as handles the defaults in the view itself, allowing for less confusion and duplicated code between the store and view.

web/src/views/DandisetLandingView/DandisetLandingView.vue
AlmightyYakob: Addressed in 7251558.

PR #413, opened 2020-06-26
Add asset download links in PublishFileBrowser.vue

AlmightyYakob:
Depends on backend functionality from https://github.com/dandi/dandi-publish/pull/53.

### DIFF
--- a/web/src/views/FileBrowserView/PublishFileBrowser.vue
+++ b/web/src/views/FileBrowserView/PublishFileBrowser.vue
@@ -63,6 +63,17 @@
                 </template>
               </v-icon>
               {{ item.name }}
+              <v-spacer />
+              <v-list-item-action v-if="itemDownloads[item.name]">
+                <v-btn
+                  icon
+                  :href="itemDownloads[item.name]"
+                >
+                  <v-icon color="primary">
+                    mdi-download
+                  </v-icon>
+                </v-btn>
+              </v-list-item-action>
             </v-list-item>
           </v-list>
         </v-card>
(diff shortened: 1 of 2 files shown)


web/src/views/FileBrowserView/PublishFileBrowser.vue
```diff
@@ -142,8 +154,27 @@ export default {
...
       }
+
+      // Create the download link in itemDownloads for each item in items
+      this.itemDownloads = {};
+      const { identifier, version, location } = this;
+      const baseAssetPath = `${publishRest.defaults.baseURL}/dandisets`;
+
+      items.filter((x) => !x.folder).map(async (item) => {
+        const relativePath = `${location}${item.name}`;
+        const assetPath = `${baseAssetPath}/${identifier}/versions/${version}/assets/`;
```
dchiquit, line 169: Organizationally, I would prefer keeping API URLs segregated in `rest.js` rather than mixing them into components. Publish API endpoints should each have a method on `publishRest` so components don't have to worry about API semantics like this.
I'm willing to negotiate if anyone else has different thoughts.

AlmightyYakob: Good idea, I'll change it, although I don't think it makes sense to try to address that for all endpoints in this PR, so I'll just make one for `assets` if that sounds good.

dchiquit: That's fine, I just wanted to justify my philosophy.

AlmightyYakob: b951680

PR #417, opened 2020-07-03
Add instructions for connecting to dandi-publish

mvandenburgh:
Should be merged alongside https://github.com/dandi/dandi-publish/pull/61

I changed the `docker-compose` ports for the girder and dandiarchive web clients to be consistent with the ports used in the "Developing Locally" instructions so the instructions can be a little simpler. Is there any specific reason why they differed to begin with/reason not to make this change?


brianhelba:
@dchiquit @mvandenburgh I didn't set up any of the Docker Compose functionality here in `dandiarchive`. I have no feelings either way about this PR.

mvandenburgh:
> There's actually another step to setting up `dandiarchive` that you might have missed. Not only does `dandi-publish` need an API key from `dandiarchive`, `dandiarchive` needs an API key from `dandi-publish`.
> 
> 1. log in to the django admin console (http://localhost:8000/admin/, probably) for `dandi-publish` as an admin user
> 2. Go to `Auth Token` > `Tokens` > `Add`
> 3. Create a new auth token and copy it
> 4. Go to the girder API page for `dandiarchive` (probably http://localhost:8080/api/v1)
> 5. Using `System`>`PUT /system/setting`, set `dandi.publish_api_url` to the publish API endpoint (include `/api` on the end) and set `dandi.publish_api_key` to the key you created. You can verify with the `GET` endpoint
> 
> You can see the settings in the code at `girder_dandi_archive/settings.py`.

I believe that is described here -
https://github.com/dandi/dandiarchive/tree/document-env-vars#publish

I also just now changed the wording a little to make it a little clearer. 

Would it be better to include the API auth steps in the `dandiarchive` readme instead of linking to the `dandi-publish` one? I was thinking it would be better to not have the same instructions in two places, but in doing so the `dandiarchive` readme is dependent on the `dandi-publish` readme.

mvandenburgh:
@dchiquit I made a small change and it looks like Github isn't let me merge until it is approved again.

mgrauer:
@mvandenburgh Can you double check this branch after I merged in master, and if it looks good, go ahead and merge.

PR #418, opened 2020-07-06
Remove jest-puppeteer-vuetify

dchiquit:
`jest-puppeteer-vuetify` has been moved to
https://github.com/dchiquit/jest-puppeteer-vuetify, so the local copy of
it has been removed and `test/package.json` now refers to the npm hosted
version.

Running `yarn` apparently updated dependencies as well, so that is also happening in this commit.


brianhelba:
@dchiquit Do you have any desire to preserve the Git history from this repo into https://github.com/dchiquit/jest-puppeteer-vuetify ? It's possible to do so with a little `git filter-branch` magic.

dchiquit:
Nope, I don't think the history is very valuable. There's still several rounds of polishing to do before I'd call it launchable, but it was needed on another project.

Issue #412, opened 2020-06-25
Refactor Metadata Editor


AlmightyYakob:
Some key items:

- Add ability to follow `$refs` in JSON Schema
- Redesign to allow more structured data viewing

AlmightyYakob:
We've discussed addressing this by replacing the existing implementation with https://github.com/koumoul-dev/vuetify-jsonschema-form

waxlamp:
This is done.

PR #423, opened 2020-07-09
Update DLP and general client to handle new schema

### LINKED ISSUE #443: Add note to GUI about metadata editor being disabled
During the schema release, the metadata editor will be broken and will have to be fixed in a later pass, compared to the DLP. So when we land the schema release, disable the metadata editor and add a note explaining how to edit with CLI and when to expect metadata editor functionality to return.

AlmightyYakob:
~~Draft PR as I'm still verifying functionality.~~

The main goal of this PR is to reduce the complexity of `ListingComponent.vue`, and tailor the new components towards rendering the new schema in the desired way.

Regarding the new schema, currently it's stored as a file at `web/src/assets/schema/dandiset.json`. I think ideally we would either fetch this from somewhere at runtime, or bundle it as a submodule (the module being https://github.com/dandi/dandi-core). I'm more in favor of the latter, as having to fetch this data at runtime would incur a noticeable delay on the DLP load time.

This PR doesn't really address the metadata editor itself, and I don't think qualifies as a fix for #412, even though it does address some of the issues.

Fixes #443 (WIP)


web/src/views/DandisetLandingView/DandisetLandingView.vue
```diff
@@ -172,6 +151,14 @@ export default {
       },
     },
   },
+  async created() {
+    const schema = SCHEMA;
+
+    // TEMPORARY: Remove known circular reference
```
dchiquit, line 157: Is this a circular reference in the schema itself, or in some data that is being used? Do all circular references need to be removed like this? Is there a solution that will deprecate this temporary fix?

AlmightyYakob: It's the schema itself. See [this](https://github.com/dandi/dandi-core/pull/4#discussion_r452352431) comment. Yes, all circular references need to be removed, as otherwise `Ajv` can't run validation. However, this appears to be the only circular reference in this schema. I'm not sure of the long-term solution.

AlmightyYakob:
Superseded by #564

PR #429, opened 2020-07-14
Exclude `tests`, not `test`

jwodder:
The name of the tests directory (specifically, the one in the Python project subdirectory) is `tests` with an S, and so the call to `find_packages()` in `setup.py` should exclude `tests`, not `test`.  (Alternatively, as `tests` does not contain an `__init__.py` file, it would also be acceptable to just remove the `exclude=[...]` bit entirely.)


girder-dandi-archive/setup.py
```diff
@@ -31,7 +31,7 @@
     include_package_data=True,
     keywords="girder-plugin, dandi_archive",
     name="girder_dandi_archive",
-    packages=find_packages(exclude=["test", "test.*"]),
+    packages=find_packages(exclude=["tests", "tests.*"]),
```
yarikoptic, line 34: And if it wasn't a web service, I would have advised not to exclude tests from the distribution - they are generally useful to ship along to be able to test  the installed instance

jwodder: I believe it's generally recommended to include tests in sdists (which this project currently does not do), but including tests in wheels shouldn't be done unless the test folder is moved underneath the main package.

yarikoptic: correct!  here situation is a bit more obscure -- `tests` is under the main package. But there is top level `test/` which is with the tests for the JS frontend component
```
(git)lena:~/proj/dandi/dandiarchive[rf-dandisetyaml]git
$> ls -ld ./girder-dandi-archive/tests test
drwx------ 1 yoh yoh  32 May 26 12:05 ./girder-dandi-archive/tests/
drwx------ 1 yoh yoh 218 Jul  8 16:31 test/
1 45890.....................................:Wed 15 Jul 2020 10:22:47 AM EDT:.
(git)lena:~/proj/dandi/dandiarchive[rf-dandisetyaml]git
$> ls -l test
total 264
-rw------- 1 yoh yoh   1617 Jul  8 16:31 README.md
-rw------- 1 yoh yoh    120 May 26 12:05 babel.config.js
-rw------- 1 yoh yoh    230 May 26 12:05 jest-puppeteer.config.js
-rw------- 1 yoh yoh    437 Jul  8 16:31 jest.config.js
-rw------- 1 yoh yoh    938 Jun 15 15:51 jest.setup.js
-rw------- 1 yoh yoh   1055 Jul  8 16:31 package.json
drwx------ 1 yoh yoh     34 May 26 12:05 src/
-rw------- 1 yoh yoh 244429 Jul  8 16:31 yarn.lock
```

PR #439, opened 2020-07-21
Require ownership for lock endpoints

### LINKED ISSUE #422: lock/unlock: scope endpoints within girder
We enable read/write but to lock/unlock needs full set of permissions atm

@mgrauer suggests that should be as easy as adding

```python
@access.user(scope=TokenScope.DATA_WRITE)
```

dchiquit:
The lock and unlock endpoints now require ownership of the dandiset.

IIRC we decided that the only permission level was ownership. Non-owners cannot modify the dandiset, so they should not be able to lock the dandiset.

Fetching the current lock owner is already public.

Fixes #422

### DIFF
--- a/girder-dandi-archive/girder_dandi_archive/rest.py
+++ b/girder-dandi-archive/girder_dandi_archive/rest.py
@@ -324,7 +324,7 @@ class DandiResource(Resource):
             "size": drafts["size"],
         }
 
-    @access.user(scope=TokenScope.DATA_OWN)
+    @access.user
     @autoDescribeRoute(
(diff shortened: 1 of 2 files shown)


mgrauer:
@dchiquit I needed to give the token full scope to properly use these endpoints, I wasn't able to get the user data own token to work and so have backed out your scoping while increasing the scope of the dandicli generated token. Can you take a look for QA?

mgrauer:
>I can't approve for some reason

Your standards are too high. But also, probably it is because you were the author of the PR or the original commit.

PR #441, opened 2020-07-22
Handle trailing slash in publishRest baseURL

### LINKED ISSUE #419: Fix extra slash in UI asset download redirect
The endpoint it directs to is:
```
https://publish.dandiarchive.org/api//dandisets/000003/versions/0.200605.1956/assets/05570abd-3f2b-4cac-8411-2061d866e7f5/download
```

where it should be:
```
https://publish.dandiarchive.org/api/dandisets/000003/versions/0.200605.1956/assets/05570abd-3f2b-4cac-8411-2061d866e7f5/download
```

`/api//` -> `/api/`

AlmightyYakob:
Fixes #419 

This was due to not handling trailing slashes in the `VUE_APP_PUBLISH_API_ROOT` environment variable, an oversight because of a difference between the local variables (in `.env.development`), and the ones being set on deploy. I've updated `.env.development` to match the format of the deploy environment.

### DIFF
--- a/web/src/rest.js
+++ b/web/src/rest.js
@@ -1,8 +1,15 @@
 import axios from 'axios';
 import { RestClient } from '@girder/components/src';
 
-const apiRoot = process.env.VUE_APP_API_ROOT;
-const publishApiRoot = process.env.VUE_APP_PUBLISH_API_ROOT;
+// Ensure doesn't contain trailing slash
+const apiRoot = process.env.VUE_APP_API_ROOT.endsWith('/')
+  ? process.env.VUE_APP_API_ROOT.slice(0, -1)
+  : process.env.VUE_APP_API_ROOT;
+
+// Ensure contains trailing slash
+const publishApiRoot = process.env.VUE_APP_PUBLISH_API_ROOT.endsWith('/')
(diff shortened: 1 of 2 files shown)


web/src/rest.js
```diff
@@ -25,7 +24,8 @@ const publishRest = axios.create({ baseURL: publishApiRoot });
 Object.assign(publishRest, {
   assetDownloadURI(asset) {
     const { uuid, version: { version, dandiset: { identifier } } } = asset;
-    return `${publishRest.defaults.baseURL}/dandisets/${identifier}/versions/${version}/assets/${uuid}/download`;
+    const baseURL = publishRest.defaults.baseURL.endsWith('/') ? publishRest.defaults.baseURL.slice(0, -1) : publishRest.defaults.baseURL;
```
brianhelba, line 27: Isn't the appropriate place for this logic up on line 5, so `publishApiRoot` always has a consistent format?

yarikoptic: There is no function in js world to join urls like urllib.parse.urljoin in python?

AlmightyYakob: Good point, I'll fix it

PR #447, opened 2020-07-24
Add download action in data browser for girder items

AlmightyYakob:
I don't see an associated issue, if I'm wrong @mgrauer feel free to replace this comment with a link to that issue.


mgrauer:
Sorry, there was no issue, but I've created #449 , it was requested by @satra as a UX improvement.

PR #453, opened 2020-07-29
Validate dandiset schema version on folder save

### LINKED ISSUE #451: girder to fail if client tries to set dandiset level metadata to some old schema
In `dandi-cli` following code (re)uploads metadata for the dandiset folder: https://github.com/dandi/dandi-cli/blob/master/dandi/cli/cmd_upload.py#L328 

```python
            # Special handling for dandiset.yaml
            # Yarik hates it but that is life for now. TODO
            #
            if op.basename(path) == dandiset_metadata_file:
                # We need to upload its content as metadata for the entire
                # folder.
                folder_rec = ensure_folder()
                remote_metadata = folder_rec["meta"]
                if remote_metadata.get("dandiset", {}) == dandiset.metadata:
                    yield skip_file("exists (same)")
                else:
                    remote_metadata["dandiset"] = dandiset.metadata
                    yield {"status": "uploading dandiset metadata"}
                    client.addMetadataToFolder(folder_rec["_id"], remote_metadata)
                    yield {"status": "done"}
                # Interrupt -- no file to upload
                return
```

dchiquito:
Closes #451
If this is merged before `schemaVersion` exists on dandisets, it will be impossible to save dandisets. I didn't see that specified on any dandisets currently in production, so this might not be merged for a while.


girder-dandi-archive/girder_dandi_archive/schema_validation.py
```diff
@@ -0,0 +1,21 @@
...
+
+def validate_schema(event):
+    meta = event.info["meta"]
+    if "dandiset" not in meta:
+        # Do not validate if there is no dandiset metadata property
+        # Only top level dandiset folders should be validated
+        return
+    dandiset = meta["dandiset"]
+    if "schemaVersion" not in dandiset:
+        raise ValidationException("No schema version specified on folder")
```
mgrauer, line 14: At first I was thinking we should tell people to upgrade the dandi-cli here, but I think it makes more sense for clients to be able to take whatever action they think is appropriate.

@yarikoptic are these error message sufficient for you?:

>"No schema version specified on folder"

>f"Schema version {schema_version} "
            f"is less than the minimum version {SCHEMA_MINIMUM_VERSION}"

mgrauer:
@AlmightyYakob This LGTM conceptually, can you review the code when you get a chance?

girder-dandi-archive/girder_dandi_archive/schema_validation.py
yarikoptic:
```suggestion
        raise ValidationException("dandiset metadata record is missing required schema version. Hint: outdated client")
```
since "folder" is a girder term here since dandiset is stored in the folder.  Also, since unlikely I should parse / match exception messages, a `Hint` would be appreciated by the users.

yarikoptic: how would `ValidationException` be visible to the client? I mean -- either I could/should parse it out to determine that the action has failed because of `ValidationException` specifically, and then on "what to tell" to the user?  I.e. when should I recommend updating the client if ValidationException has happened. I would prefer not to parse message itself since those could change etc.

yarikoptic:
Thank you @dchiquit !  We indeed should wait a bit until I kick out a client which starts to specify the schemaVersion!

girder-dandi-archive/tests/test_validate_schema.py
```diff
@@ -0,0 +1,40 @@
...
+    folder = Folder().createFolder(drafts_collection, "000001", parentType="collection")
+    with pytest.raises(ValidationException) as exception:
+        folder = Folder().setMetadata(folder, {"dandiset": {"schemaVersion": "0.0.0-rc0"}})
+    assert (
+        str(exception.value)
+        == f"Schema version 0.0.0-rc0 is less than the minimum version {SCHEMA_MINIMUM_VERSION}"
+    )
+
+
+def test_newer_schema_version_does_not_validate(server, drafts_collection):
```
AlmightyYakob, line 38: Shouldn't this be `test_newer_schema_version_validates`?

```suggestion
def test_newer_schema_version_validates(server, drafts_collection):
```

girder-dandi-archive/girder_dandi_archive/schema_validation.py
```diff
@@ -0,0 +1,21 @@
...
+    schema_version = dandiset["schemaVersion"]
+    # TODO this will break if MINIMUM_VERSION ever has 2-digit numbers
+    if schema_version < SCHEMA_MINIMUM_VERSION:
```
yarikoptic, line 17: I believe plain string comparison wouldn't work here since  
```
$> python -c 'print("1.0.0-rc1" < "1.0.0")'
False
```
I use `semantic-version` package in dandi-cli:
```
$> python3 -c 'from semantic_version import Version; print(Version("1.0.0-rc1") < Version("1.0.0"))'
True
```
but please add additional check/dedicated raise to first check if provided version is "legit semver":
```
$> python3 -c 'from semantic_version import Version; print(Version("1.0.0kukaracha") < Version("1.0.0"))'   
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/lib/python3/dist-packages/semantic_version/base.py", line 105, in __init__
    major, minor, patch, prerelease, build = self.parse(version_string, partial)
  File "/usr/lib/python3/dist-packages/semantic_version/base.py", line 303, in parse
    raise ValueError('Invalid version string: %r' % version_string)
ValueError: Invalid version string: '1.0.0kukaracha'

$> python3 -c 'from semantic_version import validate, Version; print(validate("1.0.0kukaracha"))'                     
False
```

girder-dandi-archive/girder_dandi_archive/schema_validation.py
mgrauer: @yarikoptic to clarify, you want us to add in a message saying "Hint: try to upgrade your dandi-cli" as part of the error message?

girder-dandi-archive/girder_dandi_archive/rest.py
```diff
@@ -74,6 +75,7 @@ def create_dandiset(self, params):
             "name": name,
             "description": description,
             "identifier": padded_identifier,
```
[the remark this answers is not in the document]
dchiquit: Yes, but I think it's out of scope for this issue. This is only to validate that schemaVersion is present and valid, we are not actually validating the schema at this time. Whenever we start doing that it will cause all these tests to fail, at which point we should change that.

girder-dandi-archive/girder_dandi_archive/schema_validation.py
```diff
@@ -0,0 +1,26 @@
...
+from girder.exceptions import ValidationException
+
+from .constants import SCHEMA_MINIMUM_VERSION
+
+
+def validate_schema(event):
+    meta = event.info["meta"]
+    if "dandiset" not in meta:
+        # Do not validate if there is no dandiset metadata property
+        # Only top level dandiset folders should be validated
```
yarikoptic, line 12: well -- client must not provide metadata for any other folder!  I would fail here to catch some bug where client tries to do unexpected

girder-dandi-archive/girder_dandi_archive/schema_validation.py
```diff
@@ -0,0 +1,26 @@
...
+        return
+    dandiset = meta["dandiset"]
+    if "schemaVersion" not in dandiset:
+        raise ValidationException(
+            "Missing required metadata field 'schemaVersion'. "
+            "Check that your client is up to date."
+        )
+    schema_version = dandiset["schemaVersion"]
+    if Version(schema_version) < Version(SCHEMA_MINIMUM_VERSION):
```
yarikoptic, line 21: as I have mentioned in the original comment -- `schema_version` could be not a proper semver. Instead of just bailing with a generic exception from semantic_version, IMHO it is better to raise `ValidationException` with a meaningful message

girder-dandi-archive/girder_dandi_archive/schema_validation.py
dchiquit: This method is run whenever any folder is saved by any client. I use the fact that only root level dandiset folders have the `dandiset` metadata property to ensure that only those folders are validated. I wanted to document that fact here.

mgrauer:
@yarikoptic Your request was addressed, could you re-review?

yarikoptic:
looks good, thanks!

girder-dandi-archive/girder_dandi_archive/rest.py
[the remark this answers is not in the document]
AlmightyYakob: You're right, and I forgot that I actually already did that in my other branch https://github.com/dandi/dandiarchive/blob/fe9f0cd953cbac7dd0df865f78021000c7f0fe3e/girder-dandi-archive/girder_dandi_archive/rest.py#L73-L77

PR #454, opened 2020-07-29
Remove e2e tests, temporarily

mgrauer:
E2E tests are temporarily broken. Remove them so that we can merge to master and deploy for the time being.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,49 +40,6 @@ jobs:
           command: yarn run lint --no-fix --max-warnings=0
           working_directory: test
 
-  test-e2e:
-    docker:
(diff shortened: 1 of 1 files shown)


yarikoptic:
if they are to be brought back in, I think commenting them out (instead of removing; and also adding a comment why commented out) would be better so there is a "constant reminder" whenever anyone looks into that file.  But that is just a matter of my personal preference - you might be keeping better notes on what TODO ;)

mgrauer:
I created a tracking issue when I created this PR, see #455. I'd rather have cleaner code by removing unused code and rely on source control to manage the code removal/retrieval--I recognize this might just be a matter of taste.

brianhelba:
Fixed by #456

PR #456, opened 2020-07-30
Persist the activation of virtual environments in CI

brianhelba:
Also, don't create a virtual environment just to install Ansible.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,10 +48,11 @@ jobs:
     steps:
       - checkout
       - run:
-          name: Create a virtualenv for installation
-          command: |
-            python3 -m venv install_env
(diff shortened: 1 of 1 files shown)


brianhelba:
Fixes #455.

mgrauer:
This should also close #454 and #436 when it gets merged.

Can you explain why we need a virtualenv for installing girder-dandi-archive but not for Ansible?

brianhelba:
We want to install Girder in isolation, without any of the clutter / complication from the system Python packages coming with the `circleci/python:3.6` image.

Ansible is a straightforward tool to install, and we don't care about what exact library versions are used, so there's no need to have one more line of configuration code to make a virtualenv tool for it.

PR #475, opened 2020-08-12
Add download dandiset dialog

### LINKED ISSUE #473: Add ability to download a full dandiset from the dandiset landing page
This issue is a refinement from the discussion in #471 

Features:

- [x] Add a new control to the Dandiset landing page toolbar which is called "Download"
- [x] When the Download control is clicked, it opens up a new widget with an interaction depicted in screenshot 1
- [ ] The widget should be of the form depicted in screenshot 2, with two accordion components, initially closed
- [x] Do not include the API key info as shown in screenshot 2, this isn't relevant for downloads
- [x] The download widget should have `Help ?` at the far right of the title bar of the widget, with the ? being an info icon such as currently is depicted in screenshot 2. The `Help ?` should take take the user to https://www.dandiarchive.org/handbook/10_using_dandi/#downloading-from-dandi
- [x] In the download a different version accordion section, the options are (this should be dynamic based on the current Dandiset's identifier, I'm using 000006 as an example here):
1. Download a draft, `dandi download https://dandiarchive.org/dandiset/000006/draft `
2. Download the latest version, `dandi download DANDI:000006`
3. Download a specific version, with a dropdown populated from the list of release for this dandiset, `dandi download DANDI:000006/0.200714.1807`

Screenshot 1

Download full dandiset

Screenshot 2

Screen Shot 2020-08-10 at 1 45 08 PM

dchiquit:
Fixes #473. See #473 for requirements for the component.

Add a `DownloadDialog` component which provides DANDI CLI commands to
download dandisets.

Factor out a `CopyText` component to facilitate copying text
snippets and use it in the `ApiKeyItem` component as well.

Move `ApiKeyItem` into the top level `components` directory, as it is
also used in the `DownloadDialog`.

#473 dictates that download commands for dandiset `000006` work like this:
* drafts: `dandi download https://dandiarchive.org/dandiset/000006/draft`
* latest version: `dandi download DANDI:000006`
* specific version: `dandi download DANDI:000006/0.200714.1807`

@mgrauer @yarikoptic can you confirm?


web/src/views/DandisetLandingView/DownloadDialog.vue
```diff
@@ -0,0 +1,177 @@
...
+            <v-expansion-panel-content>
+              <v-list>
+                <v-list-item>
+                  <div>
+                    Install the Python client (DANDI CLI)
+                    in a Python 3.6+ environment using command:
+                    <kbd>pip install dandi</kbd>
+                  </div>
+                </v-list-item>
+                <ApiKeyItem />
```
AlmightyYakob, line 107: I think we need to have some note here if you're not logged in. Currently, it will just display that text with an empty api key section. I think we some extra information here that informs the user that they need to create an account in order to use the DANDI CLI, and maybe a tool-tip over the `ApiKeyItem` that tells them that they need to be logged in to retrieve their API key.

web/src/components/CopyText.vue
```diff
@@ -0,0 +1,65 @@
+<template>
+  <v-text-field
+    ref="textField"
+    v-model="text"
+    readonly
+    :success-messages="messages"
+    v-bind="$attrs"
```
AlmightyYakob, line 7: Is this used so that arbitrary props can be passed through this component to the `v-text-field` (like `outlined`)?

web/src/views/DandisetLandingView/DownloadDialog.vue
satra: one of the requirements in #473 is that the reference to the API key be removed. users don't need to create an account to use the dandi cli for downloading.

web/src/views/DandisetLandingView/DownloadDialog.vue
```diff
@@ -0,0 +1,177 @@
...
+            icon-hover-text="Copy command to clipboard"
+            dense
+            filled
+            outlined
+          />
+        </v-list-item>
+        <v-expansion-panels>
+          <v-expansion-panel v-if="availableVersions.length > 0">
+            <v-expansion-panel-header>
+              Download a different version
```
mgrauer, line 53: Add a `?` at the end please, `Download a different version?`

web/src/views/DandisetLandingView/DownloadDialog.vue
mgrauer: Yes, we don't want the API key component in download.

web/src/views/DandisetLandingView/DownloadDialog.vue
```diff
@@ -0,0 +1,177 @@
...
+              href="https://www.dandiarchive.org/handbook/10_using_dandi/#downloading-from-dandi"
+              target="_blank"
+              rel="noopener"
+              icon
+            >
+              <v-icon
+                color="primary"
+                v-on="on"
+              >
+                mdi-help-circle
```
mgrauer, line 30: Can you move the help ? to the far right side of the widget with Help text first, so e.g. `Help (?)`

mgrauer:
One more thing, can you change the text they should run to `pip install "dandi>=0.6.0"` so that if people have a lower version they will be sure to get the right version?

web/src/components/CopyText.vue
dchiquit: Correct. I left a comment this at the top of the component script, but you can't really leave comments on HTML templates.

mgrauer:
This is great, I have one last request. Could you take the pip command and put it in a copyable element like you do with the DANDI CLI download commands? Apologies for not bundling this with earlier requests.

dchiquit:
@mgrauer Sadly no, that's not really feasible without putting the command into a text box, which de-emphasizes that it's a code snippet. It's relatively simple to copy from text inputs, but the `pip install dandi` thing is in a `<kbd>`, which is more or less impossible to copy from reliably. I don't think the effort is worth the QOL improvement.

mgrauer:
I see. Is it possibly from a layout perspective to have a line break so that the pip install command is all on one line?

If you can do that somewhat easily, please do, otherwise let's merge this as is.

PR #480, opened 2020-08-13
Add feature toggles to web client

dchiquit:
* Add a new feature toggle, `UNIFIED_API`, which defaults to `false`.
 * A constant is now exposed on every component for every defined
feature toggle. Components can access the new toggle with
`if (this.UNIFIED_API) {...}` or in templates with `v-if="UNIFIED_API"`.
 * Add two functions to the global scope, `enable` and `disable`.
Typing `enable('UNIFIED_API')` into the web console will enable the
feature toggle for the rest of the session, while
`disable('UNIFIED_API')` will disable it.
 * Toggles are listed in `web/src/featureToggle.js`. For development,
they can be temporarily enabled/disabled there.

Also, a minor linting change was applied to `DownloadDialog`.

You can see this in action on the netlify preview. Just open the web console and type `enable('UNIFIED_API')` and a console log should report the successful toggling.

### DIFF
--- a/web/src/featureToggle.js
+++ b/web/src/featureToggle.js
@@ -1,0 +1,53 @@
+/* eslint no-console: 0 */
+/* eslint no-throw-literal: 0 */
+import Vue from 'vue';
+
+// Define all feature toggles here
+// They can also be enabled/disabled here for development
+const TOGGLES = {
+  UNIFIED_API: false,
+};
+
+// Create a component to hold the feature toggles reactively
+// This object can be mutated at runtime to enable/disable feature toggles
+const featureToggles = new Vue({
+  data() {
+    return TOGGLES;
+  },
+});
+
+// Generate the computed prop to use in the mixin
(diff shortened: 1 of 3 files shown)


AlmightyYakob:
Would we want this feature available to the client in production? What's the main/immediate use case for this?

dchiquit:
I'm working on adding the ability to log into django instead of logging into girder. When you `enable('UNIFIED_API')`, the login form will switch from using the GWC auth to a different component that is aimed at the django auth endpoints (which I am also working on adding). In development you can also achieve that by editing `featureToggle.js`, but for branch previews, it's useful for anyone to be able to see what the changes look like.

This feature would be available in production, but it's completely undocumented and hard to find. I think it's more useful for us to use toggles in production than it is scary to have end users somehow figure them out and see our toggle-protected code.

As an aside, I figured out you can make a bookmark with the content `javascript:enable('UNIFIED_API')` to switch it on with the push of a button.

PR #484, opened 2020-08-14
Add netlify.toml file

mgrauer:
Moves the netlify configuration into a project file, to allow people without access to the netlify account to tweak it. Also allows people forking the repo to set up their own netlify which has the same configuration.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,10 +1,9 @@
 [build]
-  base = "web"
-  publish = "dist"
+base = "web"
+publish = "dist"
+command = "yarn run build"
 
-  command = "yarn run build"
-
-    [build.environment]
-    VUE_APP_API_ROOT = "https://girder.dandiarchive.org/api/v1"
-    VUE_APP_PUBLISH_API_ROOT = "https://publish.dandiarchive.org/api/"
-    VUE_APP_SENTRY_DSN = "https://425b9a012300493d867e97785fae7b88@o308436.ingest.sentry.io/5196549"
+  [build.environment]
+  VUE_APP_API_ROOT = "https://girder.dandiarchive.org/api/v1"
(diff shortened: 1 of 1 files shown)


mgrauer:
Looks like it worked, and the build logs indicate it built from the config file.

netlify.toml
```diff
@@ -0,0 +1,10 @@
+[build]
+  base = "web"
+  publish = "dist"
+
+  command = "yarn run build"
+
+    [build.environment]
```
yarikoptic, line 7: I am curious: Is such indentation required or some convention I haven't encountered yet?

netlify.toml
```diff
@@ -0,0 +1,10 @@
...
+    VUE_APP_API_ROOT = "https://girder.dandiarchive.org/api/v1"
+    VUE_APP_PUBLISH_API_ROOT = "https://publish.dandiarchive.org/api/"
+    VUE_APP_SENTRY_DSN = "https://425b9a012300493d867e97785fae7b88@o308436.ingest.sentry.io/5196549"
```
AlmightyYakob, line 10: Indentation suggestion

```suggestion
[build]
base = "web"
publish = "dist"
command = "yarn run build"

  [build.environment]
  VUE_APP_API_ROOT = "https://girder.dandiarchive.org/api/v1"
  VUE_APP_PUBLISH_API_ROOT = "https://publish.dandiarchive.org/api/"
  VUE_APP_SENTRY_DSN = "https://425b9a012300493d867e97785fae7b88@o308436.ingest.sentry.io/5196549"
```

netlify.toml
AlmightyYakob: In TOML, indentation is optional. You can see this in my other comment, but if a section is a subsection of another, I tend to indent it. Otherwise, I don't indent.

mgrauer: @yarikoptic some of my indentation was accidental, so don't read too much into it :-)

PR #493, opened 2020-08-31
Fix format issues from new version of Black

brianhelba:


brianhelba:
Force merging, since this fixes a broken `master`.

Issue #470, opened 2020-08-10
UI improvement for permalink


mgrauer:
This issue is to gather feedback. Here's a suggested UI improvement on the Dandiset Landing Page to allow people to copy the permalink, and also make it more clear what the permalink is.

![Permalink](https://user-images.githubusercontent.com/595023/89812918-d8b53200-db0e-11ea-8212-8fe4569a09fd.png)

An idea for a future iteration is to be able to share the permalink via a direct email, but this would require more discussion as the scope is larger than a UI change. Just putting it here as food for thought.

![Permalink - Next iteration - add share via email button](https://user-images.githubusercontent.com/595023/89812906-d3f07e00-db0e-11ea-80c6-e584b97121c5.png)

mgrauer:
@bendichter @satra @yarikoptic

satra:
@mgrauer - sorry had missed this. 

a few thoughts. we are going to get multiple identifiers and we should prioritize:

1. doi (we have this now through datacite, and should be able to test it right after schema migration is done)
2. identifiers.org
3. dandiarchive

the first two for published versions, the last one for all versions.

i would also couple this with:
1. a UI update on creating a citation for DANDI.
2. post on twitter link
3. a discuss on neurostars link

i would still like the schema updates and any metadata editor pieces to take current priority, then an actual publish, but as soon as those are done, we can make these changes. in the meantime if someone had the bandwidth and wanted to mock UIs or PRs to handle these different identifiers and external links that would be great.

mgrauer:
Yes, I have separate parallel bandwidth for doing mock UIs than for doing UI dev, so I wanted to get these UI requirements and mockups clarified so that they can be ready to go for a developer when cycles are freed up. I'd like to ideally work through one or two UI improvements per week with the UI/UX team. We can start with these `UX` labelled issues, and then when the desires are clarified, I can close this one out and create a new `webui` issue that is tightly scoped.

The first section you mention is identifiers, and sounds like that should go where current permalink is (would include the current permalink plus other identifiers), and have something like the dropdown I show above in the UI, with the ability to copy out different identifiers. Do you agree or am I missing something?

Then we'd like a create a citation section (which is just a string in a standard format, with e.g. dandiset name, contributor name, etc) ?

Third we'd like some external links like twitter and neurostars. Can you explain more the UX for these?

satra:
[...]

PR #497, opened 2020-09-08
Permalink dropdown

dchiquit:


mgrauer:
Sorry @dchiquit , there were some larger changes this should be rolled in with, based on #470.

I will say this is a strict improvement and should go in, so I'll approve, but let's have a sync before you do any further UI work.

mgrauer (approved):

PR #499, opened 2020-09-11
Reorganize rest utils and debug ability to login/logout of django

dchiquit:
Things that changed:
* The `dandi.publish_api_url` setting no longer requires a trailing `/api`. This was confusing in the past and is ultimately unhelpful. Whenever this is shipped that setting will need to be changed on the production girder.
* Instead of a single `rest.js` that holds all the clients, there is now a `rest/index.js` that imports clients from `rest/girder.js` and `rest/publish.js`.
* There is now a hack for "logging in" to django.
   * Generate an auth token in Django for the user you would like to log in with. This can be done from the admin console.
   * Enable the UNIFIED_API toggle in the console with `enable('UNIFIED_API')`. 
   * Run `setTokenHack('the_auth_token_you_generated`)` in the console.
   * Any requests to Django should now include the `Authorization: Token the_auth_token_you_generated` header.

This is resuscitated from the corpse of https://github.com/dandi/dandiarchive/pull/483/files. I should have fixed all the review comments on that PR.


web/src/rest/index.js
```diff
@@ -0,0 +1,13 @@
+import girderRest from '@/rest/girder';
+import publishRest from '@/rest/publish';
+import toggles from '@/featureToggle';
+
+const loggedIn = () => ((toggles.UNIFIED_API) ? !!publishRest.user : !!girderRest.user);
+const user = () => ((toggles.UNIFIED_API) ? publishRest.user : girderRest.user);
```
AlmightyYakob, line 6: Not something you changed in this PR, but I just realized this could be made a tiny bit simpler:
```suggestion
const user = () => ((toggles.UNIFIED_API) ? publishRest.user : girderRest.user);
const loggedIn = () => !!user();
```

web/src/rest/publish.js
```diff
@@ -0,0 +1,134 @@
...
+      return `${publishApiRoot}api/dandisets/${identifier}/versions/${version}/assets/${uuid}/download`;
+    },
+  },
+});
+
+// This has to be done with an interceptor because
+// the value of publishRest.token changes over time.
+// Using client.defaults.headers.common.Authorization = ...
+// would not update when the token does.
+client.interceptors.request.use((config) => {
```
AlmightyYakob, line 110: FWIW the way I've done this on [multinet](https://github.com/multinet-app/multinetjs/blob/daa20af33cd637272dcf835c64b9ad4dfe72cc87/src/index.ts#L117-L123) is by creating functions that set/remove these defaults, and use them in the login/logout methods. I'm okay leaving this as is, just thought I'd mention it, as I prefer not to use interceptors unless they're absolutely necessary.

web/src/rest/publish.js
```diff
@@ -0,0 +1,134 @@
...
+    headers: {
+      Authorization: `Token ${publishRest.token}`,
+      ...config.headers,
+    },
+  };
+});
+
+export default publishRest;
+
+// This is a hack to allow username/password logins to django
```
AlmightyYakob, line 125: A `TODO` or `FIXME` would be good here, so that it's expressed that this is temporary. Perhaps an issue to track this as well.

web/src/rest/publish.js
```diff
@@ -0,0 +1,134 @@
...
+    return {
+      client,
+      token: null,
+      user: null,
+    };
+  },
+  methods: {
+    // TODO proper OAuth login
+    async logout() {
+      // TODO proper session logout
```
AlmightyYakob, line 39: By this do you mean server side logout of the session as well?

web/src/components/AppBar/UserMenu.vue
```diff
@@ -64,7 +63,13 @@ export default {
...
   },
   methods: {
-    ...mapActions('girder', ['logout']),
+    async logout() {
+      if (toggles.UNIFIED_API) {
+        await publishRest.logout();
+      } else {
+        await girderRest.logout();
+      }
+    },
```
AlmightyYakob, line 72: Does it makes sense to have a centralized place for this logout logic? Maybe in `rest/index.js`?

AlmightyYakob (changes requested):
Looks good, just left a few comments/suggestions. Tested this myself successfully with the netlify preview and publish app :+1:

web/src/rest/publish.js
dchiquit: I liked having `token` being a reactive property of `publishRest`, which would be pretty much annihilated by having to update the client properties directly every time token changes. I agree that this interceptor code is an eye sore, but I think it makes the rest of the code substantially less complex.

web/src/rest/publish.js
dchiquit: This might turn out to be semi-permanent, since OAuth doesn't really support local logins. I anticipate that we will be using this hack a lot to locally test features which require being logged in, which is most of them. If devs are using it frequently to log in, we'll always be aware of it if we need to change it.

It's not a security risk because a django auth token is required to use it, at which point you're already compromised.

web/src/rest/publish.js
dchiquit: Yes, logout should invalidate the session token. That endpoint doesn't exist yet though.

web/src/components/AppBar/UserMenu.vue
dchiquit: Maybe. This stuff is going to be rewritten shortly with OAuth though, so I think it's fine to deal with it then.

AlmightyYakob (approved):
My concerns were addressed, looks good to me :+1:

PR #500, opened 2020-09-17
Migrate stats: take 2

dchiquit:
Things have changed since I put in the first migrate stats PR. Rather than try to recover it, I took the opportunity to clean some stuff up instead.
* Remove the stats vuex module. I was very excited to try vuex when I wrote it, but it is entirely unnecessary. Stats aren't required anywhere else and it makes juggling girder/publish harder.
* Display publish stats when UNIFIED_API is toggled on. 3 of the 6 original stats aren't available in publish, so some rendering is tweaked to look good with the 3 remaining stats.

After toggling on UNIFIED_API, you will need to navigate somewhere else, then back to the home page to trigger a refresh of the stats.


dchiquit:
This review is probably worth salvaging, though it does need some patching up.

waxlamp (approved):

PR #505, opened 2020-11-09
Temporarily disable publish API

dchiquit:


mgrauer (approved):
Works for me without errors in the preview.

PR #507, opened 2020-11-24
Rename UNIFIED_API to DJANGO_API

dchiquit:
Unification is no longer the objective, but the toggle still hides useful functionality and more functionality needs to be added, so renaming it is the easiest course of action.

### DIFF
--- a/web/src/components/AppBar/UserMenu.vue
+++ b/web/src/components/AppBar/UserMenu.vue
@@ -64,7 +64,7 @@ export default {
   },
   methods: {
     async logout() {
-      if (toggles.UNIFIED_API) {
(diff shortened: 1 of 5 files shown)


yarikoptic:
I would prefer to avoid implementation specific detail in the names.  Why not to have it `DANDI_API`?

AlmightyYakob:
> I would prefer to avoid implementation specific detail in the names. Why not to have it `DANDI_API`?

I think this is just a temporary measure, until Girder is no longer present at all. Once that's the case, this toggle will disappear entirely.

AlmightyYakob (approved):

PR #512, opened 2020-11-30
Never show django API dandisets without toggle

### LINKED ISSUE #509: 00003 is gone from listing (?) and there is a strange unidentified dataset on top
image

dchiquit:
Fixes #509


mgrauer (approved):
I'll approve this for now because it is a strict improvement. Notice though that dandiset 3 still shows two draft versions. Should that be fixed in this PR also, or is it a separate issue?

https://deploy-preview-512--gui-dandiarchive-org.netlify.app/#/dandiset/000003/draft

mgrauer (approved):
Thanks for the additional fix.

PR #514, opened 2020-11-30
Django list dandisets

dchiquit:
To test:
* Open [netlify preview](https://deploy-preview-514--gui-dandiarchive-org.netlify.app/)
* Pull up the console and run `enable('DJANGO_API');`
* Visit the public dandisets page (navigate away and back again if you are already there). When it loads it should pull data from api.dandiarchive.org instead of girder.dandiarchive.org.

### DIFF
--- a/web/src/components/DandisetList.vue
+++ b/web/src/components/DandisetList.vue
@@ -23,7 +23,7 @@
             </v-list-item-title>
             <v-list-item-subtitle>
               <v-chip
-                v-if="item.version"
+                v-if="item.version && item.version !== 'draft'"
                 small
                 color="light-blue lighten-4"
                 text-color="light-blue darken-3"
@@ -74,6 +74,7 @@ import moment from 'moment';
 import filesize from 'filesize';
 
 import { getDandisetContact } from '@/utils';
+import toggles from '@/featureToggle';
 import { girderRest } from '@/rest';
 
 export default {
@@ -95,6 +96,9 @@ export default {
   },
   asyncComputed: {
     async dandisetStats() {
+      if (toggles.DJANGO_API) {
+        return this.dandisets;
+      }
       const { items } = this;
       return Promise.all(items.map(async (item) => {
         const { identifier } = item.meta.dandiset;
--- a/web/src/rest/publish.js
+++ b/web/src/rest/publish.js
@@ -69,9 +69,9 @@ const publishRest = new Vue({
       });
       return data;
     },
-    async versions(identifier) {
+    async versions(identifier, params) {
       try {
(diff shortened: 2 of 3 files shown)


AlmightyYakob:
Nice! The dandiset listing seems to work great. While this feature flag is enable, loading the DLP fails. Is this expected and/or is it okay for merge? Since it's hidden behind a feature flag, I'm assuming this isn't supposed to work entirely.

dchiquit:
@AlmightyYakob It is not supposed to work yet, but #516 will resolve that. This review is strictly for the dandiset list page.

web/src/components/DandisetList.vue
```diff
@@ -95,6 +96,9 @@ export default {
   },
   asyncComputed: {
     async dandisetStats() {
+      if (toggles.DJANGO_API) {
+        return this.dandisets;
+      }
```
AlmightyYakob, line 101: What stats does this return? Is it just size and asset count right now?

web/src/rest/publish.js
```diff
@@ -89,16 +95,17 @@ const publishRest = new Vue({
...
         return null;
       }
-      const { count, results } = versions;
-      if (count === 0) {
-        return null;
-      }
-      const { version } = results[0];
-      return this.specificVersion(identifier, version);
+      // Look up the last version using page filters
+      const version = (await this.versions(identifier, { page: count, page_size: 1 })).results[0];
```
AlmightyYakob, line 104: This wouldn't always apply (e.g. if there's many versions) but couldn't you do the following?

```
const versions = await this.versions(identifier);
if (!versions.count) { return null; }
if (!versions.next) {
  const sortFunc = (a, b) => (a.version > b.version ? -1 : 1);
  const latestVersion = versions.results.filter(
    (version) => version.version !== 'draft'
  ).sort(sortFunc)[0];
  return girderize(latestVersion);
}

// Default to existing implementation
```

I lightly tested this in my browser and it performed as expected.

dchiquit: I did it this way specifically for the case when there are more versions than fit in a single page. My solution is the simplest one I could think of that always works, at the cost of some extra requests. Doing it your way would produce some confusing behavior at some point in the future when someone publishes one too many times.

AlmightyYakob: If a dandiset has more versions than fit on a page, the `next` field would not be `null`, and this block wouldn't run, correct?

I'm fine with not implementing this for now, just thought I'd offer a way to possibly save on requests.

AlmightyYakob: We decided to leave as is

AlmightyYakob (approved):

PR #520, opened 2020-12-04
Add TypeScript and Composition API

### LINKED ISSUE #515: Introduce TypeScript and Composition API
The only requirement for this is to update the client configuration to support this. No file conversion is required initially.

AlmightyYakob:
Fixes #515

This also includes an example conversion to Composition API, `DandisetList.vue`. This does a decent job of illustrating the differences between the old SFC style and Composition API style.

### DIFF
--- a/web/src/components/DandisetList.vue
+++ b/web/src/components/DandisetList.vue
@@ -69,7 +69,10 @@
   </v-list>
 </template>
 
-<script>
+<script lang="ts">
+import {
+  defineComponent, ref, computed, watch, PropType,
+} from '@vue/composition-api';
 import moment from 'moment';
 import filesize from 'filesize';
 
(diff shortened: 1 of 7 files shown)


web/src/components/DandisetList.vue
```diff
@@ -69,50 +69,79 @@
...
-  },
-  methods: {
-    filesize,
-    getDandisetContact,
-    formatDate(date) {
+
+      dandisetStats.value = res;
+    }
+
+    // Fetching dandiset stats must be done this way since we don't have access to asyncComputed
```
dchiquit, line 128: This could be rewritten with [watchEffect](https://composition-api.vuejs.org/api.html#watcheffect), it's basically asyncComputed for composition API.

dchiquit (approved):

web/src/components/DandisetList.vue
AlmightyYakob: You mean change the call from `watch` to `watchEffect`? I suppose I could, but since I prefer to use `watch` when possible, since I'm only watching a single dependency.

AlmightyYakob: I think you make a good observation, and it seems `watchEffect` is more suited for future changes (i.e. adding more properties to watch). However, I don't want to hold up this PR any longer. This would be an easy change in the future anyway.

waxlamp: @AlmightyYakob, are you saying you agree with @dchiquit's recommendation here (even if you didn't include it in this PR)? In particular, I'm wondering why you prefer `watch()` just because it's a single dependency you're watching? Why not just make `fetchDandisets()` itself "reactive" by embedding it in a `watchEffect()`?

AlmightyYakob: The main reason I've preferred `watch` over `watchEffect` is that when it comes to actually needing to watch something for changes (instead of using it as a way to achieve async computed props), I prefer to have it clearly defined what's being watched. However, it seems like `watchEffect` is better suited for this specific task, so in the future I'll probably go to `watchEffect` first, and then `watch` if a particular task requires it. There are different applications that are better suited for each one. More docs about this can be found [here](https://composition-api.vuejs.org/api.html#watch).

waxlamp: Thanks @AlmightyYakob. I've read those docs before and they don't really provide concrete guidance on when to use what. Your intuition seems good in general, but I think I like `watchEffect()` specifically for cases where you're trying to emulate a reactive response involving imperative actions, or otherwise trying to bolt on imperative stuff to Vue's reactive system. Things like asyncComputed, or triggering D3 rendering, qualify IMO.

Thanks for the discussion. This is useful for developing mental models for these newer features of Vue.

PR #523, opened 2020-12-07
Comment out girder nginx role

mgrauer:
To work around a bug in Girder, we have manually set the client_max_body_size
on nginx to 400M. By commenting out the girder nginx role from ansible, we
prevent nginx from reprovisioning when ansible is run, as reprovisioning nginx
would overwrite the new setting of 400M.

A part of the workaround for #517

### DIFF
--- a/ansible/playbook.yml
+++ b/ansible/playbook.yml
@@ -9,10 +9,10 @@
       vars:
         mongodb_data_path: /var/lib/mongodb
     - role: girder.girder
-    - role: girder.nginx
-      vars:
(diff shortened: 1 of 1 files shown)


mgrauer:
@brianhelba PTAL

brianhelba (approved):

mgrauer:
I also needed to manually change .girder/girder.cfg to add the line

```
server.max_request_body_size = 419430400
```

to the global block. This tells cherrypy to allow request body data up to 400M.

PR #524, opened 2020-12-09
Django manage ownership

dchiquit:
Fix the ownership chips and the Manage Ownership dialog when the DJANGO_API toggle is on.
To test:

- Open netlify preview
- Pull up the console and run enable('DJANGO_API');
- Visit the public dandisets page. It should render a list of Django dandisets.
- Click on a dandiset you own (according to Django).
- You should see the owners in the little blue chips.
- Run this in the console:
```
setTokenHack('put-your-django-API-key-here')
```
You can grab your current API key by hitting GET /auth/token/ on the swagger page.
- Click `Manage` and verify that the Manage Ownership dialog works as intended.

There is some slightly funky behavior where the initial list of owners in the Manage Ownership dialog does not include first/last names. Since it doesn't impact functionality significantly, I think it's fine to save that for later reengineering.

I added everyone I could think of to https://deploy-preview-524--gui-dandiarchive-org.netlify.app/#/dandiset/000001 so you should be able to test the dialog from there.

### DIFF
--- a/web/src/views/DandisetLandingView/DandisetDetails.vue
+++ b/web/src/views/DandisetLandingView/DandisetDetails.vue
@@ -173,12 +173,12 @@
         <v-col cols="12">
           <v-chip
             v-for="owner in limitedOwners"
-            :key="owner.id"
+            :key="owner.id /* TODO remove this */ || owner.username"
             color="light-blue lighten-4"
             text-color="light-blue darken-3"
             class="font-weight-medium ma-1"
           >
-            {{ owner.login }}
+            {{ owner.login /* TODO remove this */ || owner.username }}
           </v-chip>
           <span
             v-if="numExtraOwners"
--- a/web/src/views/DandisetLandingView/DandisetOwnersDialog.vue
+++ b/web/src/views/DandisetLandingView/DandisetOwnersDialog.vue
@@ -39,7 +39,7 @@
           <template v-for="(owner, i) in newOwners">
             <v-list-item :key="owner._id || owner.id">
               <v-list-item-title>
-                {{ owner.name }} ({{ owner.login }})
+                {{ owner.result }}
               </v-list-item-title>
               <v-list-item-action>
                 <v-btn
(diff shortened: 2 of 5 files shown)


web/src/views/DandisetLandingView/DandisetDetails.vue
```diff
@@ -173,12 +173,12 @@
...
           <v-chip
             v-for="owner in limitedOwners"
-            :key="owner.id"
+            :key="owner.id /* TODO remove this */ || owner.username"
             color="light-blue lighten-4"
             text-color="light-blue darken-3"
             class="font-weight-medium ma-1"
           >
-            {{ owner.login }}
+            {{ owner.login /* TODO remove this */ || owner.username }}
```
AlmightyYakob, line 181: Is this currently to deal with both girder and the new API at the same time?

web/src/views/DandisetLandingView/DandisetOwnersDialog.vue
```diff
@@ -94,7 +91,16 @@ const userFormatConversion = (users) => users.map(({
...
 
-const addResult = (users) => users.map((u) => ({ ...u, result: `${u.name} (${u.login})` }));
+// eslint-disable-next-line camelcase
+const girderize = (users) => users.map(({ username, first_name, last_name }) => ({
+  id: username,
+  login: username,
+  username,
+  // eslint-disable-next-line camelcase
+  name: (first_name && last_name) ? `${first_name} ${last_name}` : null,
+}));
```
AlmightyYakob, line 101: Just an FYI you don't need to ignore the eslint rule here, but it's up to you.

```suggestion
const girderize = (users) => users.map(
  ({ username, first_name: firstName, last_name: lastName }) => ({
    id: username,
    login: username,
    username,
    name: (firstName && lastName) ? `${firstName} ${lastName}` : null,
  }),
);
```

web/src/views/DandisetLandingView/DandisetOwnersDialog.vue
```diff
@@ -94,7 +91,16 @@ const userFormatConversion = (users) => users.map(({
...
+
+const addResult = (users) => users.map((u) => ({ ...u, result: (u.name) ? `${u.name} (${u.login})` : u.login }));
```
AlmightyYakob, line 103: A comment explaining why this is neccesary might be helpful. You might also consider renaming this to `injectResult`, since `addResult` makes it seem like you're adding an element to the list, but it's a very minor thing.

AlmightyYakob (approved):
Tested successfully. I left a few comments but those are all minor non-blocking things.

web/src/views/DandisetLandingView/DandisetDetails.vue
dchiquit: Yep. The Girder user object has a `login` property while the Django user object has a `username`. It was easier to just render like this than to do a proper girderization of the object.

PR #533, opened 2020-12-17
Django oauth login

dchiquit:


web/src/rest/publish.js
```diff
@@ -124,32 +137,22 @@ const publishRest = new Vue({
...
-// This has to be done with an interceptor because
-// the value of publishRest.token changes over time.
+// This is done with an interceptor because the value of
+// oauthClient.authHeaders is initialized asynchronously,
+// and doesn't exist at all if the user isn't logged in.
 // Using client.defaults.headers.common.Authorization = ...
 // would not update when the token does.
 client.interceptors.request.use((config) => {
-  if (!publishRest.token) {
+  if (!oauthClient.authHeaders) {
```
brianhelba, line 146: There's no need for a conditional. If the `oauthClient` is not logged in, then `...oauthClient.authHeaders` will typically be an empty object, but that's technically an implementation detail. In other words, it's expected that downstreams treat the actual content of `oauthClient.authHeaders` as somewhat opaque, unconditionally including it in all API requests.

Also, the boolean value of an empty `Object` in JavaScript is truthy anyway, so this will never pass.

web/src/rest/publish.js
```diff
@@ -40,9 +44,18 @@ const publishRest = new Vue({
     };
   },
   methods: {
-    // TODO proper OAuth login
+    async restoreLogin() {
+      await oauthClient.maybeRestoreLogin();
+      if (oauthClient.isLoggedIn) {
+        this.token = oauthClient.token;
```
brianhelba, line 50: `.token` is not a public property and shouldn't be accessed anywhere by downstream code. Is there a particular reason you want to access it?

brianhelba (changes requested):

web/src/rest/publish.js
dchiquit: A misguided attempt to keep the review footprint down.

web/src/rest/publish.js
```diff
@@ -124,32 +134,19 @@ const publishRest = new Vue({
...
+// would not update when the headers do.
 client.interceptors.request.use((config) => {
-  if (!publishRest.token) {
-    return config;
-  }
   return {
     ...config,
     headers: {
-      Authorization: `Token ${publishRest.token}`,
+      ...oauthClient.authHeaders,
```
brianhelba, line 146: If this runs only once, it should be fine. If this runs a second time (particularly on logout), then any existing auth token header will be carried forward and not actually removed. I think the library could start returning a null value in the header, but the order here of the spread values will have to switch.

dchiquit: This is an [axios interceptor](https://github.com/axios/axios#interceptors) that injects the `oauthClient.authHeaders` individually into every request. I think this pattern is preferable to resetting `client.defaults.headers.common` every time `oauthClient.authHeaders` might have changed.

web/package.json
```diff
@@ -9,6 +9,7 @@
   },
   "dependencies": {
     "@girder/components": "^2.2.5",
+    "@girder/oauth-client": "^0.6.3",
```
brianhelba, line 12: Can you bump this version? There are some important bugfixes in the latest one.

brianhelba (approved):
Code looks good now. Just bump the `@girder/oauth-client` version please.

PR #536, opened 2020-12-30
Django API keys

dchiquit:
This makes the API key in the user menu work with the DJANGO_API toggle. It will automatically create a new API key if one does not exist.

### DIFF
--- a/web/src/rest/publish.js
+++ b/web/src/rest/publish.js
@@ -62,7 +62,7 @@ const publishRest = new Vue({
         return data;
       } catch (e) {
         // If the request returned 404, the user doesn't have an API key yet
-        if (e.message === 'Request failed with status code 404') {
+        if (e.response.status === 404) {
           // Create a new API key
           const { data } = await client.post('auth/token/');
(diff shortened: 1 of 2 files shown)


web/src/rest/publish.js
```diff
@@ -56,6 +56,20 @@ const publishRest = new Vue({
       await oauthClient.logout();
       this.user = null;
     },
+    async apiKey() {
+      try {
+        const { data } = await client.get('auth/token/');
+        return data;
+      } catch (e) {
+        // If the request returned 404, the user doesn't have an API key yet
+        if (e.message === 'Request failed with status code 404') {
```
AlmightyYakob, line 65: Wouldn't it be a bit better to directly check the response status? I believe it'd be something like:

```suggestion
        if (e.response.status === 404) {
```

However I'm not certain.

AlmightyYakob (changes requested):
Looks good overall, just one suggestion

AlmightyYakob (approved):
:shipit:

PR #537, opened 2020-12-30
DJANGO_API works with sorting dandisets

dchiquit:
Relies on https://github.com/dandi/dandi-api/pull/45


AlmightyYakob:
The way this is currently implemented, it takes quite a while after adjusting the sort for the changes to take effect. I believe this is because up to 17(!) requests are being made when the sort is changed. 17 = 8*2 + 1. It seems to make an initial request to retrieve the relevant paginated dandisets (the 1), and then makes an unpaginated request for each dandiset's versions (8), and then makes a paginated request for each dandiset's versions (8).

Could you shed some light as to why this is happening?

dchiquit:
To be clear, it's making 1 request to list all dandisets (paginated), 8 requests to get all the versions for those dandisets (paginated), filtering that to get the most recent version for each dandiset, then 8 requests to get details about the most recent versions.

This is the result of having an exact API spec. When implementing the first draft of the dandiset list page with Django I opted to just go overboard with the requests rather than make a mess across multiple repositories. It was simpler at the time, and we have plans to refactor anyway. For now it just creates UX pain, which is a compelling reason to refactor later.

My vague plan was to eventually add an extra field to the Dandiset serializer called `most_recent` or something so when you list all dandisets you also get the details of the most recent version of that dandiset, which is generally what you really wanted anyway. That should cut the requests required down to 1.

AlmightyYakob (approved):

PR #538, opened 2020-12-31
Convert publishRest to TS

dchiquit:
I'm skipping converting girder.js because we're removing it soon, it is very small, and all it does is import a girder module that does not have types.


web/src/rest/publish.ts
```diff
@@ -128,24 +128,24 @@ const publishRest = new Vue({
...
+    async createDandiset(name: string, description: string) {
       const metadata = { name, description };
       return client.post('dandisets/', { name, metadata });
     },
-    async owners(identifier) {
+    async owners(identifier: string) {
       return client.get(`dandisets/${identifier}/users/`);
     },
-    async setOwners(identifier, owners) {
+    async setOwners(identifier: string, owners: [any]) {
```
AlmightyYakob, line 141: Is `owners` supposed to be an array? In that case, the proper syntax is:

```suggestion
    async setOwners(identifier: string, owners: any[]) {
```

What you've done is type `owners` as a [tuple](https://www.typescriptlang.org/docs/handbook/basic-types.html#tuple) of length 1, which I'm guessing is not what you intended. Also, if we add the `User` type as I mentioned above, this could become:
```suggestion
    async setOwners(identifier: string, owners: User[]) {
```

Unless an `Owner` here doesn't conform to the same spec as `User`.

web/src/rest/publish.ts
```diff
@@ -128,24 +128,24 @@ const publishRest = new Vue({
...
       return client.put(`dandisets/${identifier}/users/`, owners);
     },
-    async searchUsers(username) {
+    async searchUsers(username: string) {
       const { data } = await client.get('users/search/?', { params: { username } });
       return data;
     },
-    assetDownloadURI(asset) {
+    assetDownloadURI(asset: any) {
```
AlmightyYakob, line 148: Again, this might be a good place to define the `Asset` type, along with the others. Since these type are pretty universal, they could probably go in `web/@types/` or something similar, and be imported from there.

AlmightyYakob (changes requested):
I left a few suggestions, mainly because I think this it's the right time to to concretely define some of the types we're using for the API, as well as internal use.

web/src/rest/publish.ts
AlmightyYakob: For the record I both mistyped and chose a wrong name for the types folder, as @dchiquit pointed out offline. I believe this will end up being `/web/src/types/`.

web/src/types/index.ts
```diff
@@ -0,0 +1,40 @@
...
+  modified: string,
+  version: Version,
+}
+
+export interface Paginated<T> {
+  count: number,
+  next: string,
+  previous: string,
+  results: T[],
+}
```
AlmightyYakob, line 40: Very nice use of generics! :+1:

AlmightyYakob (commented):
Just one question otherwise looks good

web/src/rest/publish.ts
```diff
@@ -56,7 +59,11 @@ const publishRest = new Vue({
       await oauthClient.logout();
       this.user = null;
     },
-    async apiKey() {
+    async me(): Promise<User> {
+      const { data } = await client.get('users/me/');
```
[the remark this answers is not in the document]
dchiquit: I added this suggestion, but now that I'm thinking more deeply about it I don't think I agree in general. This is the only place where this method is called, everywhere else should just use `publishRest.user` to access the current user object, and that is already typed `User | null`. try/catching will basically just mask any unexpected API errors that might happen and just set `user` to `null`, which would produce strange behaviors that would be harder to diagnose.

AlmightyYakob (approved):

PR #539, opened 2020-12-31
Ts feature toggles

dchiquit:
Convert featureToggles.js to TS.
Also create a vetur.config.js (https://vuejs.github.io/vetur/guide/setup.html#advanced) to register `./web` as the actual repo root. Vetur recently added the option to configure multiple projects in a single monorepo, which we can use as our "project" is rooted at the web folder.

### DIFF
--- a/vetur.config.js
+++ b/vetur.config.js
@@ -1,0 +1,5 @@
+module.exports = {
+  projects: [
+    './web',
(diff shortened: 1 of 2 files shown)


dchiquit:
This is obsoleted by #540

PR #540, opened 2021-01-01
Convert featureToggles to TS

AlmightyYakob:
I saw this as an opportunity to leverage composition API / Typescript, instead of throwing `ts-ignore` around. It may not have been totally necessary (as this specific toggle might not be around for too much longer), but I thought it was a good exercise nonetheless. This also includes a slight reorg to how we use plugins like composition-api and girder.

### DIFF
--- a/web/src/featureToggle.ts
+++ b/web/src/featureToggle.ts
@@ -28,23 +28,27 @@ export default featureToggles;
 // These functions are made available globally so they can be called from the console
 
 /** Enables a feature toggle. Only usable from the console. */
-function enable(featureToggle) {
+function enable(featureToggle: string) {
   if (!(featureToggle in featureToggles)) {
     throw `Feature toggle ${featureToggle} is not defined`;
   }
+  // @ts-ignore
   featureToggles[featureToggle] = true;
   console.log(`Feature toggle '${featureToggle}' enabled`);
 }
 
 /** Disables a feature toggle. Only usable from the console. */
-function disable(featureToggle) {
+function disable(featureToggle: string) {
   if (!(featureToggle in featureToggles)) {
     throw `Feature toggle ${featureToggle} is not defined`;
   }
+  // @ts-ignore
   featureToggles[featureToggle] = false;
   console.log(`Feature toggle '${featureToggle}' disabled`);
 }
 
 // Register enable and disable on the window object so they are accessible from the console
+// @ts-ignore
 window.enable = enable;
+// @ts-ignore
 window.disable = disable;
--- a/web/src/featureToggle.ts
+++ b/web/src/featureToggle.ts
@@ -18,7 +18,8 @@ const featureToggles = reactive({
 
 // Used to narrow string types
 function isToggle(toggle: string): toggle is keyof (typeof featureToggles) {
-  return Object.keys(featureToggles).includes(toggle);
+  // eslint-disable-next-line no-prototype-builtins
+  return featureToggles.hasOwnProperty(toggle);
 }
 
 // Register the feature toggles as computed properties on all components
--- a/web/src/featureToggle.ts
+++ b/web/src/featureToggle.ts
@@ -1,19 +1,29 @@
 /* eslint no-console: 0 */
-/* eslint no-throw-literal: 0 */
 import Vue from 'vue';
+import { reactive } from '@vue/composition-api';
+
+// Declare additions to the global `window` object
+declare global {
+  interface Window {
+    enable: (toggle: string) => void,
+    disable: (toggle: string) => void,
+  }
+}
 
 // Create a component to hold the feature toggles reactively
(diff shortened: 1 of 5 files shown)


web/src/featureToggle.ts
```diff
@@ -1,19 +1,29 @@
...
-      DJANGO_API: false,
-    };
-  },
+const featureToggles = reactive({
+  DJANGO_API: false,
 });
 
+// Used to narrow string types
+function isToggle(toggle: string): toggle is keyof (typeof featureToggles) {
+  return Object.keys(featureToggles).includes(toggle);
```
waxlamp, line 21: I think you can use `.hasOwnProperty()` here to avoid a listifcation and a linear-time search.

web/src/featureToggle.ts
```diff
@@ -1,19 +1,29 @@
...
+}
+
 // Register the feature toggles as computed properties on all components
 // Now feature toggles are available on all components and templates without any imports
+// TODO: Convert this to use composition API.
```
waxlamp, line 26: Probably should file an issue too.

web/src/main.ts
```diff
@@ -1,21 +1,25 @@
...
+
+// Import internal items
+import App from '@/App.vue';
 import router from '@/router';
 import store from '@/store';
 import { girderRest, publishRest } from '@/rest';
-import '@/title';
-
-Vue.use(Girder);
-Vue.use(VueCompositionAPI);
```
waxlamp, line 18: Actual question: what's the benefit of abstracting these into their own "plugin" files?

waxlamp (commented):
I can't speak to the plugin initialization reorg, but the type narrowing infrastructure and the use of `reactive()` both look good to me. Withholding approval because @dchiquit is more qualified to grant it.

web/src/featureToggle.ts
AlmightyYakob: You're correct. I originally didn't use that due to eslint warnings, but they seem not to apply here, so I'll fix it.

web/src/main.ts
AlmightyYakob: This was somewhat necessary to preserve correct code-style, as well as allow for the feature toggles to use composition API. Since we need to call `Vue.use(VueCompositionAPI)` before using it anywhere, abstracting it into an import allows for those plugins to simply be imported, and then anything that uses them can come after. This is also generally applicable; Any features that may require plugins to be loaded first can just me imported/used after the plugin imports (which should probably be done first among local imports in `main.ts`).

waxlamp: But the way it is now, the `Vue.use()` calls always happen exactly one time at application startup. Putting these calls into modules means that every import will cause a call to `Vue.use()` right? That may not be a problem, but I guess I don't understand how this leads to better code correctness.

For the record, I think even just the organization into logical units is a win. I just want to understand the other argument for doing it this way.

AlmightyYakob: In JS, imports only get evaluated once, so only the initial imports will cause `Vue.use()` to be called (in our case, on `main.ts` load). Regarding your other point, the reason this leads to code correctness is that otherwise, we'd have to import vue / composition-api, and then call `Vue.use`, then import`featureToggle` / whatever else would need to use it. Something like this:

```
import Vue from 'vue';
import VueCompositionAPI from '@vue/composition-api';

Vue.use(VueCompositionAPI);

import '@/featureToggle';
...
```

Which is generally bad code style, as imports should be at the top.

waxlamp: Ah gotcha. Thanks for the explanation!

AlmightyYakob: For reference, a similar use in Viame-Web https://github.com/VIAME/VIAME-Web/tree/master/client/platform/web-girder/plugins

web/src/featureToggle.ts
dchiquit: I'm not sure if this is possible. Composition API is an alternative way to write components, but this really doesn't have anything to do with that. This injects a computed variable called `DJANGO_API` onto every component so the DJANGO_API toggle can be used in templates. I don't think Composition API introduces an alternative way to do this, even if this does use the dreaded `mixin` keyword.

dchiquit (approved):

web/src/featureToggle.ts
AlmightyYakob: The normal way to solve this is by declaring a composition-api computed variable here named `DJANGO_API`, and importing it wherever it's needed. While this involves an additional import everywhere it's used, that actually doesn't bother me. The problem is that any components which want to import this computed variable must use composition API, or use the normal options `computed` block to wrap it. I didn't think this was worth the effort since it creeps the scope somewhat, and the toggle will likely be removed soon anyway. However if you think it's worth converting this as well, I'd be happy to.

PR #542, opened 2021-01-06
Convert DandisetsPage to composition API

dchiquit:
`asyncComputed` is provided by a plugin which proved challenging to get working with TypeScript. Rather than go through the effort of installing it properly, `watchEffect` can be used to replace the behavior, so I think we should just switch everything that uses `asyncComputed` to composition API (including DandisetsPage). `PublishFileBrowser` is the only other place where `asyncComputed` is used for non-girder purposes, so it shouldn't be a problem.


web/src/components/DandisetsPage.vue
```diff
@@ -103,120 +109,146 @@ export default {
...
+      page: String(page.value),
+      sortOption: String(sortOption.value),
+      sortDir: String(sortDir.value),
+    }));
+    watch(queryParams, (params) => {
+      ctx.root.$router.replace({
+        ...route,
+        // replace() takes a RawLocation, which has a name: string
+        // Route has a name: string | null, so we need to tweak this
+        name: route.name as string,
```
waxlamp, line 222: Does coercing the type like this lead to trouble? If `route.name` is never expected to be `null`, can you instead perform the type narrowing above (before the invocation of `replace()`)?

waxlamp (approved):
This all seems fine, but if you're not mostly confident in these changes yourself, would be good to get Jake to weigh in before merge.

web/src/components/DandisetsPage.vue
```diff
@@ -103,120 +109,146 @@ export default {
...
-    },
-    queryParams() {
-      const {
-        page, sortOption, sortDir,
-      } = this;
-      return { page, sortOption, sortDir };
-    },
-    dandisets() {
+    });
+    const girderDandisetRequest: Ref<null | any> = ref(null);
```
AlmightyYakob, line 170: I believe this could be more strictly typed

```suggestion
    const girderDandisetRequest: Ref<Omit<AxiosResponse, 'config' | 'request'> | null> = ref(null);
```

You'd have to import the `AxiosResponse` type though.

web/src/components/DandisetsPage.vue
AlmightyYakob: If we were to type narrow, we'd need to take an alternative action in the case where `route.name === null`. In the `RawLocation` type, `name` is optional. So while `null` isn't valid, we could cover our bases by instead writing:
```suggestion
        name: route.name || undefined,
```

AlmightyYakob (commented):
Looks good, just a couple typing suggestions

web/src/components/DandisetsPage.vue
dchiquit: I think it's worth it to just ignore types on Girder stuff where possible because it will all be deleted shortly anyway.

PR #543, opened 2021-01-06
Fix the url in the publishRest.assets method

dchiquit:
This has apparently been broken for a while. The `api/` prefix is no longer required. This will fix the download button for assets in the file browser.

### DIFF
--- a/web/src/rest/publish.ts
+++ b/web/src/rest/publish.ts
@@ -82,7 +82,7 @@ const publishRest = new Vue({
       try {
         const {
           data,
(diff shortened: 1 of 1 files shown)


AlmightyYakob:
There's no way to test this on the deploy preview, since in the API there's no assets in any of the dandisets, correct?

dchiquit:
@AlmightyYakob The draft of dandiset 000001 has some assets. I can't link to it since the toggle has to be on when the page loads, but you can navigate to it manually.

AlmightyYakob (approved):
:shipit:

PR #544, opened 2021-01-15
DJANGO_API publish button works

dchiquit:
You can now publish if you are an admin and DJANGO_API is toggled on.


web/src/rest/publish.ts
```diff
@@ -156,6 +156,10 @@ const publishRest = new Vue({
       const { data } = await client.get('users/search/?', { params: { username } });
       return data;
     },
+    async publish(identifier: string): Promise<Version> {
+      const { data } = await client.post(`dandisets/${identifier}/versions/draft/publish/`);
```
waxlamp, line 160: What (if anything) should we do if the POST fails?

waxlamp (approved):

web/src/rest/publish.ts
dchiquit: I'd expect some kind of :warning: GUI alert :warning:, but right now we have no error notifications anywhere in the app. Eventually we should think about failure paths, but the time is not yet nigh.

PR #547, opened 2021-01-20
Optimize dandisets page

dchiquit:
Use the `most_recent_version` field now provided by the API. Most of this will need to be refactored after flipping the DJANGO_API toggle.

### DIFF
--- a/web/src/types/index.ts
+++ b/web/src/types/index.ts
@@ -9,6 +9,8 @@ export interface Dandiset {
   identifier: string,
   created: string,
   modified: string,
+  // TODO most_recent_version is girderized, it should have type Version
+  most_recent_version?: any,
 }
 
 export interface Version {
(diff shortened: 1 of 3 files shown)


web/src/types/index.ts
```diff
@@ -9,6 +9,8 @@ export interface Dandiset {
   identifier: string,
   created: string,
   modified: string,
+  // TODO most_recent_version is girderized, it should have type Version
+  most_recent_version?: any,
```
AlmightyYakob, line 13: To clarify, you're saying that we should change this type to `Version` once Girder is gone, right? Should we file an issue?

AlmightyYakob (approved):
Tested and it's much faster, code looks good. Just left a small comment :+1:

web/src/types/index.ts
dchiquit: Correct. The only reason it's any is because `girderize` isn't typed. As we remove `girderize`, I think we will run into this and retype it. I guess an issue can't hurt, though.

PR #552, opened 2021-01-25
Properly handle Django user initials

dchiquit:
Fix the `??` that show up when logging in with Django. Girder's user object has `firstName`, while Django's has `first_name`.

### DIFF
--- a/web/src/components/AppBar/UserMenu.vue
+++ b/web/src/components/AppBar/UserMenu.vue
@@ -47,10 +47,17 @@ export default {
     user,
     userInitials() {
       if (this.user) {
+        // TODO Girder code, delete
         const { firstName, lastName } = this.user;
         if (firstName && lastName) {
           return (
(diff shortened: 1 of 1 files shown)


web/src/components/AppBar/UserMenu.vue
```diff
@@ -47,12 +47,19 @@ export default {
...
           return (
             firstName.charAt(0).toLocaleUpperCase() + lastName.charAt(0).toLocaleUpperCase()
           );
         }
+        const { first_name, last_name } = this.user;
+        if (first_name && last_name) {
+          return (
+            first_name.charAt(0).toLocaleUpperCase() + last_name.charAt(0).toLocaleUpperCase()
+          );
+        }
```
AlmightyYakob, line 62: I think if we're going to implicitly check if the Django toggle is on, we should make the comment here more clear. Also, adding a comment above the second block would be nice too

AlmightyYakob (approved):
Looks good, just one comment.

PR #556, opened 2021-01-26
Add github action to update django-api with master

AlmightyYakob:


.github/workflows/sync-django-api.yml
```diff
@@ -0,0 +1,24 @@
+name: Update django-api
+
+on:
+  push:
```
dchiquit, line 4: Will this trigger on PR merge?

AlmightyYakob: A merge causes a push to master so yes

dchiquit (approved):

PR #564, opened 2021-02-05
Update application with new schema and metadata editor

AlmightyYakob:
This brings over some code from the https://github.com/dandi/dandiarchive/tree/process-schema-and-model branch, and will make #423 obsolete. The major changes are:

* Added the new Metadata Editor (`Meditor.vue`). Currently, the `DJANGO_API` feature toggle will control whether the old/new metadata editor will be used.
* Added new components for rendering various parts of the schemas (at https://github.com/dandi/dandiarchive/tree/new-schema-meditor/web/src/components/schema).
* Update `DandisetLandingView.vue` to use the new rendering components

I've tried to keep the commits well scoped, since there's a lot of new code. Please take your time!


AlmightyYakob:
The E2E tests seem to be failing, I'm not sure how this is related to my change. Will investigate.

dchiquito:
I enabled login for this branch preview, so everything can be tested using https://deploy-preview-564--gui-dandiarchive-org.netlify.app/#/dandiset

web/src/views/DandisetLandingView/DandisetLandingView.vue
```diff
@@ -1,12 +1,22 @@
...
-      :create="create"
-      @close="edit = false"
-    />
+  <div
+    v-if="schema"
+    v-page-title="meta.name"
+  >
+    <template v-if="edit && Object.entries(meta).length">
+      <meditor
+        v-if="toggles.DJANGO_API"
```
dchiquito, line 8: This isn't really documented anywhere sadly, but `DJANGO_API` is registered as a property available to all components: https://github.com/dandi/dandiarchive/blob/master/web/src/featureToggle.ts#L28-L34

I did this so that components could avoid doing exactly this. However I thought it was weird to write `if (this.DJANGO_API) {...}` in the component scripts, so I still import `featureToggle` so that I can use `toggles.DJANGO_API` instead. Sorry for that magic.

web/src/views/DandisetLandingView/ListingComponent.vue
```diff
@@ -80,24 +30,106 @@ export default {
...
-      type: Boolean,
-      default: false,
+    field: {
+      // The key in the parent object
+      type: String,
+      required: true,
     },
   },
   computed: {
+    renderComponent() {
```
dchiquito, line 40:
```suggestion
    renderedComponent() {
```
`renderComponent` sounds like an action rather than a string.

web/src/views/DandisetLandingView/ListingComponent.vue
```diff
@@ -80,24 +30,106 @@ export default {
...
+        case ObjectComponent:
+          className += 'ml-4';
+          break;
+
+        default:
+          break;
+      }
+
+      return className;
+    },
```
dchiquito, line 132: This is a little extra. Is this designed to be extensible, or just leftover code?

web/src/views/DandisetLandingView/ListingComponent.vue
```diff
@@ -1,72 +1,22 @@
...
-    </template>
-    <template v-else>
-      <!-- Base Case -->
-      <template v-if="!root">
-        <b>{{ schema.title }}</b>:
-      </template>
-      {{ data }}
-    </template>
-  </div>
+  <component
```
dchiquito, line 2: I can't figure out where this `component` element comes from. Am I missing something obvious?

dchiquito (approved):
The "Access" popup in the meditor is empty on https://deploy-preview-564--gui-dandiarchive-org.netlify.app/#/dandiset/000004. Other than that, everything seems to work as intended, to the best of my limited knowledge.

I called out a few minor stylistic things in the code and asked a few questions for my own enlightenment.

:shipit:

AlmightyYakob:
> The "Access" popup in the meditor is empty on [deploy-preview-564--gui-dandiarchive-org.netlify.app/#/dandiset/000004](https://deploy-preview-564--gui-dandiarchive-org.netlify.app/#/dandiset/000004). Other than that, everything seems to work as intended, to the best of my limited knowledge.
> 
> I called out a few minor stylistic things in the code and asked a few questions for my own enlightenment.
> 
> :shipit:

Dandiset `000004` doesn't appear to exist in the Django API. Is there a different dandiset that you were referring to or that replicates this problem? Also, are you referring to the "Access Type" section of metadata?

dchiquito:
image
I loaded https://deploy-preview-564--gui-dandiarchive-org.netlify.app/#/dandiset/000004, clicked on `Edit Metadata`, then toggled on `DJANGO_API`. Maybe that just put the schema in a bad state.

web/src/components/schema/Object.vue
```diff
@@ -0,0 +1,97 @@
...
+      required: true,
+    },
+    schema: {
+      // The data at the matching level of schema
+      type: Object as PropType<JSONSchema7>,
+      required: false,
+      default: null,
+    },
+    options: {
+      type: Object as PropType<Record<string, any>>,
```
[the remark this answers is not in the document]
AlmightyYakob: `options` is intended to be a way to pass settings to the rendering sub components (`Object.vue` being one of those sub components) to modify behavior. Since the actual component being rendered is dynamic, the `options` prop is used to be agnostic to the component, while still allowing for some functionality. However I think you're right, these could probably be typed, so I can look into adding that. If it seems like it'll be too much effort right now, I can file an issue.

web/src/main.ts
```diff
@@ -30,7 +31,10 @@ sync(store, router);
 
 Promise.all([publishRest.restoreLogin(), girderRest.fetchUser()]).then(() => {
   new Vue({
-    provide: { girderRest },
+    setup() {
+      provide('store', store);
```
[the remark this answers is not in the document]
AlmightyYakob: I believe there's some edge cases where using `ctx.root.$store` or `this.$store` is needed, so this provides that.

web/src/utils/schema/editor.ts
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * Manages the interface between the source data/schema, and the changes necessary for it to
+ * operate correctly with the Meditor.
+ */
+class EditorInterface {
+  // Not guaranteed to be up to date, use getModel()
+  private model: DandiModel;
+
+  basicModel: DandiModel;
+  complexModel: DandiModel;
```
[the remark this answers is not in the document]
AlmightyYakob: It's not two different modes, but the same mode split up and used in several places. If the `VJSF` component was not split up, and rendered the whole schema at once, it would get pretty unwieldy pretty fast, since there's a lot of data that can exist on a dandiset, it gets hard to see the when one section starts vs. ends, etc. To get around that, the schema + model are split into two different sections: The "basic" schema + model, and the "complex" schema + model.

The basic schema is rendered normally like the original would have been, but these fields are all simple or "basic" types, like strings, numbers, booleans, etc., as well as arrays of these types (since in the component arrays of something like strings is just rendered as a multi-selection vuetify form).

The complex schema is all other schema fields that aren't captured by the basic schema. Each of these complex fields are rendered individually (with an individual `VJSF` component) underneath those button dialogs at the top of the meditor.

Hopefully this clears things up.

AlmightyYakob:
> image
> I loaded [deploy-preview-564--gui-dandiarchive-org.netlify.app/#/dandiset/000004](https://deploy-preview-564--gui-dandiarchive-org.netlify.app/#/dandiset/000004), clicked on `Edit Metadata`, then toggled on `DJANGO_API`. Maybe that just put the schema in a bad state.

Ahh yes, that's the issue. This is actually trying to use the old schema, rather than the new one. To properly use this, you should instead activate the `DJANGO_API` toggle from outside of the DLP (I usually do it on the dandiset listing page), and then select a dandiset that you want to edit, etc.

I could probably update the old component to make use of reactive schema changes, but I don't think that's particularly worth the time, since this is only an artifact of activating the toggle while the site is running.

web/src/views/DandisetLandingView/DandisetLandingView.vue
AlmightyYakob: I'm a bit confused, are you recommending a change here? I actually think I used `toggles.DJANGO_API` on purpose, because I thought that in the future the mixin might go away. If you think there's something fundamentally wrong with this, I can change it.

web/src/views/DandisetLandingView/ListingComponent.vue
AlmightyYakob: Yes it's designed to be extensible. Similar to the other functions that contain these switches, this is designed to add styling classes to the specific rendering components if they need it.

web/src/views/DandisetLandingView/ListingComponent.vue
AlmightyYakob: This is a native `Vue` thing. `component` is what's used if you want to render a dynamic component, with the exact component you want to render being specific by the `is` prop. Docs: https://vuejs.org/v2/guide/components.html#Dynamic-Components

web/src/views/DandisetLandingView/DandisetLandingView.vue
dchiquito: My recommendation is:
```suggestion
        v-if="DJANGO_API"
```
and remove `toggles` from the return value in `setup()`. My intention was for the mixin to be permanent; if we ever added a new toggle, we would also add it to the mixin.

Now that I think about it though, maybe it makes more sense to expose `toggles` in the mixin rather than `DJANGO_API` directly. I was mostly trying to avoid having to include that `toggles` in the return value of `setup()` so that the footprint of using a toggle would be as small as possible.

If you think that's a bad pattern, I'm open to changing it.

waxlamp (commented):
Looks great, Jake. I don't have any substantive recommendations, so leaving it to you and Daniel to approve/merge.

satra:
@AlmightyYakob - is there still a basic and complex model in play after the schema updates and my codepen demo (https://codepen.io/satra-the-reactor/pen/NWbKbrj?editors=1010), which seems fine in terms of complexity?  if so, can you please remind me what constitutes basic and complex?

web/src/views/DandisetLandingView/DandisetLandingView.vue
AlmightyYakob: I think the idea of reducing the footprint of the feature toggles is a good one. Since right now the only toggle we have is `DJANGO_API`, and that'll be going away soon, I can just use the mixin value.

web/src/utils/typing.ts
```diff
@@ -0,0 +1,4 @@
+// eslint-disable-next-line import/prefer-default-export
```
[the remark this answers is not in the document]
AlmightyYakob: :man_shrugging: I never really liked it either. I could add it to the `.eslintrc.js`.

AlmightyYakob:
> @AlmightyYakob - is there still a basic and complex model in play after the schema updates and my codepen demo ([codepen.io/satra-the-reactor/pen/NWbKbrj?editors=1010](https://codepen.io/satra-the-reactor/pen/NWbKbrj?editors=1010)), which seems fine in terms of complexity? if so, can you please remind me what constitutes basic and complex?

Yes there is. [This](https://github.com/dandi/dandiarchive/pull/564#discussion_r577937170) comment from above explains the difference and why it's still used. The main point is that when there's lots of data, it gets hard to find the section you want to add to or edit, and splitting the schema and model up helps with that.

web/src/utils/constants.js
```diff
@@ -0,0 +1,16 @@
+const dandiUrl = 'https://dandiarchive.org';
+const dandiAboutUrl = 'https://dandiarchive.org/about';
+const dandiDocumentationUrl = 'https://www.dandiarchive.org/handbook/10_using_dandi/';
+
+const schemaRelease = process.env.VUE_APP_DANDISET_SCHEMA_RELEASE || '1.0.0-rc1';
```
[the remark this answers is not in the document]
AlmightyYakob: You make a good point, but I think it's fine to leave for now, since once https://github.com/dandi/dandi-api/pull/83 is merged, this logic will change to fetch the schema url from the API anyway.

web/src/utils/schema/utils.ts
```diff
@@ -0,0 +1,93 @@
...
+    required: newRequired,
+  };
+
+  // Description isn't needed and just causes rendering issues
+  delete newSchema.description;
+  return newSchema;
+}
+
+export function populateEmptyArrays(schema: JSONSchema7, model: DandiModel) {
+  // TODO: May need to create a similar function for objects
```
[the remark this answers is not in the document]
AlmightyYakob: I think this comment is actually old and no longer applies. I'll remove it.

satra:
@AlmightyYakob - that comment points to basic/complex schema types which is orthogonal to needed information by the schema. the UI should be split not by schema complexity but user needs. thus the more important pieces needed (title, description, contributor, keywords, resources, project) vs less important pieces.

also isn't this where the schema augmentation could help to break things into paged modes and other ui elements rather than breaking things on the meditor side. i am still trying to compress as much as we can into the schema and rely on the component that developing things on the meditor side.

web/src/views/DandisetLandingView/DandisetLandingView.vue
AlmightyYakob: I actually forgot about one thing which I'd like to point out. Currently, anywhere we use feature toggles we still import `toggles` in the component script, to make any use of it. I don't think we actually _need_ to do this, since the computed prop should still be present via the mixin, but if we don't import and use it that way eslint will yell at us. I don't quite like that we're using the computed value from the mixin in the templates, and importing it for the script. I'll change this to be consistent with the rest of the app, but just thought I'd make my concerns known.

web/src/utils/typing.ts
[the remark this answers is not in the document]
waxlamp: No need for a change here now, I was just wondering out loud.

waxlamp:
> @AlmightyYakob - that comment points to basic/complex schema types which is orthogonal to needed information by the schema. the UI should be split not by schema complexity but user needs. thus the more important pieces needed (title, description, contributor, keywords, resources, project) vs less important pieces.
> 
> also isn't this where the schema augmentation could help to break things into paged modes and other ui elements rather than breaking things on the meditor side. i am still trying to compress as much as we can into the schema and rely on the component that developing things on the meditor side.

@satra, I think it's good to think of Jake's "simple/complex" distinction as a heuristic for making the meditor usable in the short term. If you have better heuristics or other ideas in mind for how to make this aspect of the meditor more robust, we should explore them in a followup effort. Many improvements will appear in the meditor automatically if the schema is changed--for example, several of the schema items can have better descriptions which will then appear as the hint text in the meditor. Other changes along the lines you've suggested may require a more advanced signalling mechanism between the schema and the meditor, and those are the ones that should be sequestered to a dedicated followup so that we can get this basic but solid version landed.

web/src/components/schema/Object.vue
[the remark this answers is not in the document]
AlmightyYakob: 14980da

PR #590, opened 2021-03-10
Enable app startup with DJANGO_API feature flag set

waxlamp:
Even with the DJANGO_API feature flag set to `true` in the source code, the app will not correctly start up in dev without both a Girder and a Django server running. This is because the app startup routine was unconditionally asking for login status from both servers. This PR uses the feature flag to hit one or the other of the two servers.

### DIFF
--- a/web/src/main.ts
+++ b/web/src/main.ts
@@ -13,7 +13,7 @@ import '@/plugins/composition';
 import '@/plugins/girder';
 
 // Import custom behavior
-import '@/featureToggle';
+import featureToggles from '@/featureToggle';
 import '@/title';
 
 // Import internal items
@@ -29,7 +29,10 @@ Sentry.init({
 
 sync(store, router);
 
-Promise.all([publishRest.restoreLogin(), girderRest.fetchUser()]).then(() => {
+// Get login info from the correct server.
+const loginCall = featureToggles.DJANGO_API ? publishRest.restoreLogin() : girderRest.fetchUser();
+
+loginCall.then(() => {
   new Vue({
     setup() {
       provide('store', store);


AlmightyYakob:
This LGTM and I can't find anywhere in the code where `girderRest` is used when `DJANGO_API` is enabled _except_ right here:

https://github.com/dandi/dandiarchive/blob/12120b29a20a23991689bdb52d02753c0d62aa86/web/src/views/UserRegisterView/UserRegisterView.vue#L45

You can also see this by going to the gui beta depoyment register [page](https://gui-beta-dandiarchive-org.netlify.app/#/user/register) and looking at the network tab.

I suspect this is a bug, and should be changed to use dandi-api somehow. @dchiquito can you confirm this?

dchiquito (changes requested):
I actually fixed this [here](https://github.com/dandi/dandiarchive/commit/8057c0b9c88c79a5118c8d40dc143fb96044b2e2), but kept that in the `django-api` branch instead of merging it into master and rebasing it. I did not realize how disruptive it was.

My solution is functionally identical to this, but I'm pretty sure it will cause merge errors when `django-api` is rebased onto this. I will clean it up momentarily.

dchiquito:
https://github.com/dandi/dandiarchive/pull/593 should fix the problem with no merge/rebase conflicts.

dchiquito:
@AlmightyYakob I think you are correct. I think it's still working in production because Girder 3 and Django are using the same GitHub OAuth application, so registering with one also registers you with the other. It's definitely a bug that needs to be fixed for migration though. Good catch

waxlamp:
Closing in favor of #593.

PR #591, opened 2021-03-10
Add a delete asset button

### LINKED ISSUE #576: Delete asset button

waxlamp:
This PR adds a "delete asset" button to the dandiset file browser. It includes a confirmation dialog, and it takes care to update the asset list after deletion succeeds.

This is pretty simple, so it may need things like error handling, or a better confirmation dialog. But it's at least a start.

One possible pitfall: I un-async-computified the `items` calculation. If there's a better way to accomplish the asset list refresh, please let me know.

Closes #576

### DIFF
--- a/web/src/rest/publish.ts
+++ b/web/src/rest/publish.ts
@@ -189,6 +189,9 @@ const publishRest = new Vue({
       const { uuid } = asset;
       return `${publishApiRoot}dandisets/${identifier}/versions/${version}/assets/${uuid}/download`;
     },
+    async deleteAsset(identifier: string, version: string, uuid: string): Promise<AxiosResponse> {
+      return client.delete(`${publishApiRoot}dandisets/${identifier}/versions/${version}/assets/${uuid}`);
+    },
   },
 });
 
--- a/web/src/rest/publish.ts
+++ b/web/src/rest/publish.ts
@@ -190,7 +190,7 @@ const publishRest = new Vue({
       return `${publishApiRoot}dandisets/${identifier}/versions/${version}/assets/${uuid}/download`;
     },
     async deleteAsset(identifier: string, version: string, uuid: string): Promise<AxiosResponse> {
-      return client.delete(`${publishApiRoot}dandisets/${identifier}/versions/${version}/assets/${uuid}`);
+      return client.delete(`${publishApiRoot}dandisets/${identifier}/versions/${version}/assets/${uuid}/`);
     },
   },
 });
--- a/web/src/views/FileBrowserView/PublishFileBrowser.vue
+++ b/web/src/views/FileBrowserView/PublishFileBrowser.vue
@@ -101,6 +101,7 @@
 
               <v-list-item-action>
                 <v-btn
+                  v-if="!item.folder"
                   icon
                   @click="openDialog(item.name)"
                 >
--- a/web/src/views/FileBrowserView/PublishFileBrowser.vue
+++ b/web/src/views/FileBrowserView/PublishFileBrowser.vue
@@ -112,10 +112,10 @@
                 </v-btn>
               </v-list-item-action>
 
-              <v-list-item-action v-if="itemDownloads[item.name]">
+              <v-list-item-action v-if="itemDownload(item.name)">
                 <v-btn
                   icon
-                  :href="itemDownloads[item.name]"
+                  :href="itemDownload(item.name)"
                 >
                   <v-icon color="primary">
                     mdi-download
@@ -154,8 +154,7 @@ export default {
     return {
       rootDirectory,
       location: rootDirectory,
-      itemDownloads: {},
-      itemDeletes: {},
+      itemData: {},
       owners: [],
       dialogActive: false,
       dialogName: '',
@@ -220,8 +219,6 @@ export default {
       }
 
       // Create download and delete links in local data for each item in items
-      this.itemDownloads = {};
-      this.itemDeletes = {};
       const { identifier, version, location } = this;
 
       items.filter((x) => !x.folder).map(async (item) => {
@@ -230,16 +227,13 @@ export default {
           results: [asset],
         } = await publishRest.assets(identifier, version, { params: { path: relativePath } });
 
-        this.$set(
-          this.itemDownloads,
-          item.name,
-          publishRest.assetDownloadURI(identifier, version, asset),
-        );
-
-        this.$set(this.itemDeletes, item.name, {
-          identifier,
-          version,
-          uuid: asset.uuid,
+        this.$set(this.itemData, item.name, {
+          download: publishRest.assetDownloadURI(identifier, version, asset),
+          delete: {
+            identifier,
+            version,
+            uuid: asset.uuid,
+          },
         });
       });
     },
@@ -274,8 +268,16 @@ export default {
       this.dialogActive = true;
     },
 
+    itemDownload(name) {
+      return this.itemData[name] && this.itemData[name].download;
+    },
+
+    itemDelete(name) {
+      return this.itemData[name] && this.itemData[name].delete;
+    },
+
     async deleteAsset(name) {
-      const asset = this.itemDeletes[name];
+      const asset = this.itemDelete(name);
       if (asset !== undefined) {
         // Delete the asset on the server.
         const { identifier, version, uuid } = asset;
--- a/web/src/views/FileBrowserView/PublishFileBrowser.vue
+++ b/web/src/views/FileBrowserView/PublishFileBrowser.vue
@@ -1,5 +1,38 @@
 <template>
   <v-container>
+    <v-dialog
+      v-model="dialog.active"
+      persistent
+      max-width="60vh"
+    >
+      <v-card>
+        <v-card-title class="headline">
+          Really delete this asset?
+        </v-card-title>
+
+        <v-card-text>
+          Are you sure you want to delete asset <span
+          class="font-italic">{{dialog.name}}</span>?
+            <strong>This action cannot be undone.</strong>
+        </v-card-text>
+
+        <v-card-actions>
+          <v-spacer />
+          <v-btn
+            @click="dialog.active = false"
+          >
+            Cancel
+          </v-btn>
+          <v-btn
+            color="error"
+            @click="deleteAsset(dialog.name)"
+          >
+            Yes
+          </v-btn>
+        </v-card-actions>
+      </v-card>
+    </v-dialog>
+
     <v-row>
       <v-col :cols="12">
         <v-card>
@@ -67,45 +100,14 @@
               <v-spacer />
 
               <v-list-item-action>
-                <v-dialog
-                  v-model="deleteConfirmationDialog"
-                  persistent
-                  max-width="60vh"
+                <v-btn
+                  icon
+                  @click="openDialog(item.name)"
                 >
-                  <template v-slot:activator="{ on }">
-                    <v-btn
-                      icon
-                      v-on="on"
-                    >
-                      <v-icon color="error">
-                        mdi-delete
-                      </v-icon>
-                    </v-btn>
-                  </template>
-                  <v-card>
-                    <v-card-title class="headline">
-                      Really delete this asset?
-                    </v-card-title>
-                    <v-card-text>
-                      Are you sure you want to delete this asset? <strong>This
-                        action cannot be undone.</strong>
-                    </v-card-text>
-                    <v-card-actions>
-                      <v-spacer />
-                      <v-btn
-                        @click="deleteConfirmationDialog = false"
-                      >
-                        Cancel
-                      </v-btn>
-                      <v-btn
-                        color="error"
-                        @click="deleteAsset(item.name)"
-                      >
-                        Yes
-                      </v-btn>
-                    </v-card-actions>
-                  </v-card>
-                </v-dialog>
+                  <v-icon color="error">
+                    mdi-delete
+                  </v-icon>
+                </v-btn>
               </v-list-item-action>
 
               <v-list-item-action v-if="itemDownloads[item.name]">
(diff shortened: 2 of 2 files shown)


dchiquito:
I think this should be rebased onto https://github.com/dandi/dandiarchive/pull/593

waxlamp:
> I think this should be rebased onto #593

I'll just wait until #593 hits master, then rebase thereon (unless there's a pitfall with that plan).

web/src/rest/publish.ts
```diff
@@ -189,6 +189,9 @@ const publishRest = new Vue({
       const { uuid } = asset;
       return `${publishApiRoot}dandisets/${identifier}/versions/${version}/assets/${uuid}/download`;
     },
+    async deleteAsset(identifier: string, version: string, uuid: string): Promise<AxiosResponse> {
+      return client.delete(`${publishApiRoot}dandisets/${identifier}/versions/${version}/assets/${uuid}`);
```
dchiquito, line 193:
```suggestion
      return client.delete(`${publishApiRoot}dandisets/${identifier}/versions/${version}/assets/${uuid}/`);
```
The server should automatically redirect you, but I think it's good practice to have the right URLs wherever possible.

web/src/views/FileBrowserView/PublishFileBrowser.vue
```diff
@@ -65,6 +65,49 @@
               </v-icon>
               {{ item.name }}
               <v-spacer />
+
+              <v-list-item-action>
```
dchiquito, line 69: This button still renders when you are not logged, and presumably when you aren't an admin. This should have a `v-if` that checks if `@rest.user` is in the owners list for the current dandiset, or if the user is an admin.

dchiquito, line 69: This dialog also renders for folders. While we can debate if we want to allow that, currently the API doesn't support deleting a parent folder, so it should only render on assets. Probably you want something like `v-if="!item.folder"`.

web/src/views/FileBrowserView/PublishFileBrowser.vue
```diff
@@ -194,6 +230,34 @@ export default {
         this.location = `${this.location}${name}`;
       }
     },
+
+    async refreshItems() {
+      const { version, identifier, location } = this;
+
+      this.loading = true;
```
dchiquito, line 237: I just learned that `asyncComputed` gives you this feature for free: https://github.com/foxbenjaminfox/vue-async-computed#computation-status

web/src/views/FileBrowserView/PublishFileBrowser.vue
```diff
@@ -156,8 +180,9 @@ export default {
         return;
       }
 
-      // Create the download link in itemDownloads for each item in items
+      // Create download and delete links in local data for each item in items
       this.itemDownloads = {};
+      this.itemDeletes = {};
```
dchiquito, line 185: I think this is a reimplementation of https://github.com/foxbenjaminfox/vue-async-computed#recalculation. You can probably keep the original `asyncComputed` and just add some kind of hook to recalculate it.

dchiquito (changes requested):
When I go to https://deploy-preview-591--gui-dandiarchive-org.netlify.app/#/dandiset/000003/draft/files?location=, toggle on DJANGO_API, and click on any of the links, it actually change the content in the view. I suspect there's some edge case that isn't triggering a refresh.

I think this would be solved indirectly by adopting some of my suggestions regarding `asyncComputed`.

web/src/views/FileBrowserView/PublishFileBrowser.vue
waxlamp: Oh excellent. I'll revert the commit and do this in its place.

web/src/views/FileBrowserView/PublishFileBrowser.vue
waxlamp: Accomplished via 5ed937bbe7b6460782599fbc889bfde9dded6a51 and a couple of other commits.

waxlamp: Done via 3df03301b56f852bd7cd0e9e8bfe5abb1337daf3 and ebabad3ffd85f099e89183a757985373a4c3d1fd.

web/src/views/FileBrowserView/PublishFileBrowser.vue
waxlamp: See 289ac4fc0c6e7df8e5d84449006751099181f084.

web/src/views/FileBrowserView/PublishFileBrowser.vue
waxlamp: I cut that commit out of the commit stream, and used a better approach in dfd267457867743c3c74e59568569f4c3a5adafc.

web/src/views/FileBrowserView/PublishFileBrowser.vue
```diff
@@ -194,6 +253,33 @@ export default {
...
     },
+
+    showDelete(item) {
+      const me = publishRest.user ? publishRest.user.username : null;
+
+      const isAdmin = me && publishRest.user.admin;
+      const isOwner = me && this.owners.includes(me);
+
+      return !item.folder && (isAdmin || isOwner);
+    },
```
AlmightyYakob, line 264: This is a small thing, but I think the `me`, `isAdmin`, and `isOwner` parts should be moved to a computed value, since the admin/owner status won't change item to item. I think a `me` computed value and a simple `isOwner` computed value (which would encompass the current `isAdmin` and `isOwner` variables, since admins qualify as owners anyway) would be fine.

web/src/views/FileBrowserView/PublishFileBrowser.vue
```diff
@@ -107,8 +154,13 @@ export default {
...
       rootDirectory,
       location: rootDirectory,
-      loading: false,
       itemDownloads: {},
+      itemDeletes: {},
+      owners: [],
+      dialog: {
+        active: false,
+        name: '',
+      },
```
AlmightyYakob, line 163: This is another small thing, but I think in general we should discourage using objects to contain related information like this. The main reason being if down the line someone tries to watch/use/depend on the top-level `dialog` field, it will fail, because that field isn't reactive (only the inner items are). I'm curious what your opinions on this are @waxlamp and @dchiquito. All that said, if you have a strong enough reason to keep it I'm fine with that too.

web/src/views/FileBrowserView/PublishFileBrowser.vue
```diff
@@ -41,7 +75,7 @@
             </template>
           </v-card-title>
           <v-progress-linear
-            v-if="loading"
+            v-if="$asyncComputed.items.updating"
```
AlmightyYakob, line 78: As a matter of style, could this just be a check of `items === null`? Since the default value for that asyncComputed variable is `null`?

web/src/views/FileBrowserView/PublishFileBrowser.vue
waxlamp: This is a good idea, but will it work with the `publishRest` value? That's a reactive store type of thing, but standalone. If I mention it in a computed prop, will the prop correctly update when `publishRest` changes?

web/src/views/FileBrowserView/PublishFileBrowser.vue
waxlamp: Ideally the `<v-dialog>` would be embedded in its own component, but I didn't want to do that here. I am fine with splatting these props into the top level, but I also am not too worried about someone trying to `watch` the object value, since it's sort of Vue-knowledge not to do that. As a self-counterpoint, we could avoid that error altogether by splatting these values.

So I'll wait to hear from @dchiquito but I am ok with unraveling this object in this case.

web/src/views/FileBrowserView/PublishFileBrowser.vue
waxlamp: I think this is actually clearer: it says that the progress bar should show precisely when the `items` value is updating, which is the intended semantics.

What is the style objection to doing it this way?

AlmightyYakob: I guess that you're relying on an implicitly injected Vue property :man_shrugging:. Honestly I don't feel to strongly about this, do whatever you feel is best.

dchiquito: Regardless of style, if the update fails somehow and leaves `items` as `null`, using `items === null` will spin forever, which I think is worse than showing an empty list.

web/src/views/FileBrowserView/PublishFileBrowser.vue
dchiquito: I'm pretty sure the only way this code as written will work is if `publishRest` is reactive. I can confirm that clicking `Logout` while browsing assets removes the delete icon, so I'm pretty sure this will work fine as a computed property.

web/src/views/FileBrowserView/PublishFileBrowser.vue
```diff
@@ -156,8 +208,9 @@ export default {
         return;
       }
 
-      // Create the download link in itemDownloads for each item in items
+      // Create download and delete links in local data for each item in items
       this.itemDownloads = {};
+      this.itemDeletes = {};
```
dchiquito, line 213: I think these might make more sense joined into one map that is accessed for both downloading and deleting. I don't feel strongly, just a thought.

web/src/views/FileBrowserView/PublishFileBrowser.vue
dchiquito: I did not know that `dialog` would not be reactive, so make of that what you will. I prefer the object grouping syntactically, but I think I would prefer keeping all `data` reactive, if I had to have an opinion on this matter.

Vue3 pls

web/src/views/FileBrowserView/PublishFileBrowser.vue
waxlamp: > I guess that you're relying on an implicitly injected Vue property . Honestly I don't feel to strongly about this, do whatever you feel is best.

Since it's coming from a plugin we're actively using, I don't have any problem making use of the `$asyncComputed` injected property--it's about the same to me as making use of `$props` or anything else Vuey. @dchiquito's argument is strong as well, so let's keep this as it is.

web/src/views/FileBrowserView/PublishFileBrowser.vue
waxlamp: See da42a896b75e3d89ef4c19044fa2aa36bf14acc9.

web/src/views/FileBrowserView/PublishFileBrowser.vue
waxlamp: > Vue3 pls

I agree with this plaintive cry for help and would like to subscribe to your newsletter.

See 1e0f47580e8d653130fbcbd5cf5a5c067321c97e.

web/src/views/FileBrowserView/PublishFileBrowser.vue
waxlamp: See cf081408de80cb931771ef3c7279966d62c5d763.

I am not sure this is really better than what we had (it is not terrible either), but let me know what you think.

dchiquito (approved):
LGTM

web/src/views/FileBrowserView/PublishFileBrowser.vue
```diff
@@ -166,11 +227,14 @@ export default {
...
-          item.name,
-          publishRest.assetDownloadURI(identifier, version, asset),
-        );
+        this.$set(this.itemData, item.name, {
+          download: publishRest.assetDownloadURI(identifier, version, asset),
+          delete: {
+            identifier,
+            version,
+            uuid: asset.uuid,
+          },
```
AlmightyYakob, line 236: I don't see any particular reason to store a `delete` field here. For deletion, all you need is the identifier, the version, and the asset uuid. The identifier and version won't change anyway, so all you really need to store is the uuid. I'd propose something like this:

```suggestion
          uuid,
```

Then below, you would just get the `itemData` entry and look at the `uuid` field. I also think you could get rid of the `itemDownload` and `itemDelete` functions, and just access `this.itemData` and the respective fields when/where you need them.

Lmk what you think.

waxlamp: Thanks for catching this. Still need `itemDownload` for convenience (because I still want to check for `undefined` since I don't really understand the type safety of these values), but check out 5074f5ef7da82b803beb9b5c13766b9b42044142 for your main suggestion.

web/src/views/FileBrowserView/PublishFileBrowser.vue
waxlamp: I assume this thread is resolved; open it back up if not.

web/src/views/FileBrowserView/PublishFileBrowser.vue
AlmightyYakob: FWIW, this is the only place `itemDownload` is used:

```
<v-list-item-action v-if="itemDownload(item.name)">
  <v-btn
    icon
    :href="itemDownload(item.name)"
  >
  ...
```

My main point was that you could instead just do the following:

```
<v-list-item-action v-if="itemData[name]">
  <v-btn
    icon
    :href="itemData[name].download"
  >
  ...
```

The `download` field is always present on each `itemData` entry, so this wouldn't be an issue. However, it really doesn't matter to me, and what you've done addresses the core of my complaint, so I'll approve this. If you do happen to make the additional change I mentioned above (as well as removing the `itemDownload` function) anyway, my approval still stands.

AlmightyYakob (approved):

web/src/views/FileBrowserView/PublishFileBrowser.vue
waxlamp: Thanks for clarifying. Check out 7cd32721dd912c0e10ba95e73560927434b727c5.

mgrauer:
Nice, thanks @waxlamp !

PR #593, opened 2021-03-11
Only load one user on app load

dchiquito:
Cherry picking my fix for #590 out of the `django-api` branch and into master.

### DIFF
--- a/web/src/main.ts
+++ b/web/src/main.ts
@@ -29,12 +29,11 @@ Sentry.init({
 
 sync(store, router);
 
-async function loadUser() {
+function loadUser() {
   if (toggles.DJANGO_API) {
     return publishRest.restoreLogin();
-  } else {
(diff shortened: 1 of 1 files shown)


waxlamp:
I think this is in conflict with #590. Shall I close that in favor of this?

dchiquito:
@waxlamp that was my plan, yes

web/src/main.ts
```diff
@@ -29,7 +29,15 @@ Sentry.init({
 
 sync(store, router);
 
-Promise.all([publishRest.restoreLogin(), girderRest.fetchUser()]).then(() => {
+async function loadUser() {
```
waxlamp, line 32: I don't think this needs to be `async`.

waxlamp (changes requested):

waxlamp (approved):

PR #602, opened 2021-03-26
Condense login and create account buttons into one for DJANGO_API feature flag

### LINKED ISSUE #594: Add toggle around Create Account button
This LGTM and I can't find anywhere in the code where `girderRest` is used when `DJANGO_API` is enabled _except_ right here:

https://github.com/dandi/dandiarchive/blob/12120b29a20a23991689bdb52d02753c0d62aa86/web/src/views/UserRegisterView/UserRegisterView.vue#L45

You can also see this by going to the gui beta depoyment register [page](https://gui-beta-dandiarchive-org.netlify.app/#/user/register) and looking at the network tab.

I suspect this is a bug, and should be changed to use dandi-api somehow. @dchiquito can you confirm this?

_Originally posted by @AlmightyYakob in https://github.com/dandi/dandiarchive/issues/590#issuecomment-795731031_

waxlamp:
Coauthored with @dchiquito.

@AlmightyYakob, another quick review please (thank you in advance!).

Closes #594.

### DIFF
--- a/web/src/components/AppBar/AppBar.vue
+++ b/web/src/components/AppBar/AppBar.vue
@@ -103,6 +103,7 @@
         color="primary"
         outlined
         rounded
+        v-if="!DJANGO_API"
       >
         Create Account
       </v-btn>
@@ -113,7 +114,7 @@
         rounded
         @click="login"
       >
-        Login
+        {{ DJANGO_API ? 'Login or Create Account' : 'Login' }}
       </v-btn>
     </template>
   </v-app-bar>


web/src/components/AppBar/AppBar.vue
```diff
@@ -113,7 +114,7 @@
         rounded
         @click="login"
       >
-        Login
+        {{ DJANGO_API ? 'Login or Create Account' : 'Login' }}
```
AlmightyYakob, line 117: While I can verify that this works, I'd just like to point out that ideally this would be two different buttons, since they do two different things. You could use a `<template v-if="!DJANGO_API">` above, to catch the create/login buttons for the girder end, and in the `v-else` have the single `Login or Create Account` button.

AlmightyYakob (approved):

PR #612, opened 2021-04-02
Testing if one can remove the hash without detrimental side effects

### LINKED ISSUE #446: Web UI (vue): switch from "hash" to "history" mode?
For the problem statement/goals, see #427.

I am no vue anything (not even a novice) so may be this is a wrong direction:  Quick search lead me to https://stackoverflow.com/questions/34623833/how-to-remove-hashbang-from-url which in turn to https://router.vuejs.org/guide/essentials/history-mode.html#example-server-configurations which suggests that there is a "history" mode which would make it possible to make URLs look normal (without placing all args into a fragment).

with switching away from using `hash` (fragment) mode in web ui we would be able to place `dandiarchive.org` behind reverse proxy (#427) to provide any necessary consistent mapping from permalinks to web ui and/or API end points without redirects.

satra:
The default vue mechanism uses `#` to prevent page reloads. Checking if this allows for removing it without other side effects. This is a test PR.

Closes #446

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -15,3 +15,8 @@ command = "yarn run build"
   VUE_APP_OAUTH_API_ROOT = "https://api-staging.dandiarchive.org/oauth/"
   VUE_APP_PUBLISH_API_ROOT = "https://api-staging.dandiarchive.org/api/"
   VUE_APP_SENTRY_ENVIRONMENT = "staging"
+
+[[redirects]]
+from = "/*"
+to = "/index.html"
+status = 200
--- a/src/router.js
+++ b/src/router.js
@@ -12,6 +12,7 @@ import CreateDandisetView from '@/views/CreateDandisetView/CreateDandisetView.vu
 Vue.use(Router);
 
 export default new Router({
+  mode: 'history',
   routes: [
     {
       path: '/',


satra:
@AlmightyYakob - it looked to me that this works. would love to know if there are other consequences of enabling this.

AlmightyYakob:
We cannot merge this, as it requires the server to have knowledge about the page structure. [Here](https://deploy-preview-612--gui-dandiarchive-org.netlify.app/dandiset/000005/draft) is an example of a failed page load within the GUI. Accessing any GUI URL besides the base will return `Page Not Found`. Relevant `vue-router` docs [here](https://router.vuejs.org/guide/essentials/history-mode.html#html5-history-mode).

AlmightyYakob:
There may be a way to configure Netlify to serve this correctly, I can push a change to this PR later to test that out

satra:
@AlmightyYakob - thanks. that makes sense the way the router behaves.

don't worry about this too much at this state. i was trying to serve it via the redirector service instead of netlify so that people don't continue to cite gui.dandiarchive.org in publications. 

the goal would be for the url to remain as `dandiarchive.org/dandiset/00000X`  in the browser and would therefore match directly with what we are getting from identifiers.org. and potentially since we run the redirector service, i could configure nginx to do the conditional rewrite. 

i'll leave this open for now, but consider it low priority relative to other ui issues.

yarikoptic:
Prior discussion(s) are available in https://github.com/dandi/dandiarchive/issues/446 which I would consider closed if this PR is merged.

satra:
sorry @yarikoptic - i should have searched more. as discussed above, this may be something we do with the redirector since we can run the reverse proxy there with configurability. but i also understand that changing this makes it dependent on the server, which would indeed be nice to avoid.

AlmightyYakob:
The change I just pushed seems to have fixed the issue, and now accessing a nested URL like https://deploy-preview-612--gui-dandiarchive-org.netlify.app/dandiset/000004 loads without issue. If having the URLs change does interest us, I think we should hold off merging until we're sure it has no consequences, or that they've all been accounted for.

satra:
nice! yes let's hold off on merging this one to we have settled decisions.

satra:
the PR looks good. i think what we will want to do is to fold the redirector and gui into a single DNS endpoint (just dandiarchive.org). we do need to figure out the mechanics of things, do the appropriate callback changes.  we will also temporarily want to redirect any request to `gui.dandiarchive.org/#/` to the new location. so we should perhaps make a list of things merging this PR would entail.

for merging this PR we would need.
- something will have to redirect the `gui.dandiarchive.org/#/` perhaps a rewrite 
- update CLI to support both with and without # urls

if we merge with dandiarchive.org, then CLI updates to reflect that.

waxlamp:
> the PR looks good. i think what we will want to do is to fold the redirector and gui into a single DNS endpoint (just dandiarchive.org). we do need to figure out the mechanics of things, do the appropriate callback changes. we will also temporarily want to redirect any request to `gui.dandiarchive.org/#/` to the new location. so we should perhaps make a list of things merging this PR would entail.
> 
> for merging this PR we would need.
> 
> * something will have to redirect the `gui.dandiarchive.org/#/` perhaps a rewrite
> * update CLI to support both with and without # urls
> 
> if we merge with dandiarchive.org, then CLI updates to reflect that.

I think this is complex enough a change that it deserves its own design doc. Would you mind launching one to discuss this in detail?

yarikoptic:
should we worry about providing continuity for `gui.dandiarchive.org/#/` urls?  if so - then I guess it should be a little JS within main page of `gui.dandiarchive.org` providing those redirects for currently known dandisets. 

dandi-cli: filed https://github.com/dandi/dandi-cli/issues/632

satra:
i don't know if that can be done at the js level to handle both. since this is doing a bit of rewrite at the netlify server level.

AlmightyYakob:
> My guess is, we can get Netlify to serve 301 responses for certain routes (but I am not an expert in this).

I can look into this when I have some time. Redirects are possible, but I'm not sure if we're able to sufficiently modify the path as a part of that redirect.

waxlamp:
I converted this to a draft PR since it's not ready for merging yet.

mvandenburgh:
Rebased after merging https://github.com/dandi/dandiarchive/pull/814

waxlamp:
@satra, should we continue to hang onto this PR? The scope of the initial experiment has grown quite a bit, and I wonder if closing this and relying on #446 and the discussion in this thread for a future effort would be prudent.

satra:
the scope is probably moving away from gui.dandiarchive.org as general goal, so indeed we can close this for now.

PR #640, opened 2021-04-23
Fix django api check for edit metadata permissions

### LINKED ISSUE #627: edit metadata shows up for non-owner datasets
i can click on edit metadata for non-owner datasets.

three options (in order of preference):
1. change to view metadata and make it completely readonly (i believe there is an option in the component to do this)
2. change to download metadata (and link to the metadata returned by the api)
3. disable this.

AlmightyYakob:
Fixes #627


web/src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -241,6 +249,23 @@ export default {
     ...mapGetters('dandiset', ['version']),
   },
   asyncComputed: {
+    userCanModifyDandiset: {
+      async get() {
+        if (this.user.admin) {
+          return true;
+        }
```
waxlamp, line 256: I wonder if we should disallow admins from editing metadata at this point. Do we need to have this backdoor access?

waxlamp (commented):
This looks good to me, and works in the beta deployment. Wait to hear from Daniel for any code issues.

web/src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -241,6 +249,23 @@ export default {
...
+
+        const { identifier } = this.publishDandiset.meta.dandiset;
+        const { data: owners } = await publishRest.owners(identifier);
+        const userExists = owners.find((owner) => owner.username === this.user.username);
+        if (userExists !== undefined) {
+          return true;
+        }
+
+        return false;
```
dchiquito, line 265: return !!
```suggestion
        return !!userExists;
```
I'm never sure if JS hacks like this really work, but it's shorter.

dchiquito (approved):

web/src/views/DandisetLandingView/DandisetMain.vue
dchiquito: I think it's useful for admins to be able to step in and tweak metadata if people have issues or questions.

waxlamp: You're right. Maybe we can add something in the UI that shows you only have access because you're an admin. That's not important right now and if I care enough about it I'll file an issue.

web/src/views/DandisetLandingView/DandisetMain.vue
waxlamp: Actually in this case I think the right thing to do is use [`Array.prototype.includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) directly, instead of `.find()`.

AlmightyYakob: `includes` only works when searching for a direct match in a list (i.e. you'd do something like `owners.includes(this.user)`. In this case, since we're trying to match objects, this will fail, since equivalent objects don't match with a simple `===` check.

waxlamp: Bless javascript's little heart.

PR #643, opened 2021-04-26
Add staging branch deployment

dchiquito:
`netlify.toml` assumes that `[context.staging]` means that you want another deployment on the branch `staging`, so we need to maintain a `staging` branch in this repo going forward for this deployment to be useful.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -9,3 +9,9 @@ command = "yarn run build"
   VUE_APP_OAUTH_CLIENT_ID = "Dk0zosgt1GAAKfN8LT4STJmLJXwMDPbYWYzfNtAl"
   VUE_APP_PUBLISH_API_ROOT = "https://api.dandiarchive.org/api/"
   VUE_APP_SENTRY_DSN = "https://425b9a012300493d867e97785fae7b88@o308436.ingest.sentry.io/5196549"
+
+[context.staging]
+
+  [context.staging.environment]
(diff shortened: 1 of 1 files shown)


netlify.toml
```diff
@@ -9,3 +9,9 @@ command = "yarn run build"
   VUE_APP_OAUTH_CLIENT_ID = "Dk0zosgt1GAAKfN8LT4STJmLJXwMDPbYWYzfNtAl"
   VUE_APP_PUBLISH_API_ROOT = "https://api.dandiarchive.org/api/"
   VUE_APP_SENTRY_DSN = "https://425b9a012300493d867e97785fae7b88@o308436.ingest.sentry.io/5196549"
+
+[context.staging]
+
+  [context.staging.environment]
+  VUE_APP_OAUTH_API_ROOT = "https://api-staging.dandiarchive.org/oauth/"
+  VUE_APP_PUBLISH_API_ROOT = "https://api-staging.dandiarchive.org/api/"
```
AlmightyYakob, line 17: From the netlify docs:

> any property in the [build] key will be applied to all contexts unless the same property key is present in a more specific context.

I wonder if that means we can remove the `[context.staging.environment]` block, and just retain the `[context.staging]` block to define this context?

dchiquito: That would define the `staging` context, but these URLs point to `api-staging` instead of `api`, so they are still required.

AlmightyYakob: Oh my bad, nevermind then.

AlmightyYakob (approved):

PR #653, opened 2021-05-05
Switch between "view" and "edit" metadata button

### LINKED ISSUE #647: a way to trigger "Edit metadata" interface in "read-only" (no Save) mode
For the purpose of establishing testing and timing of the web ui  we need to be able to trigger "Edit metadata" by dandibot user.  Ref:
I would feel insecure to grant that user full admin privileges.  
We need some way, even if ad-hoc/hardcoded to work only for dandibot,  on web ui end to enable that.

waxlamp:
This changes the behavior of the "edit metadata" button. Formerly, that button was available for owners of a dandiset, and admins, and was disabled for all others. Now, instead of disabling the button, it instead shows "view metadata", and leads to an instance of the meditor with the underlying component set to show all its data in a read only mode.

Closes #647

### DIFF
--- a/web/src/views/DandisetLandingView/DandisetLandingView.vue
+++ b/web/src/views/DandisetLandingView/DandisetLandingView.vue
@@ -8,7 +8,7 @@
         v-if="DJANGO_API"
         :schema="schema"
         :model="meta"
-        :readonly="readonly"
+        :readonly="!userCanModifyDandiset"
         @close="edit = false"
       />
       <meta-editor
@@ -70,7 +70,8 @@
             <DandisetMain
               :schema="schema"
               :meta="meta"
-              @edit="edit = true; readonly = $event"
+              :userCanModifyDandiset="userCanModifyDandiset"
+              @edit="edit = true"
             />
           </v-col>
           <v-col
@@ -91,6 +92,7 @@ import { mapState } from 'vuex';
 import DandisetSearchField from '@/components/DandisetSearchField.vue';
 import { draftVersion } from '@/utils/constants';
 import toggles from '@/featureToggle';
+import { user } from '@/rest';
 import MetaEditor from './MetaEditor.vue';
 import Meditor from './Meditor.vue';
 import DandisetMain from './DandisetMain.vue';
@@ -155,6 +157,22 @@ export default {
       dandisetVersions: (state) => state.versions,
       schema: (state) => state.schema,
     }),
+    user,
+  },
+  asyncComputed: {
+    userCanModifyDandiset: {
+      async get() {
+        if (this.user.admin) {
+          return true;
+        }
+
+        const { identifier } = this.publishDandiset.meta.dandiset;
+        const { data: owners } = await publishRest.owners(identifier);
+        const userExists = owners.find((owner) => owner.username === this.user.username);
+        return !!userExists;
+      },
+      default: false,
+    },
   },
   watch: {
     identifier: {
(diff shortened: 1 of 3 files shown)


satra:
nice!

yarikoptic:
> Partially addresses #647. @yarikoptic, I am not sure what your ask was in that issue--if this PR actually does fix that issue, I'll update this text to reflect that. Otherwise, I'll add what you need in a followup (since this PR offers a nice improvement per se).

As for the purpose of the #647 this PR would address it fully AFAIK: we will be able to trigger the meditor (even if only in read-only mode) which IMHO would be sufficient for webshotting of our live instance.

Can't judge vue/JS code, but netflify "preview" seems to work out nicely!  Couldn't test "login" since my attempt failed with "Error: invalid_request Mismatching redirect URI." but I do not think it is related to this PR changes even if intended to work at some point.

web/src/views/DandisetLandingView/DandisetLandingView.vue
```diff
@@ -69,7 +70,7 @@
             <DandisetMain
               :schema="schema"
               :meta="meta"
-              @edit="edit = true"
+              @edit="edit = true; readonly = $event"
```
AlmightyYakob, line 73: I don't like this, since it seems the only reason this is here is to initialize `readonly` to the correct value the first time the view/edit button is clicked. I think we should instead move the `userCanModifyDandiset` async computed prop up to this level (probably renaming it to something like `readonly`), and then pass it down equally to the `DandisetMain` and `Meditor` components.

web/src/views/DandisetLandingView/Meditor.vue
```diff
@@ -196,6 +193,10 @@ export default defineComponent({
       type: Object as PropType<DandiModel>,
       required: true,
     },
+    readonly: {
+      type: Boolean,
+      required: true,
```
AlmightyYakob, line 198: A default to `true` here might be nice

web/src/views/DandisetLandingView/Meditor.vue
```diff
@@ -228,6 +229,10 @@ export default defineComponent({
       return undefined;
     }
 
+    const options = computed(() => ({
```
AlmightyYakob, line 232: Why not keep the name `CommonVJSFOptions`? It'll make the intent and syntax below more clear

AlmightyYakob (changes requested):
Overall looks good just a few suggestions

web/src/views/DandisetLandingView/Meditor.vue
AlmightyYakob: No, you'd remove `required` and just have a default of `true`. Up to you but to me this prop reads like something that would have a default setting.

web/src/views/DandisetLandingView/DandisetLandingView.vue
waxlamp: See 1b88930f68b314e620fd17e90306f4512cf5bbc3.

web/src/views/DandisetLandingView/Meditor.vue
waxlamp: See a851331e97bc80dd715fa4b581f897236e2ee994.

web/src/views/DandisetLandingView/Meditor.vue
waxlamp: See 9681a44adad9ad2d4c0547a9250727552e1722b5.

web/src/views/DandisetLandingView/DandisetLandingView.vue
```diff
@@ -90,6 +92,7 @@ import { mapState } from 'vuex';
 import DandisetSearchField from '@/components/DandisetSearchField.vue';
 import { draftVersion } from '@/utils/constants';
 import toggles from '@/featureToggle';
+import { user } from '@/rest';
```
AlmightyYakob, line 95: To fix CI
```suggestion
import { publishRest, user } from '@/rest';
```

AlmightyYakob (approved):

PR #658, opened 2021-05-07
Add ID attributes to several UI elements

### LINKED ISSUE #648: Provide "id"s for actionable elements in the UI
ATM there is no easy and reliable way (besides an xpath like `'//*[@id="app"]/div/main/div/div/div/div/div[1]/div/div[2]/div[1]/div[3]/button[1]'` for "Edit metadata") to identify elements of the UI.  I think there should be an easy way to assign dedicated IDs per each such element (e.g. `edit_metadata_button`) so even if layout changes we have a robust way to find them (for testing).

waxlamp:
@yarikoptic, will these work?

@dchiquito, @AlmightyYakob: I just need one or the other of you to review these as a sanity check.

Closes #648


yarikoptic:
sorry, missed the request to review. confirming that there is `//*[@id="login"]` xpath for login button now. 
pushed http://github.com/dandi/dandi-api-webshots/commit/367255bc3924a58fc8fdb58708444ff0c88b94f0
and will run it now -- should work!  Thank you @waxlamp !

satra:
you should also be able to specifically use select elements by id instead of xpath, but i don't know which is more performant.

PR #660, opened 2021-05-07
Prevent the need for a double "back" to get to dandiset root

### LINKED ISSUE #573: DJANGO_API: view data - going back in browser history visits top directory twice
not sure how it happens but if you go to e.g. https://gui-beta-dandiarchive-org.netlify.app/#/dandiset/000008/draft/files and then to some directory, and press "back" you end up on https://gui-beta-dandiarchive-org.netlify.app/#/dandiset/000008/draft/files?location= and only after 2nd "back" on https://gui-beta-dandiarchive-org.netlify.app/#/dandiset/000008/draft/files. It makes navigation a bit odd especially when coming from the dandiset landing page

waxlamp:
This causes the "view data" button to link to a URL with a blank `location` query param. This prevents the need for clicking "back" twice to traverse past the version of the URL without that param in it.

@yarikoptic I am not sure if this truly solves the problem. Can you check via the Netlify preview?

Closes #573


yarikoptic:
> @yarikoptic I am not sure if this truly solves the problem. Can you check via the Netlify preview?

well, does it solve for you?  if so, likely would solve for me. (I do not think I am anyhow special)

ok, went out to test it out.  From the first look it looked like it did!  but then I am not stuck on https://deploy-preview-660--gui-dandiarchive-org.netlify.app/#/dandiset/000010/draft/files?location= to which I came from https://deploy-preview-660--gui-dandiarchive-org.netlify.app/#/dandiset/000010/draft/files?location=%2F somehow and thus pressing back just redirects me back and I am in this infinite loop without a way back .  So it seems has solved one issue but possibly introduced another.

seems reproduction is to jump from a folder up by clicking on '..', e.g. in 
https://deploy-preview-660--gui-dandiarchive-org.netlify.app/#/dandiset/000010/draft/files?location=sub-210861%2F
click on '..' which goes to `%2F` version which redirects (while recording itself in history) to the one without it, and thus I get stuck

AlmightyYakob:
Can confirm this reduces the number of required back navigations. @yarikoptic The bug you mentioned actually appears in master as well, and can be reproduced on https://gui.dandiarchive.org. One solution to this could be to use `router.replace` instead of the current [`router.push`](https://github.com/dandi/dandiarchive/blob/373f7087bad6bf9ac8693246848e1d6d9da4fb50/web/src/views/FileBrowserView/PublishFileBrowser.vue#L210-L213).

yarikoptic:
> Can confirm this reduces the number of required back navigations. @yarikoptic The bug you mentioned actually appears in master as well, and can be reproduced on https://gui.dandiarchive.org. One solution to this could be to use `router.replace` instead of the current [`router.push`](https://github.com/dandi/dandiarchive/blob/373f7087bad6bf9ac8693246848e1d6d9da4fb50/web/src/views/FileBrowserView/PublishFileBrowser.vue#L210-L213).

Thank you @AlmightyYakob !  I will leave to @waxlamp to decide to either fold the fix into this one or file a new issue or PR and proceed  with this one as is.

AlmightyYakob (approved):

PR #668, opened 2021-05-10
Upgrade all npm dependencies

brianhelba:


waxlamp (approved):
Approving these changes, trusting that the app works largely as it did before. If problems crop up, we'll fix them using our normal process.

web/package.json
```diff
@@ -8,51 +8,54 @@
...
-    "@vue/composition-api": "^1.0.0-beta.20",
-    "ajv": "^6.12.3",
-    "axios": "^0.19.2",
-    "core-js": "^3.6.5",
-    "filesize": "^6.1.0",
-    "js-yaml": "^3.14.0",
-    "lodash": "^4.17.19",
-    "moment": "^2.27.0",
+    "@apidevtools/json-schema-ref-parser": "^9.0.7",
+    "@girder/components": "~2.2.5",
```
brianhelba, line 12: `@girder/components@2.3` has breaking changes to the import structure, so pin this minor version too.

web/package.json
```diff
@@ -8,51 +8,54 @@
...
+    "lodash": "^4.17.21",
+    "moment": "^2.29.1",
     "vue": "^2.6.11",
-    "vue-json-pretty": "^1.6.5",
-    "vue-router": "^3.3.4",
-    "vuetify": "^2.3.6",
-    "vuex": "^3.5.1",
+    "vue-json-pretty": "^1.8.0",
+    "vue-router": "^3.5.1",
+    "vuetify": "~2.3.6",
```
brianhelba, line 28: `vuetify` minor releases have breaking layout changes which require migration, so pin the minor version.

web/package.json
```diff
@@ -8,51 +8,54 @@
...
+    "vuex": "^3.6.2",
     "vuex-router-sync": "^5.0.0"
   },
+  "resolutions": {
+    "@girder/components/vuetify": "~2.3.6"
```
brianhelba, line 33: `@girder/components` should probably be specifying `vuetify` as a `peerdependencies`. Since it doesn't, this pin will ensure that the top-level `vuetify` and its use nested within `@girder/components` are the same version.

PR #678, opened 2021-05-13
Fix file browser navigation bug

### LINKED ISSUE #666: Back button doesn't play well with .. links in dandiset file browser
@yarikoptic wrote:
> ok, went out to test it out.  From the first look it looked like it did!  but then I am now stuck on https://deploy-preview-660--gui-dandiarchive-org.netlify.app/#/dandiset/000010/draft/files?location= to which I came from https://deploy-preview-660--gui-dandiarchive-org.netlify.app/#/dandiset/000010/draft/files?location=%2F somehow and thus pressing back just redirects me back and I am in this infinite loop without a way back .  So it seems has solved one issue but possibly introduced another.
>
> seems reproduction is to jump from a folder up by clicking on '..', e.g. in 
https://deploy-preview-660--gui-dandiarchive-org.netlify.app/#/dandiset/000010/draft/files?location=sub-210861%2F
click on '..' which goes to `%2F` version which redirects (while recording itself in history) to the one without it, and thus I get stuck

_Originally posted by @yarikoptic in https://github.com/dandi/dandiarchive/issues/660#issuecomment-834819192_

@AlmightyYakob replied:
> The bug you mentioned actually appears in master as well, and can be reproduced on https://gui.dandiarchive.org. One solution to this could be to use `router.replace` instead of the current [`router.push`](https://github.com/dandi/dandiarchive/blob/373f7087bad6bf9ac8693246848e1d6d9da4fb50/web/src/views/FileBrowserView/PublishFileBrowser.vue#L210-L213).

AlmightyYakob:
Fixes #666

### DIFF
--- a/web/src/views/FileBrowserView/PublishFileBrowser.vue
+++ b/web/src/views/FileBrowserView/PublishFileBrowser.vue
@@ -206,7 +206,8 @@ export default {
     location(location) {
       const { location: existingLocation } = this.$route.query;
 
-      if (existingLocation && existingLocation === location) { return; }
+      // Update route when location changes
+      if (existingLocation === location) { return; }
       this.$router.push({
         ...this.$route,
         query: { location },
--- a/web/src/views/FileBrowserView/PublishFileBrowser.vue
+++ b/web/src/views/FileBrowserView/PublishFileBrowser.vue
@@ -251,7 +251,8 @@ export default {
 
       if (!folder) { return; }
       if (name === parentDirectory) {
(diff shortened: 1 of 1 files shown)


waxlamp (approved):

waxlamp:
@yarikoptic, can you check the netlify preview to see if this behaves as you expect?

@AlmightyYakob, feel free to merge and if @yarikoptic reports issues we can follow those up.

PR #683, opened 2021-05-18
Revert VJSF version bump

### LINKED ISSUE #681: meditor: cannot add dandiset contributors
in dandiset 105, the owner cannot add contributors. the dropdown for selecting person/organization is grayed out.

i looked at this in the codepen, where it seems to work fine, hence this may relate to how the UI is handling those additional buttons.

https://codepen.io/satra-the-reactor/pen/NWbKbrj?editors=1010

AlmightyYakob:
Fixes #681

If we want to bump the VJSF version we can do that in a separate PR (for example `v1.23.0` seems to work fine). However this will fix the immediate issue.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -11,7 +11,7 @@
     "@apidevtools/json-schema-ref-parser": "^9.0.7",
     "@girder/components": "~2.2.5",
     "@girder/oauth-client": "^0.7.7",
-    "@koumoul/vjsf": "1.25.0-beta.12",
+    "@koumoul/vjsf": "1.21.0",
     "@sentry/browser": "^6.3.6",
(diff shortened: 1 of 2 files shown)


dchiquito (approved):

satra:
@AlmightyYakob - thanks for this PR. did you also try 1.24.1?

AlmightyYakob:
> @AlmightyYakob - thanks for this PR. did you also try 1.24.1?

I believe so, and it seemed to still show some slight UI issues, so I decided to just stick with our previous version.

satra:
thanks @AlmightyYakob - not urgent, but we may want to keep a draft PR open with the recent version so that we can test what we need to fix.

PR #682, opened 2021-05-17
Citation and filepath fix

sanuann:
- added list of contributors above description
- todo: add citation button
- fixes #599, #645

Issue #645, opened 2021-04-30
Add "Cite as" button or pane


yarikoptic:
Following the lengthy discussion with @satra in slack, we ASAP need a prominently visible to a visitor guidance on how to cite any particulat dandiset.  I think it could be quickly achieved (even before `publish` functionality lands) by adding a  "Cite as" button next to "Get shareable link " which would open a similar pane with information such as 

"{contributors} {title}, {last-modified year}, https://dandiarchive.org/dandiset/{identifier}/{version}"

Later, whenever DOI-grade metadata records are established and DOIs minted, we should be able to produce citations in a selected number of styles and export (may be just redirect queries with proper headers to doi.org) in various styles for released dandisets.

satra:
old discussion here #470

satra:
with the new metadata structure this should be relatively easy to generate.

yarikoptic:
yeap yeap... thanks for digging the elderly issue out.
And now we got a dedicated issue for it!  FWIW, duplicating from comment in slack: https://pypi.org/project/citeproc-py/ might come handy if we decide to provide similar to zenodo pane as you suggested in that https://github.com/dandi/dandiarchive/issues/470#issuecomment-674569799

yarikoptic:
For all means and purposes #684 (thanks @sanuann @satra et al) addressed this issue

PR #684, opened 2021-05-19
Citation and filepath fix

### LINKED ISSUE #599: DJANGO_API: Double / / if have 3rd level directory
See current test ds https://gui-beta-dandiarchive-org.netlify.app/#/dandiset/010000/draft/files?location=sub-1%2Fsub-1_ses-10_cell-1.z%2F (goes a bit further than our 2 level `dandi organize`'ed datasets, as eg might happen later with zarr or any BIDS dataset with sessions)

image

satra:
Moving from #682 
- added list of contributors above description
- added citation link
- fixes #599, #645

@sanuann  - i created this branch in upstream hoping this will trigger a preview. @waxlamp - left a few comments on the previous PR which you should take a look at. you should be able to send updates into this branch.


satra:
@sanuann  - the previews now work.

the generator should skip any entity in contributor where "include_in_citation" is not true. also orcid does not make sense for any object with `schemaKey` == `Organization`.

sanuann:
this PR is ready to merge. Please review.

satra:
@sanuann - thanks here are some comments

image

1. remove orcid icons and semi-colons for the funding contributors.
2. display the full citation in cite-as, with a copy icon in the top right
3. remove contributors and replace with just funders and just the funding contributors.
4. while you are at it, remove citation, repository, url, schemaVersion, version from the list of items on the page.
5. one thing is to make the ID and the keywords show up in a nicer place. (see this mockup in this comment: https://github.com/dandi/dandiarchive/issues/470#issuecomment-674571606)

sanuann:
> @sanuann - thanks here are some comments
> 
> image
> 
> 1. remove orcid icons and semi-colons for the funding contributors.
> 2. display the full citation in cite-as, with a copy icon in the top right
> 3. remove contributors and replace with just funders and just the funding contributors.
> 4. while you are at it, remove citation, repository, url, schemaVersion, version from the list of items on the page.
> 5. one thing is to make the ID and the keywords show up in a nicer place. (see this mockup in this comment: [#470 (comment)](https://github.com/dandi/dandiarchive/issues/470#issuecomment-674571606))

How can I use (in my UI locally) the same dandiset you have used above in the screenshot?

sanuann:
> > How can I use (in my UI locally) the same dandiset you have used above in the screenshot?
> 
> download and upload just the dandiset metadata using the API.

which dandiset is this?

satra:
> which dandiset is this? 

that one is 4, but you can also test 26, which has many contributors as well.

web/src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -160,6 +221,7 @@ import {
 import toggles from '@/featureToggle';
 
 import CopyText from '@/components/CopyText.vue';
+import _ from 'lodash';
```
waxlamp, line 224: Can you check to see if you can extract just the functions you need to use? Something like

```typescript
import { isEmpty, map } from 'lodash';
```

I think this will help to keep the final bundle size small (but someone can correct me if I'm wrong).

web/src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -197,6 +259,19 @@ export default {
...
+    contributors() {
+      const authors = _.map(this.meta.contributor, (author) => {
+        let affiliations = '';
+        if (author.schemaKey !== 'Organization' && author.includeInCitation) {
+          if (!_.isEmpty(author.affiliation)) {
+            affiliations = _.map(author.affiliation, (a) => a.name);
+            affiliations = affiliations.join(', ');
+          }
+          return { name: author.name, identifier: `https://orcid.org/${author.identifier}`, affiliation: affiliations };
+        } return {};
```
waxlamp, line 271: For legibility, I would suggest:

```suggestion
        }
        
        return {};
```

Even better would be to maintain a return value variable that is modified within the if-block, but this is not a big deal.

waxlamp (changes requested):
This is looking pretty good!

I think Satra already mentioned it, but it would be better to display the full citation in the popover; that way people can see the whole thing and make sure it is what they think it should be.

cite-as

@jtomeck can help you figure out the right style settings to display this properly.

yarikoptic:
FWIW, looks great to me, besides I would have placed "Cite as" right after "Get shareable link" to the left group, instead of spreading it 
image

yarikoptic:
actually is it needed to keep it as input
image
and not just some `<p>` or `<span>` so it would be seen in its entirety and not cut off?

waxlamp:
> actually is it needed to keep it as input
> image
> and not just some `<p>` or `<span>` so it would be seen in its entirety and not cut off?

There must be a way to have both. @jtomeck can help to set up the styles properly so this looks the way we want (action button positioned closer to "get shareable link", and full citation appearing in a text box).

sanuann:
> @sanuann - thanks here are some comments
> 
> image
> 
> 1. remove orcid icons and semi-colons for the funding contributors.
> 2. display the full citation in cite-as, with a copy icon in the top right
> 3. remove contributors and replace with just funders and just the funding contributors.
> 4. while you are at it, remove citation, repository, url, schemaVersion, version from the list of items on the page.
> 5. one thing is to make the ID and the keywords show up in a nicer place. (see this mockup in this comment: [#470 (comment)](https://github.com/dandi/dandiarchive/issues/470#issuecomment-674571606))

- for point 3, what should the section heading be? Can it still be “Dandiset contributors” and then list only the funding contributors i.e. organizations only? or do I need to change the heading to something else?
- I am not clear about point 5. does it refer to the ID and Keyword section at the bottom?

sanuann:
> actually is it needed to keep it as input
> image
> and not just some `<p>` or `<span>` so it would be seen in its entirety and not cut off?

I actually did that on my end but wasn't sure what is the requirement..

satra:
> for point 3, what should the section heading be?

"Funding information" should involve organization names + award numbers

> does it refer to the ID and Keyword section at the bottom?

yes

sanuann:
> > does it refer to the ID and Keyword section at the bottom?
> 
> yes

with reference to the mockup image, what do the colors(grey for "MOp", green for "Intracellular") for the keywords mean? is color important?

satra:
> what do the colors(grey for "MOp", green for "Intracellular") 

i would just use a single color for now.

web/src/views/DandisetLandingView/DandisetMain.vue
AlmightyYakob: I was curious about this too, and it turns out there's [no difference](https://www.blazemeter.com/blog/the-correct-way-to-import-lodash-libraries-a-benchmark) between the two here. Where there is a difference is when you do something like:
```typescript
import map from 'lodash/map';
import isEmpty from 'lodash/isEmpty';
```

There also may be ways to achieve the same effect with the `babel-plugin-lodash` and `lodash-webpack-plugin` modules. However, these are out of scope for the current issue.

sanuann:
> > actually is it needed to keep it as input
> > image
> > and not just some `<p>` or `<span>` so it would be seen in its entirety and not cut off?
> 
> There must be a way to have both. @jtomeck can help to set up the styles properly so this looks the way we want (action button positioned closer to "get shareable link", and full citation appearing in a text box).

is there an update on what I should do here? will a `<p>` or `<span>` do?

jtomeck:
I moved the copy icon to the left and added logic to use a textarea for the cite as popover menu. This part will likely need a refactor eventually as the code is a little bit messy I think.

web/src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -111,7 +145,59 @@
...
+            <span v-else> {{ author.name }}</span>
+            ;
+          </span>
+        </v-col>
+      </v-row>
+      <v-row :class="titleClasses">
+        <v-card-title class="font-weight-regular">
+          Keyword(s)
+        </v-card-title>
+      </v-row>
```
satra, line 188: remove mention of keywords
```suggestion
      <v-row :class="titleClasses">
        <v-card-title class="font-weight-regular">
          Keyword(s)
        </v-card-title>
      </v-row>
```

web/src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -5,6 +5,14 @@
...
       </h1>
     </v-row>
+    <v-row
+      class="mx-2"
+      align="center"
+    >
+      <v-col cols="auto">
+        DANDI: {{ meta.identifier }}
+      </v-col>
+    </v-row>
```
satra, line 15: @sanuann - move this section to before the previous row and replace `DANDI: {{ meta.identifier }}` with just ` {{ meta.id }}`

sanuann:
@satra hope all is good now.

satra:
@sanuann - unfortunately the title change for the contributors changed the title in the meditor as well:

image

the title change should not persist in the data, just in the landing page.

sanuann:
> @sanuann - unfortunately the title change for the contributors changed the title in the meditor as well:
> 
> image
> 
> the title change should not persist in the data, just in the landing page.

ok, I'll fix that.

jtomeck:
> the box does not show it

Very interesting. I'll take a look real quick. The `auto-grow` prop should take care of this, so let me check that my implementation is correct.

jtomeck:
@satra I added overflow to the box so you can now access the full citation and scroll when it is too long.

satra:
@jtomeck - thanks. works for me. 

@waxlamp - this is ready for review/merge. would be nice to have this in.

web/src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -1,5 +1,13 @@
 <template>
   <div>
+    <v-row
+      class="mx-2"
+      align="center"
+    >
+      <v-col cols="auto">
+        {{ meta.id }}
+      </v-col>
+    </v-row>
```
waxlamp, line 10: Perhaps this section was added by accident? It results in the dandiset ID appearing above the main info panel.

```suggestion
```

waxlamp (changes requested):
This looks great; thanks for the deep effort, @sanuann!

I just have one suggestion to remove some code from the template; otherwise, looks nice!

web/src/views/DandisetLandingView/DandisetMain.vue
sanuann: no it wasn't an accident. It was suggested by satra.

waxlamp: @satra, could you explain what this is for?

satra: it is the full id of the dandiset (the metadata records have not all been updated to conform to the schema yet), but it is what it says, the id :)

satra: and yes, i wanted it there instead of as a row in the sections below.

waxlamp: If you want it this way, then the PR is cleared for merge. But just so you know, it feels a lot like the webpage broke or something (like when unstyled content sometimes leaks onto the styled part of a page). My spider sense is telling me that doing it this way is a serious knock to the general UX of the site. Is there a more organic way to integrate the ID value into the DLP that we could explore in a followup?

waxlamp (approved):

web/src/views/DandisetLandingView/DandisetMain.vue
satra: i see what you mean. yes it could look like this

**ID:** id

web/src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -1,5 +1,13 @@
 <template>
   <div>
+    <v-row
+      class="mx-2"
+      align="center"
+    >
+      <v-col cols="auto">
+        {{ meta.id }}
```
satra, line 8:
```suggestion
        ID: {{ meta.id }}
```

web/src/views/DandisetLandingView/DandisetMain.vue
satra: i made a simple update for now. we can merge this and then plan a better UX overall design.

waxlamp:
@sanuann, if you are happy with the state of this PR, please merge at your convenience.

PR #690, opened 2021-05-24
Remove "back" button from DLP

### LINKED ISSUE #689: Fix icon design for going "back" to public dandisets
_Originally posted by @satra in https://github.com/dandi/dandiarchive/issues/620#issuecomment-845481473_:
> `<--` is a weird icon for taking one back to public dandisets. 
> 
> i think that icon should be changed, but then there is duplication (we already have public dandisets on that page). so ideally it is back but takes you back to where you came from, which at the moment it doesn't. it takes you to public not the specific page of public dandisets you were looking at. 
> 
> overall it's a misleading UI i think.

waxlamp:
Closes #689.

### DIFF
--- a/web/src/views/DandisetLandingView/DandisetLandingView.vue
+++ b/web/src/views/DandisetLandingView/DandisetLandingView.vue
@@ -20,15 +20,7 @@
     </template>
     <template v-else>
       <v-toolbar class="grey darken-2 white--text">
-        <v-btn
-          icon
-          @click="navigateBack"
-        >
-          <v-icon color="white">
-            mdi-arrow-left
-          </v-icon>
-        </v-btn>
-        <v-toolbar-title>
+        <v-toolbar-title class="d-none d-md-block mx-8">
           Dandiset Dashboard
         </v-toolbar-title>
         <v-progress-circular
@@ -217,11 +209,6 @@ export default {
         },
       };
       this.$router.replace(route);
-    },
-    navigateBack() {
-      const route = this.$route.params.origin || { name: 'publicDandisets' };
-      this.$router.push(route);
-      this.$store.dispatch('dandiset/uninitializeDandisets');
     },
   },
 };


satra:
@waxlamp - while this is being cleaned up, i would address the following as well.

image

- the arrow on the right removes the details of the current dandiset (i'm not sure that is evident in any way for a user, without actually clicking it)
- the search bar is about general dandi search, but stuck right next to the arrow
- the title dandiset dashboard does not provide any additional info, so could be removed

one solution would be to spread the search bar across and remove the arrow from the gray bar to a hide/show sidepanel. 

openneuro has this on the left (e.g., https://openneuro.org/datasets/ds003655/versions/1.0.0):

image

web/src/views/DandisetLandingView/DandisetLandingView.vue
```diff
@@ -20,15 +20,7 @@
...
-        <v-btn
-          icon
-          @click="navigateBack"
-        >
-          <v-icon color="white">
-            mdi-arrow-left
-          </v-icon>
-        </v-btn>
-        <v-toolbar-title>
+        <v-toolbar-title class="d-none d-md-block mx-8">
```
AlmightyYakob, line 23: If we keep the grey bar, I think the margin here should be removed, as it looks awkward.
```suggestion
        <v-toolbar-title class="d-none d-md-block">
```

AlmightyYakob (approved):
LGTM, just one comment

web/src/views/DandisetLandingView/DandisetLandingView.vue
waxlamp: I agree that this looks better. The `mx-8` class is on the public dandisets page as well, and I was copying that to maintain consistent style.

I am going to change those on a separate PR so that Jared can advise us on the design impacts.

waxlamp:
> @waxlamp - while this is being cleaned up, i would address the following as well.
> 
> image
> 
> * the arrow on the right removes the details of the current dandiset (i'm not sure that is evident in any way for a user, without actually clicking it)
> * the search bar is about general dandi search, but stuck right next to the arrow
> * the title dandiset dashboard does not provide any additional info, so could be removed
> 
> one solution would be to spread the search bar across and remove the arrow from the gray bar to a hide/show sidepanel.
> 
> openneuro has this on the left (e.g., https://openneuro.org/datasets/ds003655/versions/1.0.0):
> 
> image

I'm going to address these concerns in a followup PR (coming soon). I don't like to mix features in the same PR mainly because it extends the discussion when we're already agreed on some parts.

web/src/views/DandisetLandingView/DandisetLandingView.vue
waxlamp: FYI, I included this fix with #703.

PR #695, opened 2021-05-24
Validation status

dchiquito:
Add `status` and `validation_error` to `Version` from the API.

Use `status` to disable the Publish button with an informative error if
the metadata is not valid.

Also rip out a snippet of old girder code.

This should actually be safe to merge into master now. If `status` is not present in the API response, none of the status checks pass, which results in the system behaving exactly as it does now (publish is allowed for all admins).

### DIFF
--- a/web/src/rest/publish.ts
+++ b/web/src/rest/publish.ts
@@ -12,7 +12,16 @@ const publishApiRoot = process.env.VUE_APP_PUBLISH_API_ROOT.endsWith('/')
 
 export function girderize(publishedDandiset: Version) {
   const { // eslint-disable-next-line camelcase
-    created, modified, dandiset, version, metadata, name, size, asset_count,
+    created,
+    modified,
+    dandiset,
+    version,
+    metadata,
+    name,
+    size,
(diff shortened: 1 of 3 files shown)


web/src/rest/publish.ts
```diff
@@ -22,6 +31,8 @@ export function girderize(publishedDandiset: Version) {
     lowerName: name,
     bytes: size,
     items: asset_count,
+    status,
+    validationError: validation_error,
```
AlmightyYakob, line 35: What's the point of updating `girderize`? Is it just to maintain consistency and avoid removing the function in this PR?

dchiquito: Removing `girderize` is definitely not happening in this PR since it would impact everything everywhere that uses versions, so we have to work with it.

There are two new fields, `status` and `validation_error` that are now being returned from the Django API. For those fields to actually appear in the girderized versions that are stored in Vuex, we need to change `girderize` to include those fields.

AlmightyYakob (approved):

PR #698, opened 2021-05-26
We are in 2021 now and some development is going on ;)

yarikoptic:


waxlamp (approved):
It might be better to include the full range of copyright years. It might also be better to omit the year entirely (though I don't have strong feelings about that).

Finally: who is `DANDI`? AFAIK there's no "legal person" known as DANDI (or is there?)

These are just questions--none of this should prevent a merge.

yarikoptic:
Thank you @waxlamp , merging

> These are just questions--none of this should prevent a merge.

filed https://github.com/dandi/dandi-infrastructure/issues/70 and https://github.com/dandi/handbook/issues/16

PR #699, opened 2021-05-26
fix: reporting TiB

### LINKED ISSUE #649: Tebi (TiB) vs Tera (TB) bytes
i think the UI is converting size to tebi bytes (2^40) but showing TB. we could divide by 10^12 and show TB. i just thing it would be good to be consistent between size calculation and units.

satra:
closes #649

### DIFF
--- a/web/src/components/DandiFooter.vue
+++ b/web/src/components/DandiFooter.vue
@@ -28,7 +28,7 @@ export default {
     <v-container>
       <v-row>
         <v-col offset="2">
-          &copy; 2020 DANDI<br>
+          &copy; 2021 DANDI<br>
           version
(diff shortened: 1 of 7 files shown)


satra:
@dchiquito or @AlmightyYakob - if you could review and merge this, that would be great :)

satra:
hold that thought - technically this should be fixed for all places where size is being reported (dandisets listing, dandiset metadata). @sanuann - can you take a stab at this?

satra:
@waxlamp - this is also ready for review/merge.

AlmightyYakob (approved):

satra:
thanks @AlmightyYakob

PR #703, opened 2021-05-27
Fix small UX issues

### LINKED ISSUE #702: Small UX deficiencies
> @waxlamp - while this is being cleaned up, i would address the following as well.
> 
> image
> 
> - the arrow on the right removes the details of the current dandiset (i'm not sure that is evident in any way for a user, without actually clicking it)
> - the search bar is about general dandi search, but stuck right next to the arrow
> - the title dandiset dashboard does not provide any additional info, so could be removed
> 
> one solution would be to spread the search bar across and remove the arrow from the gray bar to a hide/show sidepanel. 
> 
> openneuro has this on the left (e.g., https://openneuro.org/datasets/ds003655/versions/1.0.0):
> 
> image

_Originally posted by @satra in https://github.com/dandi/dandiarchive/issues/690#issuecomment-847179028_

waxlamp:
Closes #702 via the following changes:
- Remove the uninformative "Dandiset Dashboard" title
- Move the search bar to the left of the navbar
- Remove the open/close widget for the dandiset detail panel

I chose to simply remove the open/close control because the detail panel is unobtrusive and contains critical information about the dandiset. Until there's a positive need to hide it, I believe this is a better design solution in general.

@satra, if you could leave a review containing your feelings on whether/how well this solves the UX issues you raised, that would be great.

@jtomeck, please leave any design/Vuetify/CSS feedback you might have.

Thanks!

### DIFF
--- a/web/src/components/DandisetSearchField.vue
+++ b/web/src/components/DandisetSearchField.vue
@@ -1,5 +1,8 @@
 <template>
-  <v-form @submit="performSearch">
+  <v-form
+    width="100vh"
+    @submit="performSearch"
+  >
     <v-text-field
       :value="$route.query.search"
       label="Search Dandisets by name, description, identifier, or contributor name"
--- a/web/src/components/DandisetsPage.vue
+++ b/web/src/components/DandisetsPage.vue
@@ -1,7 +1,7 @@
 <template>
   <div v-page-title="pageTitle">
     <v-toolbar color="grey darken-2 white--text">
-      <v-toolbar-title class="d-none d-md-block mx-8">
+      <v-toolbar-title class="d-none d-md-block">
         {{ title }}
       </v-toolbar-title>
       <v-divider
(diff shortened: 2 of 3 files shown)


satra:
@waxlamp - this looks good. could the search bar fill up the gray space, just expands into the row? i like the arrow is not there anymore. the one fix that would probably need to happen is in a narrow view those details about the dandiset in the right pane move to the top of the view, instead they should move to the bottom, or middle.

waxlamp:
> @waxlamp - this looks good. could the search bar fill up the gray space, just expands into the row? i like the arrow is not there anymore. the one fix that would probably need to happen is in a narrow view those details about the dandiset in the right pane move to the top of the view, instead they should move to the bottom, or middle.

@satra, I think I have fixed both problems. Please take a look and let me know how it looks and feels to you.

satra (approved):
Looks great. I'm sure we will do a revamp of the landing page, but for the moment this is nice.

PR #711, opened 2021-06-02
remove logo for unidentified contributors

### LINKED ISSUE #707: ORCID logo being displayed for unidentified contributors
image

The above image would seem to imply that each of the three authors has an associated ORCID, however if you click on the symbol associated with "Inema, Ingrid", it takes you to orcid.org/unidentified. I think it would be better to omit the ORCID symbol for users that do not have an associated ORCID. This would be more in line with what the user expects and will help us identify where we are missing this data for specific contributors.

sanuann:
- closes #707
- removes contributor and keywords rows if no data present

### DIFF
--- a/web/src/views/DandisetLandingView/DandisetMain.vue
+++ b/web/src/views/DandisetLandingView/DandisetMain.vue
@@ -182,23 +182,23 @@
           </span>
         </v-col>
       </v-row>
-      <v-row
-        v-if="meta.keywords.length"
-        class="mx-2"
-      >
-        <v-col>
-          <span
-            v-for="key in meta.keywords"
-            :key="key"
-          >
-            <v-chip
-              small
-              style="margin: 5px;"
-              class="grey darken-2 font-weight-bold white--text"
-            > {{ key }} </v-chip>
-          </span>
-        </v-col>
-      </v-row>
+<!--      <v-row-->
+<!--        v-if="meta.keywords"-->
+<!--        class="mx-2"-->
+<!--      >-->
+<!--        <v-col>-->
+<!--          <span-->
+<!--            v-for="key in meta.keywords"-->
+<!--            :key="key"-->
+<!--          >-->
+<!--            <v-chip-->
+<!--              small-->
+<!--              style="margin: 5px;"-->
+<!--              class="grey darken-2 font-weight-bold white&#45;&#45;text"-->
+<!--            > {{ key }} </v-chip>-->
+<!--          </span>-->
+<!--        </v-col>-->
+<!--      </v-row>-->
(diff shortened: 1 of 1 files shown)


web/src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -5,7 +5,7 @@
       align="center"
     >
       <v-col cols="auto">
-        ID: {{ meta.id }}
```
satra, line 8: why is this being removed?

sanuann: sorry by mistake

satra:
this doesn't seem to be working. see dandiset 4 as an example. still several orcid logos pointing to undefined.

web/src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -180,20 +182,23 @@
...
-          >
-            <v-chip
-              small
-              style="margin: 5px;"
-              class="grey darken-2 font-weight-bold white--text"
-            > {{ key }} </v-chip>
-          </span>
-        </v-col>
-      </v-row>
+<!--      <v-row-->
```
satra, line 185: did you mean to comment this out?

sanuann: oops, was just testing something!

satra:
one the topic of identifiers, i suspect that name is being processed the same way independent of whether it's a person or organization: 

image

we should not put a `,` in the organization names.

sanuann:
> we should not put a `,` in the organization names.

The `,` appears the names are in the "Last name, First name" format. through the UI "edit metadata" option I don't see a place where you can put the last and first names separately. any idea where that comes from?

satra:
ah no worries then. we need to fix this at the metadata level.

satra:
@waxlamp - this is ready for review. the main issue is handled.

waxlamp (commented):
There is more in this PR than just the orcid badge fix--can you explain why the other changes are included?

sanuann:
> There is more in this PR than just the orcid badge fix--can you explain why the other changes are included?

to remove the rows if there is no list of contributors or keywords.

waxlamp:
> > There is more in this PR than just the orcid badge fix--can you explain why the other changes are included?
> 
> to remove the rows if there is no list of contributors or keywords.

Thanks for explaining. In general, it's good to keep PRs scoped as narrowly as possible, to make reviewing easier to and lessen the risk of a bug being introduced in any given PR.

waxlamp (approved):

PR #714, opened 2021-06-02
Add unsaved changes warning and undo functionality to meditor

### LINKED ISSUE #685: meditor: changed state and undo
we are asking for a lot of things that could be added using the meditor. i had someone enter some data and forgot to hit save, just hit the back button or something else. it would be nice if the meditor detected a change and prompted the user that they were navigating away without saving. 

on a more general note, some of the elements are complex and we don't really have an undo mechanism. an accidental delete of a complex object (e.g. person/organization) could be quite frustrating. perhaps we can look into how to confirm any delete operation on objects.

mvandenburgh:
Closes #685.

Adds
* Unsaved changes popup when attempting to exit meditor without saving
* undo/redo functionality to meditor (can be used via buttons in the UI or CTRL-Z/CTRL-Y)

Undo/redo is implemented using a transaction stack with a pointer to the current transaction. 
The stack has a maximum size limit to prevent the browsers memory from exploding (see `MAX_TRANSACTION_STACK_SIZE` in `src/utils/transactions.ts`). When the stack size reaches its limit, the element at the bottom of the stack is discarded whenever a new transaction occurs.

### DIFF
--- a/src/utils/schema/editor.ts
+++ b/src/utils/schema/editor.ts
@@ -69,6 +69,18 @@ class EditorInterface {
     return this.model;
   }
 
+  setBasicModel(newModel: DandiModel) {
+    Object.entries(newModel).forEach(([key, value]) => {
+      Vue.set(this.basicModel.value, key, value);
+    });
+  }
+
+  setComplexModel(newModel: DandiModel) {
+    Object.entries(newModel).forEach(([key, value]) => {
+      Vue.set(this.complexModel, key, value);
+    });
+  }
+
   setComplexModelProp(propKey: string, value: DandiModel) {
     Vue.set(this.complexModel, propKey, value);
   }
--- a/src/utils/transactions.ts
+++ b/src/utils/transactions.ts
@@ -128,15 +128,15 @@ export default class MeditorTransactionTracker {
     return this.transactions[this.transactionPointer].complex;
   }
 
-  areTransactionsAhead() {
+  areTransactionsAhead(): boolean {
     return this.transactionPointer < this.transactions.length - 1;
   }
 
-  areTransactionsBehind() {
+  areTransactionsBehind(): boolean {
     return this.transactionPointer > -1;
   }
 
-  isModified() {
+  isModified(): boolean {
     return this.transactionPointer > -1 || this.lostTransactions;
   }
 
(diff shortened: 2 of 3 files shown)


web/src/views/DandisetLandingView/Meditor.vue
```diff
@@ -240,6 +269,21 @@ export default defineComponent({
       store.commit('dandiset/setPublishDandiset', girderize(payload));
     }
 
+    const modified = ref(false);
+    const isModified = computed(() => modified.value);
```
AlmightyYakob, line 273: I think this computed variable is redundant, as it will always have the same value as `modified`. Since `modified` is a `ref`, it's also reactive. `computed` is usually used for creating reactive variables that need to modify some original data.

mvandenburgh: Oops, fixed that.

satra:
thanks @mvandenburgh - i will need to deploy locally to test (@sanuann - could you please take a look). @waxlamp - another reason to get staging back on track, with oauth support :) 

the one thing i noticed even with the non editable gui is that the undo/redo icons were active even in the view mode. i suspect this should be disabled in view, but also should depend on where the current state pointer is in the buffer. (e.g., being in any state + subsequent change should remove any buffer items beyond pointer and add a state to pointer + 1)

waxlamp:
This is nice!

I noted one UX difficulty: sometimes the browser native behavior of undo/redo (e.g., while editing a text input) conflicts with the application specific undo/redo introduced in this PR. It doesn't seem to always happen so I will demo it to you live and we can record findings here afterward.

mvandenburgh:
> I noted one UX difficulty: sometimes the browser native behavior of undo/redo (e.g., while editing a text input) conflicts with the application specific undo/redo introduced in this PR. It doesn't seem to always happen so I will demo it to you live and we can record findings here afterward.

I disabled the native browser undo/redo https://github.com/dandi/dandiarchive/pull/714/commits/72eb33728150c8cdad240a64d4d0d7ae149c2493. We lose out on being able to undo changes local to a single text box, but it doesn't seem like a good idea to try to mix the native undo/redo with a custom one.

waxlamp (changes requested):
Let's just disable the ctrl-z/y connection to the meditor undo/redo, and let the browser do its standard thing for those keyboard actions.

PR #715, opened 2021-06-02
fix: description per nih recommendation

satra:
Fix the description to match NIH feedback.


satra:
This was a simple enough change that I am merging.

PR #716, opened 2021-06-03
fix: change ID and title layout to match style

### LINKED ISSUE #705: Integrate "ID" display into DLP more smoothly
There's a sort of "floating" display of the dandiset ID above the main info panel in the DLP. We should find a more organic way to display this information.

satra:
This attempts to adjust the top of the dandiset details page to improve stylistic matching of the rest of the flow.

closes #705

### DIFF
--- a/web/src/views/DandisetLandingView/DandisetMain.vue
+++ b/web/src/views/DandisetLandingView/DandisetMain.vue
@@ -1,11 +1,12 @@
 <template>
   <div>
+    <v-card class="pb-2">
     <v-row
       class="mx-2"
       align="center"
     >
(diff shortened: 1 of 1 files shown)


waxlamp (approved):
This looks good, though:
1. It seems as though the indentation for the reshuffled template code is a bit off now, and
2. @jtomeck may have a cleaner way to express this change.

Neither of these is a showstopper, though the indentation should be easy enough to fix pre-merge. As for Jared looking at the code cleanliness, he is likely going to do a slow scan through the codebase refactoring some low-hanging fruit to be more in line with best practices, so that should not hold up a merge now.

PR #717, opened 2021-06-03
remove early access

satra:
per @yarikoptic


yarikoptic (approved):
It is ok with me to just remove it.  But it is kinda a cute location to give visitors some notion of a "heart beat" - that project is alive etc.  Anyways -- for that eventually we should come up with some "news" hopefully centralizing them with our currently stale ones on https://www.dandiarchive.org/

satra:
we need a revamp of www.dandiarchive.org in many ways. let's schedule the docathon for later this month. we can later reconfigure this in a more vue like fashion to produce both news and alerts.

PR #718, opened 2021-06-03
add vue gtag to do analytics

### LINKED ISSUE #477: Add analytics to vue app
It would be nice to add analytics to the gui.

we can use this tracker: `UA-146135810-2`

here is a vue package that seems to allow richer integration: https://matteo-gabriele.gitbook.io/vue-gtag/#add-plugin-to-your-application

and here is the standard html method, which as a first pass could be put in the footer:
```
<!-- Global site tag (gtag.js) - Google Analytics -->
<script async src="https://www.googletagmanager.com/gtag/js?id=UA-146135810-2"></script>
<script>
  window.dataLayer = window.dataLayer || [];
  function gtag(){dataLayer.push(arguments);}
  gtag('js', new Date());
  gtag('config', 'UA-146135810-2');
</script>
```

satra:
closes #477

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -23,6 +23,7 @@
     "lodash": "^4.17.21",
     "moment": "^2.29.1",
     "vue": "^2.6.11",
+    "vue-gtag": "^2.0.1",
     "vue-json-pretty": "^1.8.0",
     "vue-router": "^3.5.1",
     "vuetify": "~2.3.6",
--- a/web/package.json
+++ b/web/package.json
@@ -23,7 +23,7 @@
     "lodash": "^4.17.21",
     "moment": "^2.29.1",
     "vue": "^2.6.11",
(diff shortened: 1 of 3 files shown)


satra:
i can confirm this works on the google analytics dashboard.

waxlamp (commented):
This looks good. One question: will local builds cause spurious data to appear in the GA dashboard? Or is there some built-in way GA can distinguish the production instance from local builds?

satra:
> will local builds cause spurious data to appear in the GA dashboard?

i suspect it will try to send some info. i'm not too worried about a few testing or developer builds. i think overall this is more about global analytics, which should be much larger than local developer pings. i don't know if the library has some fanciness to it, but google itself might.

waxlamp (approved):

PR #722, opened 2021-06-04
Use the output of `git describe` as the version string

### LINKED ISSUE #151: Include deployed version (output of `git describe`) somewhere on the website
Ideally -- contingent on #150

ATM hard to figure out which version is running, not necessarily centrally (which should be the most recent) but in a local dockerized container; e.g. to troubleshoot https://github.com/dandi/dandi-cli/issues/51

waxlamp:
Closes #151

### DIFF
--- a/web/vue.config.js
+++ b/web/vue.config.js
@@ -1,6 +1,7 @@
 const MomentLocalesPlugin = require('moment-locales-webpack-plugin');
+const child_process = require('child_process');
 
-process.env.VUE_APP_VERSION = process.env.COMMIT_REF;
+process.env.VUE_APP_VERSION = child_process.execSync("git describe").toString();
 
 module.exports = {
   lintOnSave: false,
--- a/web/vue.config.js
+++ b/web/vue.config.js
@@ -1,7 +1,15 @@
 const MomentLocalesPlugin = require('moment-locales-webpack-plugin');
 const child_process = require('child_process');
 
-process.env.VUE_APP_VERSION = child_process.execSync("git describe").toString();
+function getVersion() {
+  try {
+    return child_process.execSync("git describe").toString();
+  } catch (err) {
+    return process.env.COMMIT_REF;
+  }
+}
+
+process.env.VUE_APP_VERSION = getVersion();
 
 module.exports = {
   lintOnSave: false,
--- a/web/src/components/DandiFooter.vue
+++ b/web/src/components/DandiFooter.vue
@@ -40,7 +40,7 @@ export default {
                 class="version-link"
                 v-on="on"
                 @click="versionClick"
-              >{{ version.slice(0, 6) }}</a>
+              >{{ version }}</a>
             </template>
             <span>Copied to clipboard!</span>
           </v-tooltip>


waxlamp:
@brianhelba, are there any pitfalls to accomplishing this feature this way?

web/vue.config.js
```diff
@@ -1,6 +1,7 @@
 const MomentLocalesPlugin = require('moment-locales-webpack-plugin');
+const child_process = require('child_process');
 
-process.env.VUE_APP_VERSION = process.env.COMMIT_REF;
+process.env.VUE_APP_VERSION = child_process.execSync("git describe").toString();
```
satra, line 4: i think for dandiarchive is fine, but this would mean any current or future deployment would need to have git installed locally and that we are cloning rather than say downloading a zip file and building. 

@yarikoptic - as an example, in heroku for api there is no git clone (not even a possibility). we looked into it there. the standard django deployment simply downloads the latest zip file.

waxlamp: This is a good point.

The fix is to build in error detection and fall back to the git sha. I will work on this in a bit.

waxlamp: As far as dandi-api is concerned, there my objections to attaching a semver don't really exist, and if we want to buy into it, something like semantic-release could be very useful, and would obviate the need to scrounge up git shas to display in the API.

yarikoptic: FWIW -- I think `git describe`, which does include git shas past the tag/release point, would always be useful as long as there is a chance to get a "devel" version running. Yes -- comes at a cost of having/needing git to be installed.  
If we ever to provide a proper "distribution" of the dandiarchive etc, then static version string could be minted upon generation of the distirbution.  That is what e.g. `versionner` does for source tarballs for distribution/upload to pypi, making it not needed to have `git` upon installation for that "static-ly installed" version.

waxlamp: > The fix is to build in error detection and fall back to the git sha. I will work on this in a bit.

This is done now (see ac94847).

BTW, this does not completely solve the problem of how to obtain a version number for the case when this is being built neither locally, nor via Netlify. I think that would require a bigger discussion and a design doc.

But for now I believe this is ready for another review.

yarikoptic (approved):
For the original intent -- looks great in the preview!  Some stylists might like to adjust but good for me.

web/vue.config.js
```diff
@@ -1,6 +1,15 @@
 const MomentLocalesPlugin = require('moment-locales-webpack-plugin');
+const child_process = require('child_process');
 
-process.env.VUE_APP_VERSION = process.env.COMMIT_REF;
+function getVersion() {
+  try {
+    return child_process.execSync("git describe").toString();
+  } catch (err) {
+    return process.env.COMMIT_REF;
```
brianhelba, line 8: It'd be good to have a code comment explaining which environments this fallback is defined in.

web/vue.config.js
```diff
@@ -1,6 +1,15 @@
...
+  }
+}
+
+process.env.VUE_APP_VERSION = getVersion();
```
brianhelba, line 12: If `git describe` and `COMMIT_REF` both fail to produce a value, than the value of `process.env.VUE_APP_VERSION` will be `'undefined'` (a string), since [any values assigned to `process.env` are coerced as such](https://nodejs.org/dist/latest-v16.x/docs/api/process.html#process_process_env).

Maybe that string is exactly what you'd like to render on the footer component, but since this is implicit, I wanted to point it out.

web/vue.config.js
```diff
@@ -1,6 +1,15 @@
 const MomentLocalesPlugin = require('moment-locales-webpack-plugin');
+const child_process = require('child_process');
 
-process.env.VUE_APP_VERSION = process.env.COMMIT_REF;
+function getVersion() {
+  try {
+    return child_process.execSync("git describe").toString();
```
brianhelba, line 6: There's only one tag in the repo now (`0.1.0`) and it's annotated. If we anticipate ever adding additional [non-annotated tags](https://stackoverflow.com/a/11514139), then this should use `git describe --tags`. If all future tags will be annotated, then this is fine as-is.

brianhelba (commented):
This seems fine as-is for now, but we should note a few contingent cases.

PR #725, opened 2021-06-04
Remove obsolete Dockerfiles

waxlamp:
These support the Girder 3 plugin. This also removes "client.Dockerfile",
but that's now much simpler to run natively for development.

(The work in this PR was done by @brianhelba)


yarikoptic:
the cleaner the better!  @jwodder -- we aren't relying on any of those (i.e. images built from them) somehow for dandi-cli testing, do we?

jwodder:
@yarikoptic No, we're not using them any more.

waxlamp:
@brianhelba, can you verify that this is still the right way to go?

brianhelba (approved):

PR #727, opened 2021-06-04
Fix API key not refreshing when refresh icon is clicked

### LINKED ISSUE #622: Token does not refresh when refresh icon is clicked
the refresh functionality on the current or future ui is not working. it does not generate a new token with the appropriate scopes. i think we have run into it in a couple of cases. 

i also checked on the gui-beta and it didn't refresh either when i hit refresh.

i think tokens should refresh whenever a user wants them to.

mvandenburgh:
Fixes #622; the refresh button in the app bar actually retrieves a new API key for the user now.


waxlamp (changes requested):
Discussed IRL with Mike, but just to summarize, here's what we'll do before merging this:
- split the `fetch()` function into two different functions, one to unconditionally create an API key, and another to get that if it exists, and create and return it if it does not
- remove the feature toggle code paths

This will avert piling different functionality into one function (and eliminate the code repetition as well as the use of a boolean flag to control the function's behavior).

waxlamp (approved):
Ahhh looks great!

Issue #694, opened 2021-05-24
Investigate VJSF performance issues


AlmightyYakob:
This is an investigative task to help understand the root cause of #662, #646 and #588.

yarikoptic:
FWIW, I did profile in browser today but since know no JS wasn't able to identify ultimate reason

yarikoptic:
If there is no related issue filled with vjsf, I would've also failed one there, ideally with some code to reproduce/demonstrate

mvandenburgh:
We discussed this in person yesterday, but just to summarize here:

The slowdown appears to occur when VJSF renders an "array" field from a JSON schema. When an array field contains many elements that contain several nested child objects, the rendering slows down (an example is the "contributor" field on [this dandiset](https://gui.dandiarchive.org/#/dandiset/000004)). I'll file an issue with vjsf with that dandiset as an example.

satra:
this should also improve with dandi schema  `0.4.0` which simplifies affiliation  (and i might reduced the contact person model as well to a simpler form in a future schema update).

waxlamp:
This investigation is concluded, with a fix in #730 waiting on a schema upgrade.

mvandenburgh:
The slowdown still persists even after applying the fix in #730 - reopening this issue.

satra:
@mvandenburgh - i've updated this codepen to use dandiset 26  (https://codepen.io/satra-the-reactor/pen/NWbKbrj?editors=1010) perhaps you can use this as an example with vjsf issue to see what could be done to speed things up.

satra:
that being said, other dandisets are quite a bit faster now as a result of the change. and even this dropped by about 10s

mvandenburgh:
Keeping this open until the cause of https://gui-staging.dandiarchive.org/#/dandiset/000020 freezing is determined.

waxlamp:
> Keeping this open until the cause of https://gui-staging.dandiarchive.org/#/dandiset/000020 freezing is determined.

@mvandenburgh is this still valid or can we close? Dandiset 20 doesn't freeze the meditor (but maybe the dandiset in question has moved or something), and we are now moving onto a bit of an alternative metadata editing scheme.

mvandenburgh:
> @mvandenburgh is this still valid or can we close? Dandiset 20 doesn't freeze the meditor (but maybe the dandiset in question has moved or something), and we are now moving onto a bit of an alternative metadata editing scheme.

[...]

PR #729, opened 2021-06-08
Load complex schema's VJsf form lazily in meditor

mvandenburgh:
This is a quick/temporary fix for the performance issues mentioned in #694. Since the slowdown is caused by the "complex schema" VJsf forms (i.e. any form that is opened as a pop up, such as "Dandiset Contributors", "Subject Matter", etc), this PR modifies these to be loaded only when the user clicks the appropriate button.

For example, [this dandiset](https://gui.dandiarchive.org/#/dandiset/000004) would load immediately when the Meditor is opened and will only lag when the "Dandiset Contributors" form is opened (see https://github.com/dandi/dandiarchive/issues/694#issuecomment-856860270)

### DIFF
--- a/web/src/views/DandisetLandingView/Meditor.vue
+++ b/web/src/views/DandisetLandingView/Meditor.vue
@@ -121,12 +121,14 @@
               :ref="`${propKey}-form`"
               v-model="complexModelValidation[propKey]"
             >
-              <v-jsf
-                :value="complexModel[propKey]"
-                :schema="complexSchema.properties[propKey]"
-                :options="CommonVJSFOptions"
-                @input="setComplexModelProp(propKey, $event)"
-              />
+              <v-lazy>
+                <v-jsf
+                  :value="complexModel[propKey]"
+                  :schema="complexSchema.properties[propKey]"
+                  :options="CommonVJSFOptions"
+                  @input="setComplexModelProp(propKey, $event)"
+                />
+              </v-lazy>
             </v-form>
           </v-card>
         </v-dialog>


waxlamp (approved):

AlmightyYakob:
Just want to point out the inadvertent issue this causes, which is the validation indicator at the top of the meditor. Since the complex vjsf forms aren't being loaded, that indicator will show as valid, even if the underlying data is invalid/missing. Once the respective form is loaded, it will then update to show invalid. @mvandenburgh if you're already tracking this then I'll leave it alone, but if not I can file an issue to track this.

waxlamp:
> Just want to point out the inadvertent issue this causes, which is the validation indicator at the top of the meditor. Since the complex vjsf forms aren't being loaded, that indicator will show as valid, even if the underlying data is invalid/missing. Once the respective form is loaded, it will then update to show invalid. @mvandenburgh if you're already tracking this then I'll leave it alone, but if not I can file an issue to track this.

Good catch. This may become moot once the new VJSF version is deployed.

mvandenburgh:
> Just want to point out the inadvertent issue this causes, which is the validation indicator at the top of the meditor. Since the complex vjsf forms aren't being loaded, that indicator will show as valid, even if the underlying data is invalid/missing. Once the respective form is loaded, it will then update to show invalid. @mvandenburgh if you're already tracking this then I'll leave it alone, but if not I can file an issue to track this.

I actually didn't catch that. This was implemented before we were made aware that the new version of VJSF fixes the performance problems, so as @waxlamp said this isn't needed anymore. I'll file a PR to revert this.

This also might be what is blocking #714 (not the VJSF upgrade), I'll take a look at that as well.

PR #732, opened 2021-06-10
Sort by "Identifier" and "Modified"

### LINKED ISSUE #567: Web frontend: Add sort by identifier
currently there is no way to list the dandisets by identifier. originally this would have matched created in the girder server, but not in the new API.

mvandenburgh:
Closes #567.
There will need to be some small changes made to the api server as well (so this PR isn't actually functional yet locally or in the deploy preview), but first I wanted to figure out exactly what sorting options we want to add/remove. I went with @yarikoptic's suggestion in #567 to remove the `Created` sorting option and instead add in `Identifier` and `Modified` options. Either way it seems redundant to have both `Identifier` and `Created` options.

This PR also removes all the DJANGO_API feature toggle code in `DandisetsPage.vue`.

TODO:
- Merge https://github.com/dandi/dandi-api/pull/327

### DIFF
--- a/web/src/components/DandisetsPage.vue
+++ b/web/src/components/DandisetsPage.vue
@@ -85,6 +85,10 @@ const sortingOptions = [
     name: 'Name',
     djangoField: 'name',
   },
+  {
+    name: 'Identifier',
+    djangoField: 'id',
+  },
 ];
 
 export default defineComponent({
--- a/web/src/components/DandisetsPage.vue
+++ b/web/src/components/DandisetsPage.vue
@@ -82,6 +82,10 @@ const sortingOptions = [
     djangoField: 'created',
   },
   {
+    name: 'Modified',
+    djangoField: 'modified',
+  },
+  {
     name: 'Name',
     djangoField: 'name',
   },
--- a/web/src/components/DandisetsPage.vue
+++ b/web/src/components/DandisetsPage.vue
@@ -78,20 +78,16 @@ const DANDISETS_PER_PAGE = 8;
 
 const sortingOptions = [
   {
-    name: 'Created',
-    djangoField: 'created',
-  },
-  {
-    name: 'Modified',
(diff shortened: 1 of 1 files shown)


waxlamp (commented):
Code looks good. Withholding approval until the API server changes are set.

mvandenburgh:
I opened a PR for the server changes that are needed (https://github.com/dandi/dandi-api/pull/327).

There is still an open question of what sorting by "modified" means exactly for dandisets. There are two `modified` fields at play, namely those on the `VersionMetadata` and `AssetMetadata`. 
Since each `Version` can have multiple assets, including `AssetMetadata`'s modified field in the sorting would require searching the metadata of every asset associated with each dandiset, which could be a performance concern if a dandiset has a large number of assets. As far as I can tell the same problem does not apply for `VersionMetadata`, since each `Version` can only have one associated `VersionMetadata`.

satra:
i think this should not be a computed thing, but something that is updated in `Version` when either `VersionMetadata` or an `AssetMetadata` in that version is modified. this would make it not search all assets, but rather have assets trigger the update whenever they are changed.

yarikoptic:
FWIW: I think a DB operation to take max or min across dates should be negligibly fast, ie. under a ms, (unlike e.g. getting AssetStats updated) even for dandisets with thousands of assets and then would guaranteed to be "up to date".

mvandenburgh:
> i think this should not be a computed thing, but something that is updated in `Version` when either `VersionMetadata` or an `AssetMetadata` in that version is modified. this would make it not search all assets, but rather have assets trigger the update whenever they are changed.

It turns out that what you're suggesting is already implemented (thanks @dchiquito for pointing that out), so this PR should be all set.

dchiquito (approved):

waxlamp (approved):

PR #736, opened 2021-06-14
Adjust main netlify.toml

dchiquito:
This is the netlify.toml version that will live in `master`.


waxlamp (commented):
This is a cleaner expression of our intent: production does it this way; override like this for deploy previews.

Thanks!

PR #738, opened 2021-06-14
Revert lazy loading of complex schema forms in Meditor

### LINKED ISSUE #737: Meditor validation UI bug
The change made in #729 introduced a bug where the validation indicator in the Meditor UI may not show correctly (see https://github.com/dandi/dandiarchive/pull/729#issuecomment-861001847). The change can be reverted regardless, since the performance issues it was introduced to mitigate have been solved by upgrading VJSF.

mvandenburgh:
Reverts #729 and closes #737.

This will cause the slowdown to occur again until the migration to schema v0.4.1 is complete, but I think it's better to have slowdown then broken functionality.

### DIFF
--- a/web/src/views/DandisetLandingView/Meditor.vue
+++ b/web/src/views/DandisetLandingView/Meditor.vue
@@ -121,14 +121,12 @@
               :ref="`${propKey}-form`"
               v-model="complexModelValidation[propKey]"
             >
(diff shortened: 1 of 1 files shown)


AlmightyYakob:
LGTM

AlmightyYakob (approved):
LGTM

PR #744, opened 2021-06-16
Fix blank screen displaying if oauth token is invalid

### LINKED ISSUE #740: Blank screen if OAuth token is invalid
The `/api/users/me` endpoint responds with a `401 Unauthorized` response if the client's OAuth `access_token` is invalid. The client does not properly catch this error, and instead just shows a blank white screen; right now the only way to get the UI to appear again is to manually clear the cookie from the browser.

mvandenburgh:
Also snips out some more DJANGO_API feature toggle code.

Closes #740.

### DIFF
--- a/web/src/main.ts
+++ b/web/src/main.ts
@@ -14,7 +14,6 @@ import '@/plugins/composition';
 import '@/plugins/girder';
 
 // Import custom behavior
-import toggles from '@/featureToggle';
 import '@/title';
 
 // Import internal items
--- a/web/src/main.ts
+++ b/web/src/main.ts
@@ -32,14 +32,20 @@ Sentry.init({
 sync(store, router);
 
 Vue.use(VueGtag, {
-  config: { id: "UA-146135810-2" }
+  config: { id: 'UA-146135810-2' },
 }, router);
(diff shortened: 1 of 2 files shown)


AlmightyYakob:
I've got a couple opinions on this,

1. I feel like this would be better placed in `publishRest.restoreLogin`. Since that's where the error originates from, it'd be a good place to contain the error conditions
2. We should logout the user if this condition is reached. Since currently no logout is performed, the request will continue to fail until the user logs in again.

The following is what I had in mind:
```typescript
    async restoreLogin() {
      await oauthClient.maybeRestoreLogin();
      if (!oauthClient.isLoggedIn) {
        return;
      }

      try {
        // Fetch user
        this.user = await this.me();
      } catch (e) {
        // A status of 401 indicates login failed, so the exception should be supressed.
        if (e.response.status === 401) {
          await oauthClient.logout();
        } else {
          // Any other kind of exception indicates an error that shouldn't occur
          throw e;
        }
      }
    },
```

@mvandenburgh what do you think?

AlmightyYakob (approved):

PR #746, opened 2021-06-17
Fix basic form bug in meditor

### LINKED ISSUE #743: VJSF 2.0.2 breaks "basicModel" forms in Meditor
The VJSF upgrade in #730 that fixes the freezing problems seems to introduce a new bug in the Meditor. Any edits to the "basicModel" forms (i.e. any VJSF form that is present on the main Meditor screen like `Dandiset title`, `Description`, `Study Target`, `License`, etc.) are not persisted after clicking the save button. "complexForms" still work.

GIF demo of the bug, note that changes are saved when the staging GUI is used (with VJSF v1), but not when the Netlify branch preview for #730 is used (with VJSF v2):
meditor bug\

mvandenburgh:
Fixes #743. VJSF wasn't detecting changes to the models wrapped with `reactive()`, but it does work with `ref()`.

### DIFF
--- a/web/src/utils/schema/editor.ts
+++ b/web/src/utils/schema/editor.ts
@@ -2,7 +2,7 @@ import type { JSONSchema7 } from 'json-schema';
 
 import Vue from 'vue';
 import {
-  computed, reactive, ref, ComputedRef,
+  computed, reactive, ref, ComputedRef, Ref,
 } from '@vue/composition-api';
 import { cloneDeep } from 'lodash';
 
@@ -23,8 +23,8 @@ class EditorInterface {
   // Not guaranteed to be up to date, use getModel()
   private model: DandiModel;
 
-  basicModel: DandiModel;
-  complexModel: DandiModel;
+  basicModel: Ref<DandiModel>;
+  complexModel: Ref<DandiModel>;
 
   schema: JSONSchema7;
   basicSchema: JSONSchema7;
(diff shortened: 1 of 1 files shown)


AlmightyYakob:
If it really is just `basicModel` that is affected by the VJSF version change, I'd encourage to only change `basicModel` to use `ref`, as a means of investigation. If both `basicModel` and `complexModel` really should use `ref`, then I'm all for it, but I'd like to know where exactly the issue is (if there's bandwidth for it).

mvandenburgh:
> If it really is just `basicModel` that is affected by the VJSF version change, I'd encourage to only change `basicModel` to use `ref`, as a means of investigation. If both `basicModel` and `complexModel` really should use `ref`, then I'm all for it, but I'd like to know where exactly the issue is (if there's bandwidth for it).

Nope, `complexModel` seems to work fine with `reactive` so I reverted it.

VJSF 2.0 changed the way `v-model` works, which could be related to it https://koumoul-dev.github.io/vuetify-jsonschema-form/latest/1to2.

AlmightyYakob:
> VJSF 2.0 changed the way `v-model` works, which could be related to it https://koumoul-dev.github.io/vuetify-jsonschema-form/latest/1to2.

Thanks for pointing this out, it seems like the update to reactivity is exactly as we thought. Namely:
> better implementation of v-model by not mutating the value parameter and instead creating clones for each input event (better for clarity and side-effects)

This would explain why `complexModel` works as a `reactive` object, but `basicModel` doesn't.

PR #754, opened 2021-06-18
Link out to github instead of copy version string

### LINKED ISSUE #726: Make click on version to not copy to clipboard but go to github
Reviewing #722 an idea came to mind

https://github.com/dandi/dandiarchive/tree/0.1.0-421-g609c077 works just fine ;)  so URL could be easily constructed from the exact commit or current `git describe` which is sweet.

copy pasting could be done via regular select/copy, but getting to github requires "knowledge"

waxlamp:
Closes #726

### DIFF
--- a/web/src/components/DandiFooter.vue
+++ b/web/src/components/DandiFooter.vue
@@ -12,6 +12,15 @@ export default {
     version() {
       return process.env.VUE_APP_VERSION;
     },
+
+    githubLink() {
+      const gitRev = process.env.VUE_APP_GIT_REVISION;
+      if (gitRev) {
+        return `https://github.com/dandi/dandiarchive/commit/${gitRev}`;
+      } else {
+        return "https://github.com/dandi/dandiarchive";
+      }
+    },
   },
   methods: {
     versionClick() {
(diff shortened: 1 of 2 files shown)


web/src/components/DandiFooter.vue
```diff
@@ -12,6 +12,15 @@ export default {
...
       return process.env.VUE_APP_VERSION;
     },
+
+    githubLink() {
+      const gitRev = process.env.VUE_APP_GIT_REVISION;
+      if (gitRev) {
+        return `https://github.com/dandi/dandiarchive/commit/${gitRev}`;
+      } else {
+        return "https://github.com/dandi/dandiarchive";
+      }
```
yarikoptic, line 22: duplication allergy in me is really unhealthy. Isn't there some syntax support for smth like

```suggestion
      return `https://github.com/dandi/dandiarchive` + (${gitRev} ? `/commit/${gitRev}` : ``);
```

but totally feel to ignore my allergic reaction here -- it is a url/code which is not to change etc

yarikoptic (approved):
works nicely!

web/src/components/DandiFooter.vue
waxlamp: Even better would be something like

```javascript
      return `https://github.com/dandi/dandiarchive${gitRev ? `/commit/${gitRev}` : ""}`;
```

but despite your allergy to duplication (which I share a mild version of, truth be told), I find our condensed versions far less legible than what I have here with the separated `return` statements.

PR #756, opened 2021-06-22
Sort by "modified" by default instead of name

### LINKED ISSUE #734: Make "Modified" (or "Identifier") to be the default sorting
ATM default sorting is by name, which might cause people to "abuse" that to have their dataset appear first.  Overall also sorting by Name is not as useful as by Modified or Identifier.  I think making web ui sort by Modified (or Identifier) would be more "fair" and informative IMHO.

mvandenburgh:
Makes `modified` the default way to sort dandisets instead of `name`.

Closes #734.

### DIFF
--- a/web/src/components/DandisetsPage.vue
+++ b/web/src/components/DandisetsPage.vue
@@ -78,16 +78,16 @@ const DANDISETS_PER_PAGE = 8;
 
 const sortingOptions = [
   {
-    name: 'Name',
-    djangoField: 'name',
+    name: 'Modified',
+    djangoField: 'modified',
   },
   {
     name: 'Identifier',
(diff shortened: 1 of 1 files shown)


mvandenburgh:
@waxlamp I was going to exercise my new self-review powers on this PR but it's still blocking me from merging until I get an approving review :thinking:

waxlamp:
Added you to the administrator team; try again now.

mvandenburgh:
> Added you to the administrator team; try again now.

That worked, thanks.

yarikoptic:
Thank you @mvandenburgh . I already love it being the default - immediately gives an idea on the liveliness of the archive.

PR #766, opened 2021-06-25
Make publish validation error hover wrap lines

### LINKED ISSUE #762: Make validation error hover wrap lines
Currently if a validation error has multiple lines, it is rendered with no line breaks, which makes it much less legible.

mvandenburgh:
Makes validation errors that occur when hovering over the "Publish" button wrap to the next line when appropriate. Also remove all DJANGO_API feature toggle code from `DandisetMain.vue`.

Closes #762.

### DIFF
--- a/web/src/views/DandisetLandingView/DandisetMain.vue
+++ b/web/src/views/DandisetLandingView/DandisetMain.vue
@@ -1,21 +1,21 @@
 <template>
   <div>
     <v-card class="pb-2">
-    <v-row
-      class="mx-2 font-weight-regular"
-      align="center"
-    >
-      <v-col>
-        Dandiset ID: {{ meta.id }}
-      </v-col>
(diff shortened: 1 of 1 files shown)


web/src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -141,7 +140,10 @@
                 </v-btn>
               </div>
             </template>
-            {{ publishDisabledMessage }}
+            <p
+              :style="{ whiteSpace: 'pre-line' }"
```
dchiquito, line 144: I have been told that generally you put this stuff into a CSS class in a separate CSS block.
However, I think decoupling that rigorously would make this harder to read, and also this is all going to get refactored shortly anyway.

dchiquito (approved):

PR #774, opened 2021-06-30
Merge staging

mvandenburgh:
VJSF updates can be merged into `master` now that the production server is running the latest schema.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -5,15 +5,8 @@ command = "yarn run build"
 
   [build.environment]
   VUE_APP_API_ROOT = "https://girder.dandiarchive.org/api/v1"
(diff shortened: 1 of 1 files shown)


satra:
@mvandenburgh - does this mean the current deployed version doesn't have these changes? or did we deploy from staging?

mvandenburgh:
> @mvandenburgh - does this mean the current deployed version doesn't have these changes? or did we deploy from staging?

Yes, the current deployed version at https://gui.dandiarchive.org/ doesn't have these changes (well, now it does because I just merged the changes). 

This PR attempted to merge `staging` directly to `master`, which is incorrect. Fixed in https://github.com/dandi/dandiarchive/pull/776.

PR #778, opened 2021-07-02
Remove remaining girder API stuff

### LINKED ISSUE #527: Remove remaining uses of Girder REST API and fill in resulting "negative space"

mvandenburgh:
Removes `girderRest` and all other Girder API related code that's still left besides the `girderize` functions. Closes #527


web/src/views/HomeView/StatsBar.vue
```diff
@@ -44,10 +44,6 @@ export default {
       dandisets: 0,
       users: 0,
       size: 0,
-      // Girder only
-      species: 0,
-      subjects: 0,
-      cells: 0,
```
waxlamp, line 50: Not entirely sure these are supposed to go away. Need some help from @dchiquito to verify.

waxlamp (commented):
Looks good overall. One question for @dchiquito to answer and then I think this is good to go.

web/src/views/HomeView/StatsBar.vue
dchiquito: These do need to go away. If you visit gui.dandiarchive.com you will notice that they are absent in the Django deployment.

dchiquito (approved):

PR #782, opened 2021-07-05
Remove `girderize` function

mvandenburgh:
Removes the `girderize` function from the codebase and updates dependent code to directly use the Girder 4 dandiset format.

### DIFF
--- a/web/src/types/index.ts
+++ b/web/src/types/index.ts
@@ -8,7 +8,6 @@ export interface Dandiset {
   identifier: string,
   created: string,
   modified: string,
-  // TODO these versions are girderized, they should have type Version
   draft_version?: any,
   most_recent_published_version?: any,
 }
--- a/web/src/types/index.ts
+++ b/web/src/types/index.ts
@@ -8,8 +8,10 @@ export interface Dandiset {
   identifier: string,
   created: string,
   modified: string,
-  draft_version?: any,
-  most_recent_published_version?: any,
+  // eslint-disable-next-line no-use-before-define
+  draft_version?: Version,
+  // eslint-disable-next-line no-use-before-define
+  most_recent_published_version?: Version,
 }
 
 export interface Version {
--- a/web/src/utils/index.js
+++ b/web/src/utils/index.js
@@ -48,7 +48,7 @@ function copyToClipboard(text) {
 }
 
 function getDandisetContact(dandiset) {
-  if (dandiset.meta.dandiset.contributors) {
+  if (dandiset.meta?.dandiset.contributors) {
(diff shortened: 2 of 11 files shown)


web/src/types/index.ts
```diff
@@ -8,7 +8,6 @@ export interface Dandiset {
   identifier: string,
   created: string,
   modified: string,
-  // TODO these versions are girderized, they should have type Version
```
mvandenburgh, line 11: It isn't possible to set these to type `Version` since `Version` is defined after `Dandiset` and has a field of type `Dandiset`.

dchiquito: Something like https://stackoverflow.com/a/48419531 might help. This feels like there is a solution, especially because both definitions are in the same file.

web/src/utils/index.js
```diff
@@ -48,7 +48,7 @@ function copyToClipboard(text) {
 }
 
 function getDandisetContact(dandiset) {
-  if (dandiset.meta.dandiset.contributors) {
+  if (dandiset.meta?.dandiset.contributors) {
```
dchiquito, line 51: Shouldn't this `meta` be removed? Is it related to the fact that this feature just doesn't work correctly right now?

web/src/views/DandisetLandingView/DandisetOwnersDialog.vue
```diff
@@ -140,7 +131,7 @@ export default {
       this.$emit('close');
     },
     async save() {
-      const { identifier } = this.publishDandiset.meta.dandiset;
+      const { identifier } = this.publishDandiset.metadata;
```
dchiquito, line 134:
```suggestion
      const identifier = this.publishDandiset.dandiset.identifier;
```

dchiquito (changes requested):

web/src/utils/index.js
mvandenburgh: Yes. If the optional chaining operator isn't there then it crashes and the page won't load. This is fixed by #783, which is set to merge into this PR's branch but I wanted to hold off on that so its code doesn't show up in the diff here.

web/src/types/index.ts
mvandenburgh: How about this? https://github.com/dandi/dandiarchive/pull/782/commits/d7c3167340f223f8b973e614f25925022ae63c12

web/src/views/DandisetLandingView/DandisetOwnersDialog.vue
mvandenburgh: https://github.com/dandi/dandiarchive/pull/782/commits/393589792701b396379b63a32efdd109c099e0fd

dchiquito (approved):

PR #784, opened 2021-07-06
Allow Ben to click the publish button

dchiquito:


waxlamp (commented):
Same question for @bendichter as in https://github.com/dandi/dandi-api/pull/394; besides that, LGTM.

PR #787, opened 2021-07-10
turn assetsSummary into a proper Summary display

sanuann:
- proper display for assetsSummary.
- remove `schemaKey` from landing page.


sanuann:
@waxlamp can you please let me know why this is failing?

sanuann:
@waxlamp please review

web/src/components/schema/Object.vue
```diff
@@ -1,12 +1,37 @@
...
     <template v-for="(val, key) in data">
-      <v-row
-        v-if="val || (!val && !omitEmpty)"
-        :key="key"
-      >
-        <strong class="mr-2">{{ objectKey(key) }}:</strong> {{ renderedValue(val) }}
-      </v-row>
+      <template v-if="isObjectArray(val)">
+        <v-row
+          v-if="val || (!val && !omitEmpty)"
```
waxlamp, line 6:
```suggestion
          v-if="val || !omitEmpty"
```

web/src/components/schema/Object.vue
```diff
@@ -61,8 +87,20 @@ export default defineComponent({
...
 
+    function renderListOfObjects(value: any[]) {
+      const newVal:any = _.map(value, (v) => v);
+      return newVal;
+    }
+
+    function isObjectArray(value: unknown) {
+      if (Array.isArray(value) && (value[0] instanceof Object)) {
+        return true;
+      } return false;
```
waxlamp, line 98: This can be simplified to

```suggestion
      return Array.isArray(value) && (value[0] instanceof Object);
```

web/src/components/schema/Object.vue
```diff
@@ -61,8 +87,20 @@ export default defineComponent({
...
+    }
     function renderedValue(value: unknown) {
-      if (Array.isArray(value)) {
+      if (Array.isArray(value) && (value[0] instanceof Object)) {
```
waxlamp, line 101: I think you should use the `isObjectArray()` function here.

web/src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -356,11 +356,13 @@ export default {
...
       );
-      const remove_list = ['citation', 'repository', 'url', 'schemaVersion', 'version', 'id', 'keywords'];
+      const remove_list = ['citation', 'repository', 'url', 'schemaVersion', 'version', 'id', 'keywords', 'schemaKey'];
       extra = extra.filter((n) => !remove_list.includes(n));
       const extra_obj = extra.reduce((obj, key) => ({ ...obj, [key]: meta[key] }), {});
       extra_obj.contributor = _.filter(meta.contributor, (author) => author.schemaKey !== 'Person');
-      // console.log(362, extra_obj);
+      delete extra_obj.assetsSummary.schemaKey;
+      delete extra_obj.assetsSummary.numberOfBytes;
+      delete extra_obj.assetsSummary.numberOfFiles;
```
waxlamp, line 365: Why remove these values? They are an important part of the asset summary.

web/src/components/schema/Object.vue
```diff
@@ -1,12 +1,37 @@
...
+      <template v-if="isObjectArray(val)">
+        <v-row
+          v-if="val || (!val && !omitEmpty)"
+          :key="key"
+          class="mx-1"
+        >
+          <strong class="mr-2">{{ objectKey(key) }}:</strong>
+        </v-row>
+        <template v-for="item in val">
+          <template v-for="(v, k) in item">
```
waxlamp, line 13: I don't believe you need this `<template>` here; instead you can put the `v-for` directive in the `<v-row>` on line 14.

web/src/components/schema/Object.vue
```diff
@@ -1,12 +1,37 @@
...
+            <v-row
+              v-if="k !== 'schemaKey'"
+              :key="`${k}`"
+              class="mx-4"
+            >
+              <strong class="mr-2">{{ k }}:</strong> {{ renderedValue(v) }}
+            </v-row>
+          </template>
+          <v-divider :key="`${item}-divider`" />
```
waxlamp, line 22: I don't think you need this `:key` attribute.

web/src/components/schema/Object.vue
```diff
@@ -1,12 +1,37 @@
...
+      <template v-else>
+        <v-row
+          v-if="val || (!val && !omitEmpty)"
+          :key="key"
+          class="mx-1"
+        >
+          <strong class="mr-2">{{ objectKey(key) }}:</strong>{{ renderedValue(val) }}
+        </v-row>
+      </template>
+      <br :key="key">
```
waxlamp, line 34: Don't need the `:key` attribute.

web/src/components/schema/Object.vue
```diff
@@ -61,8 +87,20 @@ export default defineComponent({
       );
     });
 
+    function renderListOfObjects(value: any[]) {
+      const newVal:any = _.map(value, (v) => v);
+      return newVal;
+    }
```
waxlamp, line 93: What is the purpose of this function? It looks like it just creates a shallow copy of the input array.

At the very least, this function needs better typing.

web/src/components/schema/Object.vue
waxlamp: Also, I have a feeling that this function should be a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates).

waxlamp (changes requested):
The output looks good (as seen, for example, [here](https://deploy-preview-787--gui-dandiarchive-org.netlify.app/#/dandiset/000001)), though I think @jtomeck should think about how to improve the display of the asset summary data as part of the DLP redesign.

I have made several code-related suggestions and asked some clarification questions. Please do reach out if you need any help in interpreting what I'm suggesting.

web/src/views/DandisetLandingView/DandisetMain.vue
sanuann: satra suggested not to display these. should I create a different object with only the key-values to be displayed?

sanuann: > Why remove these values? They are an important part of the asset summary.

satra suggested not to display these. should I create a different object with only the key-values to be displayed?

waxlamp: If the intent was not to display them, then this is fine.

@satra, what's the reason to skip showing these values?

satra: because they are also in the side panel. we can decide what to use where when we redo the overall design of the landing page.

web/src/components/schema/Object.vue
sanuann: sorry, this is redundant. I forgot to delete it.

web/src/components/schema/Object.vue
sanuann: if I remove the `:key` it shows linting errors. so I have put it back.

web/src/components/schema/Object.vue
sanuann: same here.

web/src/views/DandisetLandingView/DandisetMain.vue
waxlamp: Ah I see. I realized this while discussing design with Jared just now. There is also `numberOfSubjects`; could that similarly go with the number of files/assets? Perhaps with a little mouse icon?

waxlamp (approved):

PR #794, opened 2021-07-15
Provide an id for the user menu

### LINKED ISSUE #793: Provide more ids
Follow up to #648

In our particular use case we would like to get an id for the modal window with api key which appears when you click on the avatar, i.e. this :

image

edit: forgot to mention that ATM I have not spotted id for any related element there besides auto generated like `list-item-64`.

waxlamp:
@yarikoptic, let me know if this is what you had in mind.

Closes #793.


yarikoptic:
yes, that is great!  while at it if you could also provide id for that input which is I guess in the ApiKeyItem - it would also be handy so we do not need to just search for a readonly input (of which may be eventually we get multiple)

waxlamp:
> yes, that is great! while at it if you could also provide id for that input which is I guess in the ApiKeyItem - it would also be handy so we do not need to just search for a readonly input (of which may be eventually we get multiple)

Done (via 6183e9c).

Let me know if you'd like anything else taken care of in this PR!

yarikoptic (approved):
I think this should be enough for now. But I think it could be a generally good practice to tend to provide non-random `id`s to all "valuable" UI elements than "not to provide".

attn @djarecka

waxlamp:
> I think this should be enough for now. But I think it could be a generally good practice to tend to provide non-random `id`s to all "valuable" UI elements than "not to provide".
> 
> attn @djarecka

I can see the value in it, but I don't want to start an initiative to add them throughout the codebase without having a specific need. Perhaps in new components we can add some, but even there we should be careful, especially for components meant to be non-singleton within the app.

For now I'd prefer if when you run into a need for an id attribute, you just file an issue for the narrow use case.

Issue #735, opened 2021-06-14
Better UX for validation failures


dchiquito:
Right now you need to hover over the grayed out Publish button to see what failed. This is hard to find and can be communicated better.

Maybe a little hovering superscript number indicating how many validation problems there are?

(Related: https://github.com/dandi/dandiarchive/issues/459)

waxlamp:
@jtomeck, could you give us some design ideas for this?

jtomeck:
I've actually been thinking about integrating a notifications menu to the page, but want sure if there was anything to be notified about. This could be the use case for it though. I can have a mockup for you in the morning.

jtomeck:
<img width="473" alt="Screen Shot 2021-06-17 at 1 17 56 PM" src="https://user-images.githubusercontent.com/1560779/122444668-ca297980-cf6e-11eb-95a2-362152ffa312.png">

Sorry this took longer than expected. I sat with Mike and Roni to demo how and what validation failures occurred, so my mockup could be a bit better informed. The issues would show up in this dropdown next to the publish button and the "fix" button would take you to the meditor to show you specifically what isn't valid.

**An alternate option**
Now that I'm thinking about it though, we might not even need the list since everything that's wrong gets highlighted in the meditor anyway. It could just be the warning icon with the number of issues next to it (like shown in the mockup), as well as a tooltip that says "fix issues," and clicking the icon brings you to the meditor.

satra:
making publish green would require addressing both meditor issues and asset issues. it may be worthwhile discussing this and figuring out a clean interface.

jtomeck:
> asset issues

By asset issues do you mean issues with the actual files? Regardless, I think the alternate option I discussed above would be best. 

1. Show 2 icons next to the publish button:
    - [this one](https://materialdesignicons.com/icon/clipboard-alert) for metadata issues, although there could be a better one.
    - [this one](https://materialdesignicons.com/icon/database-alert) for file issues)
2. Each of them have a tooltip:
    - "Fix metadata issues"
    - "Fix asset issues"
3. Clicking on the icon will bring you to the correct place to make those fixes.

--

This is my assessment with limited knowledge on the "asset" part of it, so if more functionality is needed there let me know.

waxlamp:
> **An alternate option**
[...]

PR #796, opened 2021-07-16
Validation error UX improvements 

mvandenburgh:
Start implementing UX improvements for publish validation errors on the DLP based on the mockups in #735.

This requires some breaking server changes so I want to get this watered-down version merged first and then open another PR to finish implementing Jared's full design from #735.

Requires https://github.com/dandi/dandi-api/pull/417 to be merged first.

Screenshot from 2021-07-19 13-18-42
Screenshot from 2021-07-19 13-18-59
Screenshot from 2021-07-19 13-18-51

### DIFF
--- a/src/views/DandisetLandingView/DandisetMain.vue
+++ b/src/views/DandisetLandingView/DandisetMain.vue
@@ -119,15 +119,14 @@
         <template v-if="publishDandiset.version === 'draft'">
           <v-tooltip
             left
-            :disabled="publishDisabledMessage === null"
+            :disabled="!publishDisabledMessage"
           >
             <template #activator="{ on }">
               <div v-on="on">
-                <!-- TODO for now only admins can publish -->
                 <v-btn
                   id="publish"
                   text
-                  :disabled="publishDisabledMessage !== null || !user
+                  :disabled="publishDisabled || !user
                     || !(user.admin ||
                       user.username === 'bendichter' || user.username === 'ben.dichter@gmail.com')"
                   @click="publish"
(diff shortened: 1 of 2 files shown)


src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -116,18 +116,18 @@
           </v-icon>
           {{ metadataButtonText }}
         </v-btn>
+        <!-- TODO for now only admins can publish -->
```
dchiquito, line 119:
```suggestion
```
I think this is just an artifact of a merge.

src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -141,12 +141,116 @@
...
+                v-text="publishDisabledMessage"
+              />
             </template>
-            <p
-              :style="{ whiteSpace: 'pre-line' }"
-              v-text="publishDisabledMessage"
-            />
           </v-tooltip>
+          <v-menu
+            v-if="publishDandiset.version_validation_errors.length"
```
dchiquito, line 151:
```suggestion
            v-if="publishDandiset.version_validation_errors"
```
Is `.length` required?

src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -141,12 +141,116 @@
...
+                  <v-divider />
+                </div>
+              </v-list>
+              <v-btn color="primary">
+                Fix issues
+              </v-btn>
+            </v-card>
+          </v-menu>
+          <v-menu
+            v-if="publishDandiset.asset_validation_errors.length"
```
dchiquito, line 203: Same question about .length

src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -393,6 +498,13 @@ export default {
       }
       return true;
     },
+    getValidationErrorIcon(errorField) {
+      const icons = Object.keys(VALIDATION_ICONS).filter((field) => field.includes(errorField));
```
dchiquito, line 502: What's wrong with the naive `return VALIDATION_ICONS[errorField] || VALIDATION_ICONS.default`?

Also, what's the motivation for doing `field.includes(errorField)` instead of `errorField.includes(field)`? IIRC `errorField` might be something nested like `assetsSummary.creator.0.name`, but `field` will basically never be a superset of an `errorField`.

dchiquito (changes requested):
`asset_validation_errors` and `version_validation_errors` should be added to `types/index.ts#Version`.

src/views/DandisetLandingView/DandisetMain.vue
mvandenburgh: Yup. Empty arrays are "truthy" in JS :roll_eyes: :

src/views/DandisetLandingView/DandisetMain.vue
mvandenburgh: https://github.com/dandi/dandiarchive/pull/796#discussion_r681058708

src/views/DandisetLandingView/DandisetMain.vue
mvandenburgh: > What's wrong with the naive `return VALIDATION_ICONS[errorField] || VALIDATION_ICONS.default`?

The intention was to support something like `assetsSummary.creator.0.name`, where the correct entry in `VALIDATION_ICONS` would be `name`.

> Also, what's the motivation for doing `field.includes(errorField)` instead of `errorField.includes(field)`? IIRC `errorField` might be something nested like `assetsSummary.creator.0.name`, but `field` will basically never be a superset of an `errorField`.

You're right, I swapped those around. https://github.com/dandi/dandiarchive/pull/796/commits/2ac2d1ef637b9e945caffbc942c1c5ec575f8220

src/views/DandisetLandingView/DandisetMain.vue
mvandenburgh: https://github.com/dandi/dandiarchive/pull/796/commits/d4194bfef2bf4e5a8df4192ae89d8d99842567cf

dchiquito (approved):

PR #799, opened 2021-07-19
Disable publish button when there are no changes

dchiquito:
When dandiset status is `Published`, the publish button was still enabled, even though the publish request would fail (as of https://github.com/dandi/dandi-api/pull/419).

Fix that with a useful error message.


mvandenburgh (approved):
LGTM. tested locally and it no longer lets me overload the server by spamming the publish button :smile:.

PR #820, opened 2021-07-29
Add notice about cookie usage

mvandenburgh:
There was some discussion about adding a notice about cookies to comply with the GDPR. This PR adds a banner to the bottom of the webpage that indicates cookies are being used when a user navigates to the archive for the first time -  not sure if that is sufficient for compliance.


waxlamp (approved):
This looks great. I think we'll eventually need to offer the option to not use non-essential cookies, but we can figure that out in a followup.

PR #822, opened 2021-07-30
Allow any owner to push the publish button

dchiquito:
Leaving this in draft for now so we don't accidentally merge.

Requires https://github.com/dandi/dandi-api/pull/452 for pushing the button to actually work.


waxlamp (approved):
:clap: :clap: :clap:

PR #823, opened 2021-08-02
fix ethics approval view

sanuann:
- fixes ethics approval view
- removes `schemaKey` from all array type values.

### DIFF
--- a/src/views/DandisetLandingView/DandisetMain.vue
+++ b/src/views/DandisetLandingView/DandisetMain.vue
@@ -369,7 +369,7 @@ export default {
       _.forEach(extra_obj, (val) => {
         if (Array.isArray(val) && val.length !== 0) {
           val.forEach((each_val) => {
-            if (each_val instanceof Object && Object.keys(each_val).includes('schemaKey')) {
+            if (Object.prototype.hasOwnProperty.call(each_val, 'schemaKey')) {
               // eslint-disable-next-line no-param-reassign
               delete each_val.schemaKey;
             }
--- a/src/views/DandisetLandingView/DandisetMain.vue
+++ b/src/views/DandisetLandingView/DandisetMain.vue
@@ -366,13 +366,12 @@ export default {
       delete extra_obj.assetsSummary.schemaKey;
       delete extra_obj.assetsSummary.numberOfBytes;
       delete extra_obj.assetsSummary.numberOfFiles;
-      _.map(extra_obj, (val) => {
+      _.forEach(extra_obj, (val) => {
         if (Array.isArray(val) && val.length !== 0) {
-          _.map(val, (each_val) => {
+          val.forEach((each_val) => {
             if (each_val instanceof Object && Object.keys(each_val).includes('schemaKey')) {
               // eslint-disable-next-line no-param-reassign
               delete each_val.schemaKey;
-              console.log(372, each_val, Object.keys(each_val));
             }
           });
         }
--- a/src/views/DandisetLandingView/DandisetMain.vue
+++ b/src/views/DandisetLandingView/DandisetMain.vue
@@ -370,7 +370,8 @@ export default {
         if (Array.isArray(val) && val.length !== 0) {
           val.forEach((each_val) => {
             if (Object.prototype.hasOwnProperty.call(each_val, 'schemaKey')) {
-              // eslint-disable-next-line no-param-reassign
+              /* eslint no-param-reassign:["error",
+              {"ignorePropertyModificationsFor":["each_val"] }] */
               delete each_val.schemaKey;
             }
           });
--- a/src/views/DandisetLandingView/DandisetMain.vue
+++ b/src/views/DandisetLandingView/DandisetMain.vue
@@ -366,6 +366,17 @@ export default {
       delete extra_obj.assetsSummary.schemaKey;
       delete extra_obj.assetsSummary.numberOfBytes;
       delete extra_obj.assetsSummary.numberOfFiles;
+      _.map(extra_obj, (val) => {
+        if (Array.isArray(val) && val.length !== 0) {
+          _.map(val, (each_val) => {
+            if (each_val instanceof Object && Object.keys(each_val).includes('schemaKey')) {
+              // eslint-disable-next-line no-param-reassign
+              delete each_val.schemaKey;
+              console.log(372, each_val, Object.keys(each_val));
+            }
+          });
+        }
+      });
       return extra_obj;
     },
     schemaPropertiesCopy() {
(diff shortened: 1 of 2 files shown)


src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -366,6 +366,17 @@ export default {
       delete extra_obj.assetsSummary.schemaKey;
       delete extra_obj.assetsSummary.numberOfBytes;
       delete extra_obj.assetsSummary.numberOfFiles;
+      _.map(extra_obj, (val) => {
+        if (Array.isArray(val) && val.length !== 0) {
+          _.map(val, (each_val) => {
+            if (each_val instanceof Object && Object.keys(each_val).includes('schemaKey')) {
```
waxlamp, line 372: This can be simplified to:

```suggestion
            if (each_val.hasOwnProperty('schemaKey')) {
```

I am also not sure the condition is even necessary. This first checks to see if the `schemaKey` property exists and removes it if so. I think you can skip straight to the `delete` statement and skip the conditional check.

src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -366,6 +366,17 @@ export default {
...
+              // eslint-disable-next-line no-param-reassign
+              delete each_val.schemaKey;
+              console.log(372, each_val, Object.keys(each_val));
```
waxlamp, line 375:
```suggestion
```

src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -366,6 +366,17 @@ export default {
       delete extra_obj.assetsSummary.schemaKey;
       delete extra_obj.assetsSummary.numberOfBytes;
       delete extra_obj.assetsSummary.numberOfFiles;
+      _.map(extra_obj, (val) => {
+        if (Array.isArray(val) && val.length !== 0) {
+          _.map(val, (each_val) => {
```
waxlamp, line 371: These two `_.map()` calls should be `_.forEach()` instead.

Even better would be to use the JavaScript native `Array.forEach()` function rather than lodash.

waxlamp (changes requested):

src/views/DandisetLandingView/DandisetMain.vue
sanuann: what if the value is not an Object?

waxlamp: It's probably not harmful to check whether it's an `Object`, but the check for `schemaKey` isn't needed.

You should keep in mind that other things in JavaScript are also objects, specifically arrays and functions.

src/views/DandisetLandingView/DandisetMain.vue
sanuann: the outer `_.map()` is on an object, not an array.

satra:
@sanuann - since the DLP is tied to the metadata schema, you should take a look at the metadata schema to see what the different things are.

sanuann:
> @sanuann - since the DLP is tied to the metadata schema, you should take a look at the metadata schema to see what the different things are.

is this a general comment or something to fix in my PR?

satra:
> is this a general comment or something to fix in my PR?

a bit of a general comment so that you could check ahead of time if other elements if added to the metadata would have resulted in the same issue.

src/views/DandisetLandingView/DandisetMain.vue
sanuann: for the outer one i changed it to `_.forEach` and the inner one to `Array.forEach`

src/views/DandisetLandingView/DandisetMain.vue
sanuann: if I use `each_val.hasOwnProperty('schemaKey')` , I get this ESLint message:
`ESLint: Do not access Object.prototype method 'hasOwnProperty' from target object.(no-prototype-builtins)`.

How does it matter if I use `Object.keys(each_val).includes('schemaKey')` rather than `each_val.hasOwnProperty('schemaKey')` ? The first one doesn't throw a linting error too.

waxlamp: It will be important to Google some of these topics as your work in order to build an understanding of best practices. In this case, I searched on the linting message you got and found this [documentation page](https://eslint.org/docs/rules/no-prototype-builtins) that explains the dangers of calling methods like `.hasOwnProperty()` directly, and what to do instead.

To answer your question, `Object.keys(each_val).includes()` has O(n) runtime, and though it likely will not make a difference in this case, it is better to do the standard thing: in this case you solve the performance problem, and you write code that adheres to best practices and thus will cause less burden on reviewers, other developers, and even yourself in the future if and when you come back to this code.

waxlamp:
> > is this a general comment or something to fix in my PR?
> 
> a bit of a general comment so that you could check ahead of time if other elements if added to the metadata would have resulted in the same issue.

But you can certainly rely on the types expressed in the schema to eliminate some of the runtime typechecking in this PR too.

src/views/DandisetLandingView/DandisetMain.vue
sanuann: thank you for the explanation.

src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -366,6 +366,16 @@ export default {
       delete extra_obj.assetsSummary.schemaKey;
       delete extra_obj.assetsSummary.numberOfBytes;
       delete extra_obj.assetsSummary.numberOfFiles;
+      _.forEach(extra_obj, (val) => {
+        if (Array.isArray(val) && val.length !== 0) {
+          val.forEach((each_val) => {
+            if (Object.prototype.hasOwnProperty.call(each_val, 'schemaKey')) {
+              // eslint-disable-next-line no-param-reassign
+              delete each_val.schemaKey;
```
waxlamp, line 374: This rule doesn't seem to apply here. Do you get an error when you remove it?

sanuann: yes. and i found this [documentation](https://eslint.org/docs/rules/no-param-reassign) . so perhaps it's ok to disable the rule?

waxlamp: It looks like this [more specific version](https://eslint.org/docs/rules/no-param-reassign#props) is the thing you want here.

If you can verify that and you change it here, you can merge without another review.

waxlamp (approved):
Looks good, possibly with a small alteration to the eslint rule disabling statement.

src/views/DandisetLandingView/DandisetMain.vue
sanuann: did you mean that I should write it this way:
```
/* eslint no-param-reassign:["error", {"ignorePropertyModificationsFor":["each_val"] }] */
delete each_val.schemaKey;
```
?

PR #826, opened 2021-08-03
Show correct last-modified time in dandiset detail panel

### LINKED ISSUE #825: "Last updated" field shows current datetime
image

The "Last updated" field shows the current datetime. When I refresh the page, the last updated field changes to the now current datetime, even though we are not making any changes to the dataset.

waxlamp:
The non-existent `updated` property was previously being passed to
momentjs as the last modified date. Since that value was always
`undefined`, and momentjs returns the current date and time when called
without arguments (or with `undefined`, apparently), this explains why
the frontend was always showing the current time for the last modified
value.

Closes #825.

### DIFF
--- a/src/views/DandisetLandingView/DandisetDetails.vue
+++ b/src/views/DandisetLandingView/DandisetDetails.vue
@@ -26,8 +26,8 @@
       :class="rowClasses"
     >
       <v-icon>mdi-update</v-icon>
-      <span :class="labelClasses">Last updated</span>
-      <span :class="itemClasses">{{ lastUpdated }}</span>
+      <span :class="labelClasses">Last modified</span>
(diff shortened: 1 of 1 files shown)


waxlamp:
@bendichter, take a look at https://deploy-preview-826--gui-dandiarchive-org.netlify.app/#/ to see the fix in action (on staging data, so you won't see your dandisets there).

mvandenburgh (approved):
Looks like another bug introduced by a schema change that went unnoticed. LGTM

bendichter:
lgtm, thanks @waxlamp

waxlamp:
> Looks like another bug introduced by a schema change that went unnoticed. LGTM

More reason to be adopting more TypeScript and Composition API in this project.

PR #829, opened 2021-08-06
Re-fetch dandiset metadata after modifying w/ meditor

mvandenburgh:
#827 is happening because the web client doesn't fetch the new version metadata from the server after clicking the save button in the meditor. Instead, it updates the UI using the "local copy" of the version metadata that the user has just edited; this approach is no longer feasible since we are now displaying validation errors in the UI. This PR fixes this by triggering an API request for fresh metadata when the user saves changes in the meditor.

Note that validation is an asynchronous process, so there isn't actually a guarantee that it will be complete by the time the user makes a request for fresh data. To solve this "correctly", a flag in the database that is set when a version's/asset's validation errors are "stale" could be implemented in a future PR.

### DIFF
--- a/src/views/DandisetLandingView/Meditor.vue
+++ b/src/views/DandisetLandingView/Meditor.vue
@@ -245,8 +245,11 @@ export default defineComponent({
         );
 
         if (status === 200) {
-          await store.dispatch('dandiset/fetchPublishDandiset', { identifier: data.dandiset.identifier, version: data.version });
-          closeEditor();
+          // wait 0.5 seconds to give the celery worker some time to finish validation
+          setTimeout(async () => {
+            await store.dispatch('dandiset/fetchPublishDandiset', { identifier: data.dandiset.identifier, version: data.version });
+            closeEditor();
+          }, 500);
         }
       } catch (error) {
         if (error.response.status === 403) {
--- a/src/views/DandisetLandingView/Meditor.vue
+++ b/src/views/DandisetLandingView/Meditor.vue
@@ -235,10 +235,6 @@ export default defineComponent({
     }));
     const publishDandiset = computed(() => store.state.dandiset.publishDandiset);
     const id = computed(() => publishDandiset.value?.dandiset.identifier || null);
-    function setDandiset(payload: any) {
-      // TODO: Replace once direct-vuex is added
-      store.commit('dandiset/setPublishDandiset', payload);
-    }
 
     async function save() {
       const dandiset = editorInterface.getModel();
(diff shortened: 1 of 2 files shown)


src/views/DandisetLandingView/Meditor.vue
```diff
@@ -249,8 +245,11 @@ export default defineComponent({
         );
 
         if (status === 200) {
-          setDandiset(data);
-          closeEditor();
+          // wait 0.5 seconds to give the celery worker some time to finish validation
+          setTimeout(async () => {
+            await store.dispatch('dandiset/fetchPublishDandiset', { identifier: data.dandiset.identifier, version: data.version });
+            closeEditor();
+          }, 500);
```
waxlamp, line 252: I think this is the place where we would have to do what you described in your PR description. Could you file an issue to change the application behavior to have an intermediate validation state (something like `computing`) that the frontend can use to display a throbber or some sort in place of the error list?

waxlamp (approved):

src/views/DandisetLandingView/Meditor.vue
mvandenburgh: I just realized there's already [a flag like that](https://github.com/dandi/dandi-api/blob/master/dandiapi/api/models/version.py#L70) implemented, so this will just be a matter of adding that field to a django serializer and making use of it in the frontend. I filed an issue for the frontend changes in #830.

yarikoptic:
>  This PR fixes this by triggering an API request for fresh metadata when the user saves changes in the meditor.

is there any check that (re)loaded metadata corresponds to the original one?  if not -- then user might be saving outdated "edited" if someone else edited it also meanwhile.  github e.g. IIRC would refuse to save e.g. issue/PR description if it was modified interim -- it will display a message about that and IIRC would then allow to save it only on 2nd try or smth like that

waxlamp:
> > This PR fixes this by triggering an API request for fresh metadata when the user saves changes in the meditor.
> 
> is there any check that (re)loaded metadata corresponds to the original one? if not -- then user might be saving outdated "edited" if someone else edited it also meanwhile. github e.g. IIRC would refuse to save e.g. issue/PR description if it was modified interim -- it will display a message about that and IIRC would then allow to save it only on 2nd try or smth like that

I believe this race condition is indeed possible. We probably have never run into it but it is worth thinking about defenses against it. @mvandenburgh can you file one more issue about this?

mvandenburgh:
> if not -- then user might be saving outdated "edited" if someone else edited it also meanwhile.

Yeah this is exactly what would happen, good catch!

> I believe this race condition is indeed possible. We probably have never run into it but it is worth thinking about defenses against it. @mvandenburgh can you file one more issue about this?

#831

PR #833, opened 2021-08-09
Prevent crash if cookies are disabled

### LINKED ISSUE #768: Test for being allowed to place a cookie and inform if not
#750 provided a workaround for my initial symptom but not the underlying issue -- not allowed cookies: https://github.com/dandi/dandiarchive/pull/750#issuecomment-864098495

if browser has cookies disabled -- the same issue happens without any iframe

image

Didn't look into details of current policies (AFAIK required for GDPR at least), but as most of the sites request confirmation of the "Cookie consent" -- we need to

- request "Cookie consent"
- ask to enable mandatory cookie(s) 
- not crash (and just not disable login , but display some button to bring up some modal information on cookies consent/enabling information)

After this addressed, #750  could then be reverted AFAIK

mvandenburgh:
Previously, the site would crash and show a white screen if cookies were disabled by the user in their browser settings. This was due to both our Girder OAuth client and the vjsf library both assuming they have access to `localStorage` at all times (they don't have access when cookies are disabled). The issue in Girder OAuth client is fixed by this PR, and the issue with vjsf has been fixed upstream.

Now if cookies are disabled:
* the user will see a message warning about reduced functionality in place of the message notifying them about the use of cookies.
* the login button will be grayed out with a tooltip telling the user to enable cookies if they wish to login.

Closes #768 

To test out the changes, disable cookies in your browser settings.

### DIFF
--- a/src/components/AppBar/AppBar.vue
+++ b/src/components/AppBar/AppBar.vue
@@ -95,22 +95,35 @@
         <UserMenu />
       </template>
       <template v-else>
-        <v-btn
-          id="login"
-          class="mx-1"
-          color="primary"
-          rounded
-          @click="login"
+        <v-tooltip
+          bottom
+          :disabled="cookiesEnabled"
         >
-          Log In with GitHub
-        </v-btn>
+          <template #activator="{ on }">
+            <div v-on="on">
+              <v-btn
+                id="login"
+                class="mx-1"
+                color="primary"
+                rounded
+                :disabled="!cookiesEnabled"
+                @click="login"
+              >
+                Log In with GitHub
+              </v-btn>
+            </div>
+          </template>
+          <span>Enable cookies to log in.</span>
+        </v-tooltip>
       </template>
     </div>
   </v-app-bar>
 </template>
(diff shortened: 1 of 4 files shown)


dchiquito (approved):
Looks great!

yarikoptic:
I think this works nice!  Here is my "tester":

## http

in http://datasets.datalad.org/.git/tmp/iframe.html  and the "Staging" is blank for me (due to oauth script error), whenever bottom one produced that  nice message at the bottom
image

interestingly in brave browser (probably allows all cookies, heh) it did not warn about cookies not being allowed and said only "We use cookies to ensure you get the best experience on DANDI." and worked (unlike staging -- blank)

## https

https://datasets.datalad.org/.git/tmp/iframe.html 

similar to above -- works nice with this PR

in brave (cookies allowed) -- identical view of staging and this PR with the regular "we use cookies".

yarikoptic (approved):

waxlamp:
Thank you for testing, @yarikoptic!

PR #848, opened 2021-08-19
Re-enable E2E tests

### LINKED ISSUE #832: Update and re-enable dandiarchive e2e tests
The e2e tests were written for the old girder 3 setup and were disabled after the django migration. They should be updated so they work on the current web client and Django server.

This has been a thing for a while, but I didn't see an issue for this filed anywhere.

mvandenburgh:
Updates the end-to-end tests to work with the current system and re-enables them in CI.

Closes #832.

### DIFF
--- a/test/src/tests/account.test.js
+++ b/test/src/tests/account.test.js
@@ -1,8 +1,11 @@
 import {
   vAvatar,
+  vBtn,
+  vIcon,
+  vListItem,
 } from 'jest-puppeteer-vuetify';
 import {
-  logoutUser, registerNewUser, LOGIN_BUTTON_TEXT, authorize,
+  registerNewUser, LOGIN_BUTTON_TEXT, LOGOUT_BUTTON_TEXT,
 } from '../util';
 
 describe('account management', () => {
@@ -10,8 +13,7 @@ describe('account management', () => {
     await registerNewUser();
 
     await expect(page).toClickXPath(vAvatar('??'));
-
-    await logoutUser();
+    await expect(page).toClickXPath(vListItem(LOGOUT_BUTTON_TEXT, { action: vIcon('mdi-logout') }));
 
     // this text is only displayed when not logged in
     await expect(page).toMatch(LOGIN_BUTTON_TEXT);
@@ -20,11 +22,12 @@ describe('account management', () => {
   it('logs the user in', async () => {
     await registerNewUser();
 
+    // Logout
     await expect(page).toClickXPath(vAvatar('??'));
+    await expect(page).toClickXPath(vListItem(LOGOUT_BUTTON_TEXT, { action: vIcon('mdi-logout') }));
 
-    await logoutUser();
-
-    await authorize();
+    // Test logging in
+    await expect(page).toClickXPath(vBtn(LOGIN_BUTTON_TEXT));
 
     // the user avatar contains the initials and is only rendered when logged in successfully
     await expect(page).toContainXPath(vAvatar('??'));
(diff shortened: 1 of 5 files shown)


test/src/tests/account.test.js
```diff
@@ -1,41 +1,30 @@
...
 
 describe('account management', () => {
   it('logs the user out', async () => {
     await registerNewUser();
 
     await expect(page).toClickXPath(vAvatar('??'));
-    await page.waitFor(500);
-    await expect(page).toClickXPath(vListItem({ content: 'Logout', action: vIcon('mdi-logout') }));
+
+    await logoutUser();
```
dchiquito, line 14: This was actually a deliberate design decision, not a redundant bit of code.

For frontend tests, I've found that there are frequently multiple ways of getting the GUI into a certain state, some of which are faster than others. The `util` module is a collection of helpers so tests can set themselves up quickly and easily. All the `util` helpers should only be used when you need something done and you don't care how. 

When writing the test, `utils` should only be used to set up and tear down. The actual thing being tested should be written out explicitly, even if it is technically duplicating code that already exists in the helper. We might eventually optimize the helper and do things a different way, in which case the test is now testing something completely different, with no indication that the behavior has changed.

I've found that violating DRY in this case makes maintaining tests more reliable, but I'm happy to discuss.

dchiquito (changes requested):

test/src/tests/account.test.js
mvandenburgh: I was wondering about that, thanks for the explanation. Thinking about future tests, some of the operations currently being tested (signing up a user, logging in, registering a dandiset, etc) will be needed to setup more involved tests, so having those basic tests be explicit while we use helper functions for the more involved tests makes sense to me.

test/src/tests/stats.test.js
```diff
@@ -31,6 +31,10 @@ describe('home page stats', () => {
     const dandisetName = `dandiset${uniqueId()}`;
     const dandisetDescription = `Description! ${uniqueId()}`;
 
+    // The value that `homePage.getStat` looks for is retrieved via XHR, so
+    // wait until all requests are finished before trying to retrieve it.
+    await page.waitForNavigation({ waitUntil: 'networkidle0' });
```
dchiquito, line 36: This is a really neat feature in puppeteer, but I have a vague memory of some problem with a third party package (google analytics or sentry or something) that was sending requests every few seconds, which prevented `waitForNavigation` from ever finishing. I'm happy to try it again, but be aware if it ever starts acting up.

test/src/util.js
```diff
@@ -15,15 +18,7 @@ export function uniqueId() {
  * Authorizes the user account.
  */
 export async function authorize() {
```
dchiquito, line 20:
```suggestion
export async function login() {
```
This was originally named after the Girder 3 feature, which is no longer relevant.

dchiquito (approved):

test/src/tests/stats.test.js
mvandenburgh: https://github.com/dandi/dandiarchive/pull/848/commits/e8102c3b44c010b145b25affdb1b39552b8702e3 I updated it so it times out at 10 seconds (which should be more then enough time for any of our http requests to complete anyways)

test/src/util.js
mvandenburgh: https://github.com/dandi/dandiarchive/pull/848/commits/ee1cc5efc656756f0f9eb55c11a47c27ca52cd55

waxlamp (commented):
Really nice work here, thanks @mvandenburgh!

PR #849, opened 2021-08-20
[wip] adding nwb viewer

### LINKED ISSUE #821: Add a context menu with options for "opening" an asset (or dandiset, folder?) on external services
original idea: https://github.com/dandi/dandiarchive/wiki/WiP:-dump-of-thoughts-on-%22services-registry%22

An [issue](https://github.com/MetaCell/nwb-explorer/issues/244) with metacell nwb viewer resolved , so we have no roadblock to start working on this feature to gain initial integrations. And bisweb might be just a CORR tune up away: https://github.com/dandi/dandi-infrastructure/pull/88 .

Similar implementation [is ongoing](https://github.com/datalad/datalad-deprecated/pull/27) for https://datasets.datalad.org and could be previewed in action for a sample dandiset at http://datasets-dev.datalad.org/?dir=/dandi/dandisets/000027/sub-RAT123 (hover over file to see ⠇)

Instead (or in addition) to a context menu it could be an action to take upon a click on an asset (https://github.com/dandi/dandiarchive/issues/780) but [I still think](https://github.com/dandi/dandiarchive/issues/780#issuecomment-885061015) we better reserve that for a "view" of some kind without resorting to external services.

also ref: https://github.com/dandi/dandi-infrastructure/issues/10 as a generic metacell deployment issue (not necessarily just linking out)

djarecka:
@waxlamp , @mvandenburgh - please take a look if this is a good place and way to add this extra button. 

This is still not finished, since I'm having problem with providing a correct url to the bioimagesuiteweb viewer. 
@waxlamp - I remember it worked for you, but I'm having some issues now. Trying different URLs from the dandi info website, but each time the viewer gives me an error.
 Could you please recreate a proper url for [this file](https://api.dandiarchive.org/api/dandisets/000126/versions/0.210813.0327/assets/902af029-f411-40b5-81f1-67930d56ee05/) as an example. 

closes #821

### DIFF
--- a/src/views/FileBrowserView/PublishFileBrowser.vue
+++ b/src/views/FileBrowserView/PublishFileBrowser.vue
@@ -226,6 +226,7 @@ const EXTERNAL_SERVICES = [
       '.nii',
       '.nii.gz',
     ],
+    maxsize: 4e6,
     endpoint: 'https://bioimagesuiteweb.github.io/unstableapp/viewer.html?image=',
   },
 
@@ -235,6 +236,7 @@ const EXTERNAL_SERVICES = [
     extensions: [
       '.nwb',
     ],
+    maxsize: 4e6,
     endpoint: 'http://nwbexplorer.opensourcebrain.org/nwbfile=',
   },
 ];
@@ -294,8 +296,8 @@ export default {
           ).sort(sortByName),
           ...Object.keys(data.files).map(
             (key) => {
-              const { id } = data.files[key];
-              const services = this.getExternalServices(id, key);
+              const { id, size } = data.files[key];
+              const services = this.getExternalServices(id, key, size);
               return {
                 ...data.files[key],
                 name: key,
@@ -353,10 +355,10 @@ export default {
       return publishRest.assetDownloadURI(this.identifier, this.version, asset_id);
     },
 
-    getExternalServices(asset_id, name) {
+    getExternalServices(asset_id, name, size) {
       const { identifier, version } = this;
       return EXTERNAL_SERVICES
-        .filter((service) => new RegExp(service.regex).test(name))
+        .filter((service) => new RegExp(service.regex).test(name) && size <= service.maxsize)
         .map((service) => ({
           name: service.name,
           url: `${service.endpoint}${publishRest.assetDownloadURI(identifier, version, asset_id)}`,
--- a/src/views/FileBrowserView/PublishFileBrowser.vue
+++ b/src/views/FileBrowserView/PublishFileBrowser.vue
@@ -140,17 +140,49 @@
                 </v-btn>
               </v-list-item-action>
 
-              <v-list-item-action v-if="item.asset_id && fileExtension(item.name)">
-                <v-btn
-                  icon
-                  :href="viewURI(item.asset_id, item.name)"
-                  target="_blank"
-                  rel="noopener"
+              <v-list-item-action v-if="item.asset_id">
+                <v-col
+                  cols="12"
+                  sm="6"
+                  offset-sm="3"
                 >
-                  <v-icon color="primary">
-                    mdi-file-eye
-                  </v-icon>
-                </v-btn>
+                  <v-menu
+                    bottom
+                    left
+                  >
+                    <template #activator="{ on, attrs }">
+                      <v-btn
+                        color="primary"
+                        icon
+                        v-bind="attrs"
+                        v-on="on"
+                      >
+                        <v-icon>mdi-dots-vertical</v-icon>
+                      </v-btn>
+                    </template>
+                    <v-list dense>
+                      <v-subheader
+                        v-if="getExternalServices(item.asset_id, item.name).length"
+                        class="font-weight-medium"
+                      >
+                        EXTERNAL SERVICES
+                      </v-subheader>
+                      <v-subheader v-else>
+                        EXTERNAL SERVICES: not found
+                      </v-subheader>
+
+                      <v-list-item
+                        v-for="el in getExternalServices(item.asset_id, item.name)"
+                        :key="el.name"
+                        :href="el.url"
+                      >
+                        <v-list-item-title class="font-weight-light">
+                          {{ el.name }}
+                        </v-list-item-title>
+                      </v-list-item>
+                    </v-list>
+                  </v-menu>
+                </v-col>
               </v-list-item-action>
 
               <v-list-item-action
@@ -185,6 +217,27 @@ const sortByName = (a, b) => {
   }
   return 0;
 };
+
+const EXTERNAL_SERVICES = [
+  {
+    name: 'Bioimagesuite/Viewer',
+    regex: '.nii(.gz)?$',
+    extensions: [
+      '.nii',
+      '.nii.gz',
+    ],
+    endpoint: 'https://bioimagesuiteweb.github.io/alphaapp/viewer.html?image=',
+  },
+
+  {
+    name: 'MetaCell/NWBExplorer',
+    regex: '.nwb$',
+    extensions: [
+      '.nwb',
+    ],
+    endpoint: 'http://nwbexplorer.opensourcebrain.org/nwbfile=',
+  },
+];
 
 export default {
   name: 'PublishFileBrowser',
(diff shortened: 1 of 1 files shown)


yarikoptic:
> This is still not finished, since I'm having problem with providing a correct url to the bioimagesuiteweb viewer.

No URL from dandi would work for it just yet, https://github.com/bioimagesuiteweb/bisweb/issues/127  but I hope the fix is soon to come and IMHO it should be just the asset download url (to API, which mints signed url with content disposition) which web ui already provide for download

djarecka:
@yarikoptic - thanks for clarifying. I thought, that @waxlamp was able to see the image last week, but maybe it was something special about that image

yarikoptic:
re bisweb - fix was implemented, and available in "alpha" deployment to try out on e.g. https://bioimagesuiteweb.github.io/alphaapp/viewer.html?image=https://api.dandiarchive.org/api/assets/5ab0247b-659a-4593-b251-b338676de725/download/

src/views/FileBrowserView/PublishFileBrowser.vue
```diff
@@ -278,6 +291,11 @@ export default {
       return publishRest.assetDownloadURI(this.identifier, this.version, asset_id);
     },
 
+    viewURI(asset_id) {
+      return "https://bioimagesuiteweb.github.io/webapp/overlayviewer.html?image=" + publishRest.assetDownloadURI(this.identifier, this.version, asset_id);
+    },
+
+
```
dchiquito, line 298: This should resolve the current linting errors. `yarn lint` should autofix and show any new errors.
```suggestion
    viewURI(asset_id) {
      return `https://bioimagesuiteweb.github.io/webapp/overlayviewer.html?image=${publishRest.assetDownloadURI(this.identifier, this.version, asset_id)}`;
    },

```

djarecka:
@yarikoptic - your link works! but any idea why this is not working?
https://bioimagesuiteweb.github.io/alphaapp/viewer.html?image=https://api.dandiarchive.org/api/assets/902af029-f411-40b5-81f1-67930d56ee05/download/

just trying to make it works for [this example](https://api.dandiarchive.org/api/dandisets/000126/versions/0.210813.0327/assets/902af029-f411-40b5-81f1-67930d56ee05/)

djarecka:
@yarikoptic - wait, I think this might just work for `nii.gz`. I thought that this was for `nwb`...

djarecka:
I will need some help with testing this functionality

src/views/FileBrowserView/PublishFileBrowser.vue
```diff
@@ -278,6 +291,31 @@ export default {
...
     },
 
+    viewURI(asset_id, name) {
+      const file_ext = this.fileExtension(name);
+      if (file_ext === 'nii') {
+        return `https://bioimagesuiteweb.github.io/alphaapp/viewer.html?image=${publishRest.assetDownloadURI(this.identifier, this.version, asset_id)}`;
+      }
+      if (file_ext === 'nwb') {
+        return `http://nwbexplorer.opensourcebrain.org/nwbfile=${publishRest.assetDownloadURI(this.identifier, this.version, asset_id)}`;
+      }
```
yarikoptic, line 301: I would prefer if code was separated from registry of services, so that registry could be reused later by multiple portals.
See https://github.com/datalad/datalad-deprecated/blob/master/datalad_deprecated/resources/website/assets/js/main.js#L21 and how used

src/views/FileBrowserView/PublishFileBrowser.vue
```diff
@@ -278,6 +291,31 @@ export default {
       return publishRest.assetDownloadURI(this.identifier, this.version, asset_id);
     },
 
+    viewURI(asset_id, name) {
+      const file_ext = this.fileExtension(name);
+      if (file_ext === 'nii') {
+        return `https://bioimagesuiteweb.github.io/alphaapp/viewer.html?image=${publishRest.assetDownloadURI(this.identifier, this.version, asset_id)}`;
```
yarikoptic, line 297: Can now be
```suggestion
        return `https://bioimagesuiteweb.github.io/unstableapp/viewer.html?image=${publishRest.assetDownloadURI(this.identifier, this.version, asset_id)}`;
```
I believe for a bit more stable version

src/views/FileBrowserView/PublishFileBrowser.vue
yarikoptic: Note: we need to add maxsize field there and set it to eg 1GB or might be even less for bisweb. Sure to limitations extracted data must not exceed 4GB and anyways, download might be too slow etc. So I wouldn't bother exposing that service for known to fail out be problematic files

src/views/FileBrowserView/PublishFileBrowser.vue
```diff
@@ -278,6 +291,31 @@ export default {
...
+      }
+      if (file_ext === 'nwb') {
+        return `http://nwbexplorer.opensourcebrain.org/nwbfile=${publishRest.assetDownloadURI(this.identifier, this.version, asset_id)}`;
+      }
+
```
yarikoptic, line 302: after https://github.com/bioimagesuiteweb/bisweb/issues/129 is addressed, we should also add `.tiff` into this "registry" (also with similar limits on the file size). ATM bisweb falls into "it is a nifti" if there is no file extension (as our endpoints and blob files have none), but IMHO should work whenever content-disposition is respected (where we do provide the extension)

src/views/FileBrowserView/PublishFileBrowser.vue
```diff
@@ -278,6 +291,31 @@ export default {
...
+    fileExtension(name) {
+      let ext;
+      const name_split_ar = name.split('.');
+      if (name_split_ar[name_split_ar.length - 1] === 'nwb') {
+        ext = 'nwb';
+      } else if (name_split_ar[name_split_ar.length - 1] === 'nii' || name_split_ar[name_split_ar.length - 2] === 'nii') {
+        ext = 'nii';
+      } else {
+        ext = false;
```
mvandenburgh, line 314: While returning either a `boolean` or `string` based on the function input is a completely "legal" operation in JavaScript, I would advise against it.
1) It's not great for code readability, especially when there's places in the code that rely on this side effect. 
2) We have been moving a lot of the Vue code to TypeScript (where this is not valid), so this will eventually need to be refactored anyways.

I don't think it's necessary to parse the file extension though, and I agree with @yarikoptic's suggestion at https://github.com/datalad/datalad-deprecated/blob/master/datalad_deprecated/resources/website/assets/js/main.js#L21. It'd be really nice to have a single place where you can simply add new file extensions and regexes and have things just work.

mvandenburgh (changes requested):
Looks great, I think the button placement and icon selection is perfect.

Just for future reference, it's better to create a branch and work off of that for PRs then to fork the repo. Doing so enables [Netlify deploy previews](https://www.netlify.com/blog/2016/07/20/introducing-deploy-previews-in-netlify/), which makes testing out changes easier.

djarecka:
@mvandenburgh , @yarikoptic - thanks for your suggestions. Trying to figure out the code from datalad and wondering if we want to have a list of multiple services and multiple buttons to open the file? Or we want to provide just a default viewer for some extension?

mvandenburgh:
> @mvandenburgh , @yarikoptic - thanks for your suggestions. Trying to figure out the code from datalad and wondering if we want to have a list of multiple services and multiple buttons to open the file? Or we want to provide just a default viewer for some extensions?
> 
> **edit**
> question answered - instead of the icon, I should use a context menu to list all the possible websites

btw @djarecka not sure if you are already aware, but the vuetify docs are really good and have a bunch of different examples with the corresponding Vue/Vuetify code for you to copy/paste and modify to your liking. Looks like this example is similar to what @yarikoptic was showing today https://vuetifyjs.com/en/components/menus/#use-in-components

djarecka:
Thank you @mvandenburgh ! that's very useful!

src/views/FileBrowserView/PublishFileBrowser.vue
```diff
@@ -278,6 +344,23 @@ export default {
...
+        const regex = new RegExp(service.regex);
+        if (regex.test(name)) {
+          validServices.push({
+            name: service.name,
+            url: service.endpoint + publishRest.assetDownloadURI(ident, vers, asset_id),
+          });
+        }
+      });
+      return validServices;
+    },
```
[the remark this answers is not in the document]
mvandenburgh: ES6 [`filter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) and [`map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) functions are preferred over building up an array with `forEach` and `push`. Similarly, using [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) on line 348 is prefereed over pulling out the fields manually.

I believe [string interpolation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) (see backticks `` on line 354) is also preferred over using the `+` operator, as it eliminates any possible confusion surrounding the use of `+` for both string concatenation and addition w/ numbers.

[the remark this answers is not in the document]
yarikoptic: I am all for string interpolations, and have no clue in JS. I was [told](https://github.com/datalad/datalad-deprecated/pull/26#discussion_r666501515) that they are not supported in ES5... are we supporting ES5 ATM or could we figure out if anyone is accessing from such elderly systems if we still do?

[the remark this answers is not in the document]
mvandenburgh: We're using string interpolations in a bunch of places in the web app already. Our Babel build configuration should handle any support issues with ES5.

djarecka:
@mvandenburgh - thanks for review, I will add one more thing that @dchiquito suggested and I will create a branch from this repo, to enable CI

djarecka:
open #854 instead (created a new branch on upstream)

PR #853, opened 2021-08-24
Change order of E2E test spinup steps in CI

mvandenburgh:
**~~yarn install is being executed twice in CI, once in test-gui and once in test-e2e. The E2E tests need to run it in order to spin up an instance of the web app, so remove it from test-gui. ~~**

I also moved up the installation and running of the web app to happen before the backend django server is spun up. This means CI will immediately fail if the web app fails to start or fails linting (as opposed to waiting for the django server to start, which takes longer than the web app).

The re-enabling of the E2E tests has made CI take much longer to run and it will continue to get longer as more tests are added in the future, so I think it's worth making this optimization.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,9 @@ jobs:
     steps:
       - checkout
       - run:
+          name: Install web app
+          command: yarn install --frozen-lockfile
+      - run:
           name: Lint web app
           command: yarn run lint --no-fix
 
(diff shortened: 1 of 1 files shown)


mvandenburgh:
Oops, I guess `yarn install` is required to run linting in `test-gui`. I still think it's good to perform the web app setup before the django server though.

dchiquito (approved):

PR #857, opened 2021-08-30
Dandiset Landing Page Redesign

mvandenburgh:
This isn't ready for review yet, I'm opening a PR now so I can test this more thoroughly on the staging datasets.


waxlamp (changes requested):
Some things I've noticed that are worth fixing now rather than later (and please engage with @jtomeck for help with any of these):
- the whole app needs some padding or something--it's current crammed right to the edges of the window, to the point where it feels like some content is rolling off the edges
- when the window gets narrower, some of the content goes a bit out of spec, like the "get shareable link" button becomes too wide, and
- the "count" badges for files and MB wrap a bit funny, and
- the dates in the "other versions" listing wrap badly, even at quite large window sizes
- the display of validation errors seems to have something going on: the message in the example I'm looking at says `: Metadata version 0.4.3 is not allowed. Allowed are: 0.5.1.`. The leading colon shouldn't be there, and the trailing period is not optimal either. If these are errors in the message itself, then of course it's not something to address in this PR

The first bullet point should really be addressed now, but the others are more negotiable. If any of those are considered to be true issues, let's just file an actual issue if we're not fixing them now.

Thanks Mike, this looks really great so far!

src/views/DandisetLandingView/DandisetOwners.vue
```diff
@@ -0,0 +1,159 @@
...
+      search: null,
+      loadingUsers: false,
+      selection: null,
+      newOwners: appendResult(this.owners),
+      items: [],
+      throttledUpdate: _.debounce(this.updateItems, 200),
+    };
+  },
+  asyncComputed: {
+    async userCanModifyDandiset() {
```
dchiquito, line 103: In `DandisetLandingView` this same value is calculated using a `ref` and a `watchEffect`. If it's used in multiple components like this it's probably worth moving it to the store or going all in on the props thing and only calculating it once.

dchiquito (changes requested):
* You can add owners from any version of a dandiset, but the little `X` in the chips only show up for drafts, so you cannot delete owners from any version. I think it would be better to just hide the ownership email box when not logged in or when on a published version.
* The publish button is present and disabled when the user is not logged in. I personally think it should be removed completely when not logged in + an owner, and be greyed out with a tooltip to go to the draft version if you're on a published version + an owner.

src/views/DandisetLandingView/DandisetOwners.vue
mvandenburgh: Oh I realized after seeing that `asyncComputed` that I didn't merge the TS/Composition conversionI made to `DandisetOwners` into this branch, I did so in https://github.com/dandi/dandiarchive/pull/857/commits/1433effd00f0de2a5079679ad9a150c147cc0f8d

I moved `userCanModifyDandiset` into the store https://github.com/dandi/dandiarchive/pull/857/commits/00d82e4716a5313c81ea513049da456677803d57, it seems to fit in nicely as a getter.

dchiquito (changes requested):
Now when not logged in or when viewing published versions, ownership doesn't show up at all. I thought that was useful information, I just mean that the `X` in the chips and the email box should be hidden for non-owners. Perhaps the ownership information is not actually something that non-owners care about.

image
When the screen is too small, the version chip changes it's divider from vertical to horizontal.

src/views/DandisetLandingView/DandisetLandingView.vue
```diff
@@ -129,20 +129,9 @@ export default defineComponent({
...
-      } else if (user.value?.admin) {
-        userCanModifyDandiset.value = true;
-      } else {
-        const { data: owners } = await publishRest.owners(identifier);
-        const userExists = owners.find((owner) => owner.username === user.value?.username);
-        userCanModifyDandiset.value = !!userExists;
-      }
-    });
+    const userCanModifyDandiset: ComputedRef<boolean> = computed(
+      () => store.getters['dandiset/userCanModifyDandiset'],
```
dchiquito, line 133: My vuex is rusty, but I thought the whole point of direct-vuex was typing this kind of access, i.e.
```
const userCanModifyDandiset = store.getters.dandiset.userCanModifyDandiset;
```
https://github.com/paroi-tech/direct-vuex#use-typed-wrappers-from-outside-the-store mentions using `const store = ctx.root.$store.direct;`, perhaps that is required?

src/views/DandisetLandingView/DandisetOwners.vue
```diff
@@ -0,0 +1,172 @@
...
+    const search: Ref<string> = ref('');
+    const loadingUsers: Ref<boolean> = ref(false);
+    const selection: Ref<UserResult|null> = ref(null);
+    const items: Ref<UserResult[]> = ref([]);
+
+    const store = ctx.root.$store;
+
+    const currentDandiset: ComputedRef<Version> = computed(
+      () => store.state.dandiset.publishDandiset,
+    );
```
dchiquito, line 111:
```suggestion
    const currentDandiset = store.state.dandiset.publishDandiset;
```
I'm pretty sure direct-vuex uses all the reactive magic, so you can just grab things out of it directly and it will infer the type automatically.

dchiquito (changes requested):

src/views/DandisetLandingView/DandisetLandingView.vue
mvandenburgh: We're not using direct-vuex currently. That does look really nice, though. I'll add it and rewrite all the `store` computed vars once I address all of the other needed code changes

mvandenburgh:
> * the whole app needs some padding or something--it's current crammed right to the edges of the window, to the point where it feels like some content is rolling off the edges

Can you check the netlify preview and let me know if it's better now? Here's a screenshot comparison since the old version is inaccessible unless you checkout an older commit (first one is older ui, second is newer with additional padding):
old

new

> * when the window gets narrower, some of the content goes a bit out of spec, like the "get shareable link" button becomes too wide

The shareable link button is blocked on #865 since that button will need to be changed to a `v-menu` or something anyways, so I'll leave that one out of this PR.

> * the "count" badges for files and MB wrap a bit funny,
> * the dates in the "other versions" listing wrap badly, even at quite large window sizes

Yeah I agree it would be best to address these in this PR, I'll start looking into them next.

>     * the display of validation errors seems to have something going on: the message in the example I'm looking at says `: Metadata version 0.4.3 is not allowed. Allowed are: 0.5.1.`. The leading colon shouldn't be there, and the trailing period is not optimal either. If these are errors in the message itself, then of course it's not something to address in this PR

I suspect this is indeed an error in the message coming from the server rather then anything in the client. Which dandiset is this?

waxlamp:
> > * the whole app needs some padding or something--it's current crammed right to the edges of the window, to the point where it feels like some content is rolling off the edges
> 
> Can you check the netlify preview and let me know if it's better now?

It does look better to me now, but I was seeing something different than your before screencap, where the content was running right up into the corners and edges of the browser window.

What you have now looks good to me, but we may want to ask Jared if the newer approach damages the original intent of the design.

> > * the display of validation errors seems to have something going on: the message in the example I'm looking at says `: Metadata version 0.4.3 is not allowed. Allowed are: 0.5.1.`. The leading colon shouldn't be there, and the trailing period is not optimal either. If these are errors in the message itself, then of course it's not something to address in this PR
> 
> I suspect this is indeed an error in the message coming from the server rather then anything in the client. Which dandiset is this?

https://deploy-preview-857--gui-dandiarchive-org.netlify.app/#/dandiset/000001

mvandenburgh:
> > > * the whole app needs some padding or something--it's current crammed right to the edges of the window, to the point where it feels like some content is rolling off the edges
> > 
> > 
> > Can you check the netlify preview and let me know if it's better now?
> 
> It does look better to me now, but I was seeing something different than your before screencap, where the content was running right up into the corners and edges of the browser window.
> 
> What you have now looks good to me, but we may want to ask Jared if the newer approach damages the original intent of the design.

Just for the record in case Jared sees this - Roni and I met in person and decided to revert this change, so the padding now resembles the first screenshot (i.e. the design in the figma mockups).

djarecka:
I'm wondering if one of the link (`https://osf.io/hv7ja/`) should be clickable. Clicking on the first one expands or hides the panel, the second one (in `URL of the resource`) doesn't do anything

Screen Shot 2021-09-09 at 6 01 21 PM

djarecka:
the right panel moves to bottom of the page when the window became narrow (not even extremely narrow).
I think the ratio between width of the main panel and the width of the right panel could be smaller

Screen Shot 2021-09-09 at 6 43 52 PM

djarecka:
There is nothing under `Contact`

Screen Shot 2021-09-09 at 6 54 17 PM

waxlamp:
> I'm wondering if one of the link (`https://osf.io/hv7ja/`) should be clickable. Clicking on the first one expands or hides the panel, the second one (in `URL of the resource`) doesn't do anything
> 
> Screen Shot 2021-09-09 at 6 01 21 PM

This is one of the issues we need to address via future design efforts. In general, that panel should not just show the contents of the metadata item in a generic way like this, but instead should be a designed component that would display the information appropriately (including linkifying URLs as you suggest).

waxlamp:
> There is nothing under `Contact`
> 
> Screen Shot 2021-09-09 at 6 54 17 PM

Good catch. I notice that the validation error list does say the contact is missing (though it does so by reporting an error about regexes, which doesn't seem ideal). Additionally, the contact info should either be removed from the header, or it should say explicitly that the contact is missing.

This is a problem with the design, not this PR, so I'll file an issue to fix this later (after discussion of what exactly to do, etc.).

waxlamp:
> the right panel moves to bottom of the page when the window became narrow (not even extremely narrow).
> I think the ratio between width of the main panel and the width of the right panel could be smaller
> 
> Screen Shot 2021-09-09 at 6 43 52 PM

Let's file an issue for this, because it requires further discussion.

mvandenburgh:
RE: the remaining issues from the ones @waxlamp pointed out

> * the "count" badges for files and MB wrap a bit funny,

In light of #873, I'd like to postpone this one to a later PR.

> * the dates in the "other versions" listing wrap badly, even at quite large window sizes

I fixed this so they no longer wrap at large window sizes. The version strings are too long to avoid wrapping at small window sizes, but I changed it so they wrap nicer.

This leaves the `direct-vuex` implementation as the last TODO on this PR. I took a look at it, and it's a bit more involved then just installing the library and dropping it into the code. So, I'd like to hold off on that and make a follow up PR that solely adds direct-vuex support. Does that sound good @dchiquito?

waxlamp:
> RE: the remaining issues from the ones @waxlamp pointed out
> 
> > * the "count" badges for files and MB wrap a bit funny,
> 
> In light of #873, I'd like to postpone this one to a later PR.

👍🏾 

> 
> > * the dates in the "other versions" listing wrap badly, even at quite large window sizes
> 
> I fixed this so they no longer wrap at large window sizes. The version strings are too long to avoid wrapping at small window sizes, but I changed it so they wrap nicer.

🥳 

> 
> This leaves the `direct-vuex` implementation as the last TODO on this PR. I took a look at it, and it's a bit more involved then just installing the library and dropping it into the code. So, I'd like to hold off on that and make a follow up PR that solely adds direct-vuex support. Does that sound good @dchiquito?

Yes, I agree that this belongs in a followup, if only because this is a big change that should be handled separately from the design update encoded in this PR.

waxlamp (approved):
I think this is ready to merge.

@djarecka, please take one more look for any glaring problems (everything that wasn't fix in this PR should be in a `dlp redesign` tagged issue).

satra:
@waxlamp -   it looks like this will be merged into master, which means that we won't be able to make any changes to the current DLP till we finish the whole redesign. 

let's hold off on merging this.

waxlamp:
> @waxlamp - it looks like this will be merged into master, which means that we won't be able to make any changes to the current DLP till we finish the whole redesign.
> 
> let's hold off on merging this.

My thinking was that the DLP in this PR is at least as good as the current DLP (and in many ways, better), and that we could release it and work incrementally to get the whole redesign in. Finishing the whole redesign will force us to engage in parallel development on this branch and on `master` simultaneously, which is a thorny road to walk.

Instead, can we come up with a list of requirements for the DLP that would enable us to merge and release this branch? I am looking for ways for this branch to not become long-lived (I am already hearing the initial strains of a dirge entitled "Girder Three Migration" wafting from the horizon).

satra:
> My thinking was that the DLP in this PR is at least as good as the current DLP

it's unfortunately not. i'm happy to walk someone through the issues and concerns. here are some of them:

- no authors listed
- arbitrary decisions on what is useful for our clients
- file count and size is in the search bar, which does not display properly when size has a few digits/letters (see dandiset 20)
- font size is not appropriate for headings of side panels
- the citation link has gone to the side panel, which makes users share the url
- this DLP looks off on a phone, the current DLP does not.

i've requested jarod's permission to comment on the design itself. it needs input from scientists around what information is key. for example:
- licenses don't need a box, they can be iconified. 
- keywords: typically don't need the word keywords.

satra:
i would also suggest loading some of the more comprehensive metadata from the actual archive to test. dandiset 3, 4, 8

waxlamp:
Thanks for the comprehensive review.

I attempted to convert this back to draft status, but GitHub seems to be having some trouble at the moment.

Let's go back to the design and carry on more discussion there.

waxlamp:
> > My thinking was that the DLP in this PR is at least as good as the current DLP
> 
> it's unfortunately not. i'm happy to walk someone through the issues and concerns. here are some of them:
> 
> * no authors listed
> * arbitrary decisions on what is useful for our clients
> * file count and size is in the search bar, which does not display properly when size has a few digits/letters (see dandiset 20)
> * font size is not appropriate for headings of side panels
> * the citation link has gone to the side panel, which makes users share the url
> * this DLP looks off on a phone, the current DLP does not.
> 
> i've requested jarod's permission to comment on the design itself. it needs input from scientists around what information is key. for example:
> 
> * licenses don't need a box, they can be iconified.
> * keywords: typically don't need the word keywords.

@satra, could you outline what *has* to be fixed before we can merge this, vs. what could be fixed incrementally as we go forward? Some of these items are quite large (for example, fixing the "arbitrary decisions" we've made to be in line with scientists' use cases) and should be addressed incrementally, so long as the core functionality works properly.

satra:
@waxlamp - i think these all have to implemented.

- no authors listed
- file count and size is in the search bar, which does not display properly when size has a few digits/letters (see dandiset 20). the search bar should be left alone. 
- font size is not appropriate for headings of side panels
- the citation link has gone to the side panel, which makes users share the url
- this DLP looks off on a phone, the current DLP does not.

PR vs current



the current is not perfect, and i would remove the metadata and download related things to the side bar. but i would keep keywords right below the title and a clear share link close by. the id is not essential, but because we are trying to let people know that the id is important, it should be visible as well.

ps. in the "current" image above, the authors intentionally decided to use the institution as the author. here is one with the authors listed.

yarikoptic:
FWIW @jwodder implemented webshotting for this PR now, https://github.com/dandi/dandi-api-webshots-prs/tree/pr-857 should be getting updates shortly (cron job runs every 5 minutes if not still running from previous run, and reacts to new commit changes in this PR)

dchiquito (approved):

yarikoptic:
FWIW, confirming that after the "crash" fix, all dandisets seems to have non blank DLP: https://github.com/dandi/dandi-api-webshots-prs/tree/pr-857.

FWIW: rendering of dates for versions in a narrow window is quite suboptimal though and seemingly for no good reason:
image

satra:
@mvandenburgh - can we make sure the share icon shows up similar to the link option next to the title ?  the devices where people are likely to share from will show this links all the way at the end.

mvandenburgh:
> @mvandenburgh - can we make sure the share icon shows up similar to the link option next to the title ? the devices where people are likely to share from will show this links all the way at the end.

I did this for mobile/small screens only, is that something we should have on larger screens too?

satra:
moving the share icon only for small screens would be fine. for large screens share/permalink can stay on the side. also it may make sense to combine the permalink and social sharing under the same share icon.

Issue #728, opened 2021-06-04
directory depth error


bendichter:
![image](https://user-images.githubusercontent.com/844306/120861711-b9800900-c555-11eb-829d-491dbbe5b4b6.png)

dandiset 108 is at least 9 directories deep. When I click on the 9th "0", the website teleports me back to level 1. I am not sure if this is a level issue or maybe there are too many elements to display or some other reason.

![image](https://user-images.githubusercontent.com/844306/120861946-1d0a3680-c556-11eb-9204-eb3a99e6d72d.png)

satra:
that's a bad dataset (i stopped him on partial upload). he will fix it. i should really delete the files.

waxlamp:
I suppose the remaining question is, how did 108 get to be in a bad state without being stopped by validation or some other serverside mechanism?

@satra, if we can get an understanding thereof, and @bendichter, if you're able to fix 108 to your satisfaction, then I think we can close this issue.

satra:
there are some datasets that don't go through validation. these are datasets where we are working with the community. this was an example of someone directly uploading a zarr folder. and i had to stop early because there was a mismatch of concepts. our current backend would destroy the zarr structure, so it doesn't make sense to upload it.

waxlamp:
Ah I see. So if these datasets are currently involving personalized care from a member of our team, I think we can close this issue.

yarikoptic:
I don't think that we should dismiss this issue: even though it is a "non conventional" dataset, web UI must not just crash while traversing it.  We IMHO should take this opportunity as a lucky occasion that we managed to run into some unguarded assumption within the codebase and address it.  FWIW -- the jumping back to the top I still observe while going from https://gui.dandiarchive.org/#/dandiset/000108/draft/files?location=sub-MITU01%2Fses-20210119h14m35s15%2Fmicroscopy%2Fsub-MITU01_run-1_sample-mysample_stain-mystain_chunk-1_spim.ngff%2F0%2F0%2F0%2F0%2F0%2F to yet another `0/` but I do not see error message in console.

yarikoptic:
The underlying issue seems to be within `dandi-api` though since 
```
[...]

PR #861, opened 2021-09-02
Test publish

mvandenburgh:


dchiquito (commented):
Sadly I think there's a much less complicated way to get publishable data into the API:
```
      - run:
          name: Create dev dandiset
          command: python manage.py migrate && python manage.py createsuperuser ... && python manage.py create_dev_dandiset --owner super@user.com
          working_directory: dandi-api
          environment:
            DJANGO_DATABASE_URL: postgres://postgres:postgres@localhost:5432/django
```

Then the test can just log in with the new super user credentials and they should be able to click publish on dandiset `000001` immediately.

mvandenburgh:
> Sadly I think there's a much less complicated way to get publishable data into the API:
> 
> ```
>       - run:
>           name: Create dev dandiset
>           command: python manage.py migrate && python manage.py createsuperuser ... && python manage.py create_dev_dandiset --owner super@user.com
>           working_directory: dandi-api
>           environment:
>             DJANGO_DATABASE_URL: postgres://postgres:postgres@localhost:5432/django
> ```
> 
> Then the test can just log in with the new super user credentials and they should be able to click publish on dandiset `000001` immediately.

I thought about doing this but ran into some issues:

- How would we run this locally? (maybe the answer is we don't run these types of tests locally? or we could make the user run that management command before running the tests I suppose)
- We can't guarentee that the test dandiset has an id of `000001` because the other tests also create dandisets, and there's [no way to control the order that tests are executed in Jest](https://github.com/facebook/jest/issues/4032#issuecomment-315210901)
  - We could do some tricks with environment variables to tell the tests which dandiset is the test dandiset, but this limits us to just this one set of assets (see next bullet point)
- I think it'd be nice to have a `test_data` directory where we can have different types of assets (invalid assets, assets with large directory depths as mentioned in https://github.com/dandi/dandiarchive/issues/728, etc.) that can be used in testing.
  - Maybe the answer to this is that it's not worth pursuing these kind of tests

dchiquito:
1.  Good point. We could perhaps augment the script to specify a dandiset ID to replace if it already exists, but the test user is a slightly trickier proposition. Perhaps a new command `setup_test_data` that destructively builds a test environment would be better?
1.  I think in this case it would be guaranteed to be 000001, because that step is run before the tests begin executing. This would also be solved by a destructive `setup_test_data` script.
1. Agreed, we need some kind of facility for providing data of various flavors. Right now the `create_dev_dandiset` script uses gibberish as the asset contents, but I think some kind of `test_data` could be used instead.

I think I'm talking myself into a `setup_test_data` command that destroys any existing conflicting data and ensures that some body of test data is always present. It would be run before the integration test in CI and devs would have to run it before running integration tests to ensure the environment is ripe for testing.

It would mean that we can't run the test suite in prod/staging, although I think it's better to be able to run it on demand against an arbitrary branch.

mvandenburgh:
Having a `setup_test_data` function would also work, but then we would need to designate certain dandisets as certain flavors. Like we would need to hardcode (or use env var) somewhere that dandiset 000001 is guaranteed to be a certain flavor, 00002 is another, etc. And this would have to be synchronized across two different repos, since the tests are in the `dandiarchive` repo and the `setup_test_data` command would be in the `dandi-api` repo. 

The way I'm doing it in this PR introduces some complexity, but all of that complexity is wrapped in the `uploadAssets()` function. Thinking ahead, all the user has to do to add new assets is drop them in `test_data` (with some slight modifications to `dandiset.yaml`) and call `uploadAssets()` on it during the test. On the other hand if we used the `setup_test_data` function, adding new test data would require making an additional PR to dandi-api to modify the `setup_test_data` command. I'm seeing it as a tradeoff between
- having a simpler way to add assets in testing that makes it more of a burden to add future tests
vs
-  having a more complex way to add assets in testing that makes it easier to add future tests

What do you think?

dchiquito:
That is an excellent summation of the tradeoffs.

In general, I think that maintenance burden is the biggest obstacle to browser testing, since it's sort of like the cherry balanced on top of a house of cards. Anything that can reduce that burden is good.

I agree that the upload.sh method you have here makes it much easier to just drop a new asset in and add it as needed to only the tests that need it. It's nice that the test starts from a blank slate and explicitly sets up everything it needs right there. However, it involves a long pipeline of things that might break: fetching API key -> installing dandi-cli -> invoking a shell script -> invoking dandi-cli to upload the asset. There's also the bit about remembering to update `dandiset.yml` whenever a new asset is added, which is kinda hard to document and remember to do.

Additionally, this only solves uploading assets; as the app grows, we will need to provision more things and invent similar pipelines (admin users, zarr files, embargo'd dandisets, for example). The maintenance burden will only grow.


On the other hand, we have the `setup_test_data` solution. Basically, we would define "blank slate" to include some baseline of existing data. It has drawbacks, mostly that we have to separate the tests from the test data. It does suck to work with two repositories to set up test data. Maintaining the script itself will also be burdensome, as it will have to do some weird DB stuff, and it will presumably get bigger and more complicated as we add more tests and need more data. And resetting the DB prior to every run would involve an extra command before every `yarn run test`.

However, we would have the freedom to put the DB into any state that we wanted prior to the test running. Any simple test data (dandisets, new users, maybe metadata?) could be added through the UI, and we could save the script for the really gross, difficult data.

My value judgement leans toward `setup_test_data` as more maintainable long term.

dchiquito:
incidentally `setup_test_data` would also resolve https://github.com/dandi/dandi-api/issues/495 by simply provisioning users with predefined names.

mvandenburgh:
> In general, I think that maintenance burden is the biggest obstacle to browser testing, since it's sort of like the cherry balanced on top of a house of cards. Anything that can reduce that burden is good.

Good analogy, can confirm after working on these tests the last few weeks :smile: 

> Additionally, this only solves uploading assets; as the app grows, we will need to provision more things and invent similar pipelines (admin users, zarr files, embargo'd dandisets, for example). The maintenance burden will only grow.
> 
> On the other hand, we have the `setup_test_data` solution. Basically, we would define "blank slate" to include some baseline of existing data. It has drawbacks, mostly that we have to separate the tests from the test data. It does suck to work with two repositories to set up test data. Maintaining the script itself will also be burdensome, as it will have to do some weird DB stuff, and it will presumably get bigger and more complicated as we add more tests and need more data. And resetting the DB prior to every run would involve an extra command before every `yarn run test`.

I spent some time thinking about this, and I agree with your `setup_test_data` solution. 

I agree with your point about maintainability - the `setup_test_data` solution is guaranteed to continue working for any future features (embargo + others) since it's executing within Django, while the node execSync/dandi-cli method will probably have to be expanded or rewritten to accommodate changes to the upload process. We would also have to start worrying about changes to dandi-cli, especially since we're using dev-only options which I assume are subject to change.

> However, we would have the freedom to put the DB into any state that we wanted prior to the test running.

Yeah I agree that this is huge, and would allow us to write more focused tests.

re: having to work across both repos, the E2E tests have access to the contents of the dandi-api repo prior to running, so we could maybe do some environment variable trickery to keep test data in the dandi-api repo and avoid having to keep it "synced" across both repos. Or maybe not, either way I'm okay with the `setup_test_data` solution.

dchiquito:
@mvandenburgh are you planning on a new PR with the alternative test data scheme? If so there were some frontend fixes that could be cherrypicked out of this PR.

mvandenburgh:
> @mvandenburgh are you planning on a new PR with the alternative test data scheme? If so there were some frontend fixes that could be cherrypicked out of this PR.

Yes, I was planning on a new PR in dand-api for this. All of the frontend fixes here should also be present in the `landing-page-redesign` branch so I don't think we need to do anything here.

PR #862, opened 2021-09-02
Add a package.json script for auto-generating metadata typings from json schema

mvandenburgh:
One of the issues I've run into with the DLP redesign is that dandiset metadata has no typings (obviously, since that would mean having to maintain and change them every time there's a schema change), so the benefits of TypeScript are mostly lost for a pretty significant part of the redesign. I tried to come up with a solution to this in this PR.

I found [this nice package](https://www.npmjs.com/package/json-schema-to-typescript) that converts JSON schemas to TypeScript typings; using this, we can generate typings for any metadata schema version. I tried out the typings it generated for schema 0.5.1 while doing some of the DLP redesign stuff and it seems to work great. This PR adds a yarn script to automatically download and generate typings for a given schema version.

This does introduce some overhead for developers in that someone has to run the script whenever a new schema version is released, but I think it's well worth it considering that we'll get all of the benefits of TypeScript in any metadata-related code (including catching any breaking schema changes at compile time :astonished: )


yarikoptic:
Didn't look in detail, but can that script
- be ran by pre commit
- CI run it too, and if it introduced changes, means that developer didn't run it, or ran using different/outdated version of dandischema - error out

waxlamp:
> Didn't look in detail, but can that script
> 
> * be ran by pre commit
> * CI run it too, and if it introduced changes, means that developer didn't run it, or ran using different/outdated version of dandischema - error out

A CI job to run this script and fail or warn if the results introduce changes is a good idea, but I don't think we should immediately enforce the latest schema version as a requirement on all PRs (at least, to start).

Is there a reason to believe that any PR that doesn't work with the latest schema version is flawed from the start?

waxlamp (approved):
This is AWESOME. Thanks for putting this together, Mike!

dchiquito (approved):
This looks really cool!

Would it be appropriate to move this functionality to a new node package in https://github.com/dandi/schema and simply import the package+types in dandiarchive? Then the types would be generated automatically by CI in `schema`, and updating the schema version would be as simple as updating the dependency.

I'm approving because that sounds like more work and this solution looks good for how we're currently using the schema. This way we also get a nice little git diff in the migration PR.

mvandenburgh:
> Would it be appropriate to move this functionality to a new node package in https://github.com/dandi/schema and simply import the package+types in dandiarchive? Then the types would be generated automatically by CI in `schema`, and updating the schema version would be as simple as updating the dependency.

I think that sounds like a great idea. I'll merge this PR for now since it will be immediately helpful to me as I work on more of the DLP redesign, but I vote that we do this at some point. There's even a [json-schema-to-typescript CLI](https://github.com/bcherny/json-schema-to-typescript#cli) that should make generating the types in CI easy.

yarikoptic:
Great idea indeed.  If I got it right -- this would be the actual workflow https://github.com/dandi/dandischema/blob/master/.github/workflows/release.yml of `dandischema` which produces a release within `schema` repository.  Should be trivial to extend and we could populate those files if needed for already existing release there.

PR #881, opened 2021-09-14
Use direct-vuex

### LINKED ISSUE #518: Switch from using Vuex to Direct-vuex
This depends on #515, and should be addressed separately from that issue.

mvandenburgh:
- Add direct-vuex as a dependency and configure the store to use it
- Update all components to use direct-vuex
- Convert `DownloadDialog.vue` to TypeScript/Composition API

Closes #518


src/views/FileBrowserView/PublishFileBrowser.vue
```diff
@@ -271,7 +270,9 @@ export default {
       return this.me && this.owners.includes(this.me);
     },
 
-    ...mapState('dandiset', ['publishDandiset']),
+    publishDandiset() {
+      return this.$store.direct.state.dandiset.publishDandiset;
```
AlmightyYakob, line 274: Same question as in `FileBrowser.vue`.

AlmightyYakob (changes requested):
Looks good! A couple questions

src/components/DandisetStats.vue
```diff
@@ -41,12 +41,8 @@ interface DandisetStats {
...
 export default defineComponent({
   name: 'DandisetStats',
-  setup(props, ctx) {
-    const store = ctx.root.$store;
-
-    const currentDandiset: ComputedRef<Version> = computed(
-      () => store.state.dandiset.publishDandiset,
-    );
+  setup() {
+    const currentDandiset = computed(() => store.state.dandiset.publishDandiset);
```
dchiquito, line 45:
```suggestion
    const currentDandiset = store.state.dandiset.publishDandiset;
```
I'm pretty sure this works and saves some boilerplate.

src/views/DandisetLandingView/CiteAsDialog.vue
```diff
@@ -23,28 +23,17 @@
...
-    const currentDandiset: ComputedRef<Version> = computed(
-      () => store.state.dandiset.publishDandiset,
-    );
-
-    const citation: ComputedRef<string|undefined> = computed(
-      () => currentDandiset.value.metadata?.citation,
-    );
-
+  setup() {
+    const citation = computed(() => store.state.dandiset.publishDandiset?.metadata?.citation);
```
dchiquito, line 36:
```suggestion
    const citation = store.state.dandiset.publishDandiset?.metadata?.citation;
```
same as above

src/components/DandisetStats.vue
waxlamp: Interesting--where does the reactivity come from?

src/views/FileBrowserView/FileBrowser.vue
```diff
@@ -36,7 +35,9 @@ export default {
     };
   },
   computed: {
-    ...mapState('dandiset', ['publishDandiset']),
+    publishDandiset() {
+      return this.$store.direct.state.dandiset.publishDandiset;
```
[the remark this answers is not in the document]
mvandenburgh: Nope you're right, I changed it. For some reason I thought importing it directly would only work with composition api

src/components/DandisetStats.vue
mvandenburgh: This doesn't seem to work, the variable loses its reactivity if it's not wrapped with `computed`

dchiquito: Lame, I assumed that vuex stuff would be reactive, but I guess they don't want direct state mutations?

AlmightyYakob: To provide more context, I believe if you were to try to access the state directly in normal `vuex`, it also wouldn't be reactive. This is why calls to `mapState` are always placed in the `computed` section of a standard Vue 2 component. So from that standpoint, this is still a computed var, and all that's changed is how it's declared.

AlmightyYakob (approved):
LGTM

src/components/DandisetStats.vue
waxlamp: Right, the point of VueX is to make state transitions pure and explicit. I'm actually not sure what happens if you try to make an assignment to the state in direct-vuex...

AlmightyYakob: Since `direct-vuex` is mostly a type wrapper around `vuex`, I'd expect it to fail in the same way that `vuex` does, by raising an error at runtime, but I'm not actually sure.

PR #894, opened 2021-09-23
Test PR to see how forks affect github actions

waxlamp:
DO NOT MERGE


mvandenburgh:
Looks like the CircleCI check is still showing up :frowning_face:

waxlamp:
> Looks like the CircleCI check is still showing up

I went into the repo settings and "suspended" the CircleCI integration (but didn't uninstall it, though we can do that too). That was after I saw it show up here, so the trick will be to launch another PR and see if it's included there too. Stand by.

PR #897, opened 2021-09-24
make funding section mandatory

### LINKED ISSUE #876: Make funding section mandatory
The funding section is present but it's just that people aren’t populating it. So we may want to make that section mandatory and say “No funding information available”.

sanuann:
closes #876


src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -258,7 +258,10 @@ export default defineComponent({
       const remove_list = ['citation', 'repository', 'url', 'schemaVersion', 'version', 'id', 'keywords', 'schemaKey'];
       extra = extra.filter((n) => !remove_list.includes(n));
       const extra_obj: any = extra.reduce((obj, key) => ({ ...obj, [key]: meta[key] }), {});
-      extra_obj.contributor = _.filter(meta.contributor, (author) => author.schemaKey !== 'Person');
+      const contributorList = _.filter(meta.contributor, (author) => author.schemaKey !== 'Person');
+      if (contributorList.length) {
+        extra_obj.contributor = contributorList;
+      } else extra_obj.contributor = 'No funding information available';
```
waxlamp, line 264: I think there is a better way to implement the logic. It looks like you want to display a message ("no funding information available") if there are no `Person` objects in the contributor list. I think I would just keep the old logic in this block, and use the `v-if` construct in your other change in this PR to look at whether `data` is a length-0 array.

With this approach, you are mixing rendering information (the error message) into the same slot that normally holds the list of contributors. This is confusing to understand, and it may cause issues in other parts of the code where the list of contributors is needed.

Ideally, we need a dandiset that tests both situations: one with contributors, and one without. You can create such dandisets by hand, or identify existing ones that would work here, but this is information that could definitely go into the PR description.

Finally, I don't understand what the contributor list has to do with funding sources. This is another aspect of the PR that could use some explanation in the PR description.

Let me know if you'd like to discuss this in person.

waxlamp: One other comment here about code style: the way you have the `else` clause written can be improved. Generally, you want to enclose all parts of if-else blocks in curly brackets, even if one of those parts only has one line. This avoids problems of parsing in case you accidentally add code to that part without adding the brackets, and it also falls in line with current best practices for JavaScript.

waxlamp (commented):
Thank you for working on this, Sanu. I left a lengthy comment on the main part of the logic that you added.

mvandenburgh:
@sanuann fyi some of the dandisets in the netlify preview (like [this one](https://deploy-preview-897--gui-dandiarchive-org.netlify.app/#/dandiset/100268)) crash on the landing page due to a bug that I recently fixed in #899, so I would recommend rebasing/merging `master`.

waxlamp (approved):

PR #901, opened 2021-09-27
Convert `AppBar` components to TypeScript/Composition API

mvandenburgh:
Converts all components in `components/AppBar/` to TS/Composition API.

### DIFF
--- a/src/components/AppBar/AppBar.vue
+++ b/src/components/AppBar/AppBar.vue
@@ -120,67 +120,74 @@
   </v-app-bar>
 </template>
 
-<script>
+<script lang="ts">
+import { computed, defineComponent } from '@vue/composition-api';
+
 import {
-  cookiesEnabled, loggedIn, insideIFrame, publishRest,
+  cookiesEnabled as cookiesEnabledFunc,
+  loggedIn as loggedInFunc,
+  insideIFrame as insideIFrameFunc,
+  publishRest,
 } from '@/rest';
 import { dandiAboutUrl, dandiDocumentationUrl, dandiHelpUrl } from '@/utils/constants';
 import UserMenu from '@/components/AppBar/UserMenu.vue';
 
-export default {
+export default defineComponent({
   name: 'AppBar',
-  components: {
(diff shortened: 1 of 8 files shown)


src/components/AppBar/AppBar.vue
```diff
@@ -120,67 +120,74 @@
...
       return JSON.stringify({ name, query, params });
-    },
-  },
-  methods: {
-    login() {
+    });
+
+    const cookiesEnabled = computed(cookiesEnabledFunc);
+    const loggedIn = computed(loggedInFunc);
+    const insideIFrame = computed(insideIFrameFunc);
```
dchiquito, line 146: Probably out of scope for this PR, but we could probably just import `computed` in `rest.ts` and define these as `computed(() => ...)` instead of `() => ...`, which saves you having to do this song and dance in every component that needs to use them.

src/components/AppBar/UserMenu.vue
```diff
@@ -60,19 +58,18 @@ export default {
             );
           }
         }
-        // If first name + last name aren't specified, try to use the login instead
-        const { login } = this.user;
-        if (login) {
-          return login.slice(0, 2);
-        }
```
dchiquito, line 67: Is there a reason you're axing this? I'm not morally opposed to just having `??` for users with no name defined in GH, but I just want to make sure the decision is deliberate.

dchiquito (approved):
Only minor quibbles, I'd be OK merging this now.

src/components/AppBar/AppBar.vue
mvandenburgh: I agree that would be better. I'll leave that for a future PR since that would affect a lot of other components too.

mvandenburgh: Filed #909 for this.

src/components/AppBar/UserMenu.vue
dchiquito: got me there, I thought we were still providing that. Exterminate

PR #902, opened 2021-09-27
Convert `CreateDandiset` and `StatsBar` to TypeScript/Composition API

mvandenburgh:


src/views/HomeView/StatsBar.vue
```diff
@@ -31,40 +31,42 @@
...
+      { name: 'users', value: users.value },
+      { name: 'total data size', value: filesize(size.value, { round: 0, base: 10, standard: 'iec' }) },
+    ]);
+
+    // equivalent of async created method in options API
+    publishRest.stats().then((data) => {
+      dandisets.value = data.dandiset_count;
+      users.value = data.user_count;
+      size.value = data.size;
+    });
```
dchiquito, line 65: Do you prefer the promise chaining syntax? Personally I generally prefer `await`, although I admit it's pretty ugly in this case:
```
const { dandiset_count, user_count, size } = await publishRest.stats();
dandisets.value = dandiset_count;
users.value = user_count;
size.value = size;
```

dchiquito (approved):
Only minor quibbles, LGTM

src/views/HomeView/StatsBar.vue
mvandenburgh: Yeah I generally prefer `async/await` too. In this case I used `.then()` because I think it makes it more obvious that those 4 lines are performing a single operation together, and separates them from the rest of the `setup()` boilerplate.

Issue #903, opened 2021-09-27
Remove "Publish API" terminology from codebase


mvandenburgh:
I've come to realize that the use of "publish" as a prefix to variables/functions like `publishDandiset`, `publishRest`, etc, was originally done to distinguish things from their Girder equivalents (`girderDandiset`, `girderRest`) back when the Django API was only being used to handle published dandisets. This is a potential source of confusion now that the Django API has completely replaced Girder, especially for someone new to the codebase. 

These names should be changed to something like `dandiRest`, `dandiset`, etc.

PR #904, opened 2021-09-27
Convert `FileBrowser` to TS/Composition API

mvandenburgh:
Also merges the `PublishFileBrowser` component into `FileBrowser.vue` as part of #903.

I also changed the CI config so tests will run on any branch, not just `master` or branches that have PRs associated with them; GitHub Actions minutes are unlimited for public repositories so I don't see why not.

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -63,3 +63,26 @@ export interface DandisetStats {
   asset_count: number,
   size: number,
 }
+
+export interface AssetStats {
+  folder: boolean,
+  name: string,
+  created: string|undefined,
+  modified: string|undefined,
+  size: number|undefined,
+}
+
+export interface AssetFolder extends AssetStats {
+  num_files: number|undefined,
+}
+
+export interface AssetServices {
+  name: string,
+  url: string,
+}
+
+export interface AssetFile extends AssetStats {
+  asset_id: string,
+  path: string,
+  services: AssetServices[],
+}
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -67,13 +67,13 @@ export interface DandisetStats {
 export interface AssetStats {
   folder: boolean,
   name: string,
-  created: string|undefined,
-  modified: string|undefined,
-  size: number|undefined,
+  created?: string,
+  modified?: string,
+  size?: number,
 }
 
 export interface AssetFolder extends AssetStats {
-  num_files: number|undefined,
+  num_files?: number,
 }
 
 export interface AssetServices {
--- a/src/views/FileBrowserView/FileBrowser.vue
+++ b/src/views/FileBrowserView/FileBrowser.vue
@@ -298,6 +298,8 @@ export default defineComponent({
       ];
     }
 
+    // Force the watcher  to run synchronously to prevent it from triggering itself infinitely
+    // when it modifies `updating`. If we were using Vue 3, we'd use watchSyncEffect instead
     watchEffect(async () => {
       updating.value = true;
       await getAssets();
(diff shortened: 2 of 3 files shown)


src/rest/publish.ts
```diff
@@ -99,7 +99,8 @@ const publishRest = new Vue({
         throw error;
       }
     },
-    async assetPaths(identifier: string, version: string, location: string): Promise<string[]> {
+    // TODO: define return type here
+    async assetPaths(identifier: string, version: string, location: string): Promise<any> {
```
dchiquito, line 103:
```suggestion
    async assetPaths(identifier: string, version: string, location: string): Promise<{ folders: AssetFolder[], files: AssetFile[] }> {
```
I'm not sure it's worth it's own type, so using a custom type definition seems like the easy way out.
You probably need to import those types in this file though.

src/types/index.ts
```diff
@@ -63,3 +63,26 @@ export interface DandisetStats {
   asset_count: number,
   size: number,
 }
+
+export interface AssetStats {
+  folder: boolean,
+  name: string,
+  created: string|undefined,
+  modified: string|undefined,
+  size: number|undefined,
```
dchiquito, line 72:
```suggestion
  created: string?,
  modified: string?,
  size: number?,
```
This is the convenient syntax for marking something as optional, which is the same as `whatever|undefined`.

src/views/FileBrowserView/FileBrowser.vue
```diff
@@ -204,15 +204,21 @@
...
 import { publishRest } from '@/rest';
 import store from '@/store';
+import { draftVersion } from '@/utils/constants';
+import { AssetStats } from '@/types';
 
 const parentDirectory = '..';
 const rootDirectory = '';
 
-const sortByName = (a, b) => {
+const sortByName = (a: any, b: any) => {
```
dchiquito, line 221: Aren't these `AssetStats`?

src/views/FileBrowserView/FileBrowser.vue
```diff
@@ -30,30 +256,123 @@ export default {
...
+
+    const currentDandiset = computed(() => store.state.dandiset.publishDandiset);
+    const splitLocation = computed(() => location.value.split('/'));
+
+    const items: Ref<AssetStats[]|null> = ref(null);
+
+    async function getAssets() {
+      const { version, identifier } = props;
+
+      const data = await publishRest.assetPaths(identifier, version, location.value);
```
dchiquito, line 273:
```suggestion
      const { files, folders } = await publishRest.assetPaths(identifier, version, location.value);
```

src/views/FileBrowserView/FileBrowser.vue
```diff
@@ -30,30 +256,123 @@ export default {
...
     }
-  },
-  methods: {
-    fetchPublishDandiset() {
+
+    watchEffect(async () => {
+      updating.value = true;
+      await getAssets();
+      updating.value = false;
+    }, { flush: 'sync' });
```
dchiquito, line 305: What does `flush: 'sync'` mean? Never seen that before

dchiquito (changes requested):
Some nitpicks

src/types/index.ts
mvandenburgh: https://github.com/dandi/dandiarchive/pull/904/commits/77269760e88c29e960c08158ba518fdb104b711d

mvandenburgh:
@dchiquito this should be all set for you to review again, PTAL. FYI to save you some time, the only code that changed since your last review is the `assetPaths()` return type in `rest.ts` and the contents of the `<script>` section of `FileBrowser.vue`. Mentioning this b/c I removed a redundant top-level `<div>` in that component and now the diff view is showing the entire template as new code.

dchiquito (approved):

PR #906, opened 2021-09-27
Remove references to publish API

### LINKED ISSUE #903: Remove "Publish API" terminology from codebase

mvandenburgh:
Closes #903.


src/rest.ts
```diff
@@ -212,4 +212,15 @@ client.interceptors.request.use((config) => ({
   },
 }));
 
-export default publishRest;
+const user = () => dandiRest.user;
+const loggedIn = () => !!user();
+const insideIFrame = (): boolean => window.self !== window.top;
+const cookiesEnabled = (): boolean => navigator.cookieEnabled;
```
dchiquito, line 218: Is https://github.com/dandi/dandiarchive/issues/909 going to work without these exports?

dchiquito (approved):

src/rest.ts
mvandenburgh: They're still being exported, see lines 220-225.

PR #914, opened 2021-09-29
Fix contributor list filtering

### LINKED ISSUE #912: erroneous entity listed in contributor section
for this dandiset (https://dandiarchive.org/dandiset/000026/draft), the last person listed in contributors has `false` for `includeInCitation`

mvandenburgh:
This bit of code somehow got lost when I implemented the redesign. I think there's a lot of improvements that can be made to the code (to make it work with TypeScript, but also in general), but I don't think it's worth spending time on because this component will go away in the near future anyways.

Fixes #912

### DIFF
--- a/src/views/DandisetLandingView/DandisetContributors.vue
+++ b/src/views/DandisetLandingView/DandisetContributors.vue
@@ -1,3 +1,7 @@
+/*####################################################################################
+# TODO: This is just a temporary stopgap until the contributors list is implemented  #
+# as seen in the DLP redesign. This component should be removed at that point.       #
+####################################################################################*/
 <template>
   <div>
     <v-row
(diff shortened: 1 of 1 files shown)


src/views/DandisetLandingView/DandisetContributors.vue
```diff
@@ -42,16 +46,38 @@
...
+          let orcid_id = author.identifier;
+          if ((author.affiliation as any)?.length) {
+            affiliations = (author.affiliation as any).map((a: any) => a.name).join(', ');
+          }
+          let author_name = author.name;
+          if (index < persons.length - 1) {
+            author_name = `${author.name};`;
+          }
+          if (orcid_id) {
+            orcid_id = `https://orcid.org/${orcid_id}`;
```
satra, line 74: if this is being done then the template `:href` that displays it needs to change back

PR #915, opened 2021-09-29
fix: revert change to remove double orcid

satra:
now that the contributor code is back. there is a double orcid. this revert step should fix it.


mvandenburgh (approved):
Ah that explains where the orcid issue came from

PR #918, opened 2021-09-30
fix:  filesize transform in DLP

satra:
this matches what is shown in the dandiset listings and makes the value be in a 10^N power rather than 2^N power


satra:
thank you @mvandenburgh

PR #920, opened 2021-10-02
adding no contact option

### LINKED ISSUE #878: Contact should show something to indicate missing contact info
> There is nothing under `Contact`
> 
> Screen Shot 2021-09-09 at 6 54 17 PM

Good catch. I notice that the validation error list does say the contact is missing (though it does so by reporting an error about regexes, which doesn't seem ideal). Additionally, the contact info should either be removed from the header, or it should say explicitly that the contact is missing.

This is a problem with the design, not this PR, so I'll file an issue to fix this later (after discussion of what exactly to do, etc.).

_Originally posted by @waxlamp in https://github.com/dandi/dandiarchive/issues/857#issuecomment-916928985_

djarecka:
fixes #878

### DIFF
--- a/src/views/DandisetLandingView/DandisetMain.vue
+++ b/src/views/DandisetLandingView/DandisetMain.vue
@@ -39,7 +39,14 @@
         <v-col :cols="$vuetify.breakpoint.xs ? 12 : 3">
           <span>
             <v-icon class="grey--text text--lighten-1">mdi-account</v-icon>
-            Contact <strong>{{ currentDandiset.contact_person }}</strong>
+            <template
+              v-if="['', ','].includes(currentDandiset.contact_person.trim())"
+            >
+              No contact information
+            </template>
+            <template v-else>
+              Contact <strong>{{ currentDandiset.contact_person }}</strong>
+            </template>
           </span>
         </v-col>
         <v-col :cols="$vuetify.breakpoint.xs ? 12 : 3">


src/views/DandisetLandingView/DandisetMain.vue
```diff
@@ -39,7 +39,14 @@
         <v-col :cols="$vuetify.breakpoint.xs ? 12 : 3">
           <span>
             <v-icon class="grey--text text--lighten-1">mdi-account</v-icon>
-            Contact <strong>{{ currentDandiset.contact_person }}</strong>
+            <template
+              v-if="['', ','].includes(currentDandiset.contact_person.trim())"
```
mvandenburgh, line 43: When a dandiset has a contact person with the name `','` that indicates that there is a contact person associated with it, but they do not have a first and last name. This happens when a user doesn't have a name set in their GitHub account. I wonder if it'd be better to fall back to their email if a name isn't set?

waxlamp: What's the modern guidance on publishing people's email addresses? I've been allergic to it since 2003, on the assumption that evil bots are slurping up email addresses for spam. I wonder if the GitHub username would be a better fallback, and I wonder further if during the verification process (i.e., the user questionnaire flow) should require providing a name as well.

djarecka: ok, I will try to explore how to get the username

mvandenburgh: > ok, I will try to explore how to get the username

:+1: 
just as a heads up, getting the github username would be something that happens on the API server (dandi-api). So, I think this PR could just be changed like this:

```suggestion
              v-if="!currentDandiset.contact_person"
```

And then it could be merged

djarecka: but assuming that `dandi-api` will return an empty string if there is no name, right now it returns ", ", so the if statement will always give `false`

Issue #865, opened 2021-09-03
Dropdowns in DLP should not behave like material design buttons


waxlamp:
The "download", "shareable link", and "cite as" dropdowns behave like they are regular buttons with attached popovers. Specifically, they show the material design "bloop" effect when clicking on them, and there's a slight delay before the menu portion appears.

If this is standard vuetify behavior for dropdowns, then this can probably be closed. But it would be better to not have the buttons boop or bloop, and show the menu immediately--then it would feel like a real dropdown.

djarecka:
@waxlamp - I've changed slightly the button: removing "down arrows", but I'm not sure about "bloop"...

PR #921, opened 2021-10-02
changing download/cite buttons

djarecka:
changing buttons for `download` and `cite as`: removing "down arrows" 

partially addressing #865

### DIFF
--- a/src/views/DandisetLandingView/DandisetActions.vue
+++ b/src/views/DandisetLandingView/DandisetActions.vue
@@ -31,9 +31,6 @@
               </v-icon>
               Download
               <v-spacer />
-              <v-icon right>
-                mdi-chevron-down
-              </v-icon>
             </v-btn>
           </template>
         </DownloadDialog>
@@ -58,9 +55,6 @@
               </v-icon>
               Cite As
               <v-spacer />
-              <v-icon right>
-                mdi-chevron-down
-              </v-icon>
             </v-btn>
           </template>
         </CiteAsDialog>


mvandenburgh:
I looked more into this "bloop" issue, and I'm starting to think that this issue isn't actually an issue at all, but rather is in line with Vuetify's intended look/behavior. When implementing a standard dropdown with a `v-menu` there is a "bloop" upon clicking the button to open the dropdown in [this example](https://vuetifyjs.com/en/components/menus/#usage) from the Vuetify docs, just like our Download and Cite buttons.

So I'm thinking we should leave the buttons as they are, down arrows included. 

On a separate note, I think it'd be good to apply the [`slide-y-transition` Vuetify prop](https://vuetifyjs.com/en/styles/transitions/#slide-y) to make the menu behave more like a dropdown.

@djarecka @waxlamp what do you think?

waxlamp:
> I looked more into this "bloop" issue, and I'm starting to think that this issue isn't actually an issue at all, but rather is in line with Vuetify's intended look/behavior. When implementing a standard dropdown with a `v-menu` there is a "bloop" upon clicking the button to open the dropdown in [this example](https://vuetifyjs.com/en/components/menus/#usage) from the Vuetify docs, just like our Download and Cite buttons.
> 
> So I'm thinking we should leave the buttons as they are, down arrows included.
> 
> On a separate note, I think it'd be good to apply the [`slide-y-transition` Vuetify prop](https://vuetifyjs.com/en/styles/transitions/#slide-y) to make the menu behave more like a dropdown.
> 
> @djarecka @waxlamp what do you think?

I think you're right Mike, and I am just tired of the material design era. I really don't like the artificial delay built in to v-menu, but since that's standard we should just roll with it.

@djarecka, if you agree, I would revert the arrow change, add the slide-y-transition prop instead, and we can review that atlernate change instead.

waxlamp:
> @mvandenburgh - I've been trying to add `transition="slide-y-transition"` or `<v-slide-y-transition>` in various place, but I'm failing to change the behavior. Any suggestion where this should be added?

If you look at the doc page Mike linked, you'll see some examples of how to use the prop in real code (click on the `< >` button on any example to see the code).

djarecka:
yes, I saw the examples, and I even tried to add `<v-menu>`, but the button disappeared.

waxlamp:
Closing this as not a needed change. Thanks for your effort on this wild goose chase, @djarecka.

PR #930, opened 2021-10-07
Revert File browser TS/Composition API conversion

mvandenburgh:
This reverts commit a8d5f0d537c945cee0fb9792721fdc4f34698ff0, reversing
changes made to 9c32917a9d6f9503bc4832ac60675faf32af52ce.

There's been a lot of subtle bugs cropping up that came from the TS/Composition API rewrite I did for the FileBrowser component. I need to spend some time rethinking how to do it, and I don't want to leave the file browser in a broken state in the meantime.

### DIFF
--- a/src/rest.ts
+++ b/src/rest.ts
@@ -2,7 +2,7 @@ import axios, { AxiosInstance, AxiosRequestConfig, AxiosResponse } from 'axios';
 import Vue from 'vue';
 import OAuthClient from '@girder/oauth-client';
 import {
-  Asset, Dandiset, Paginated, User, Version, Info, AssetFile, AssetFolder,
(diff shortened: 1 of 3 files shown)


yarikoptic:
It would have been nice if the motivation for the change/revert was mentioned in the PR description.

mvandenburgh:
> Thank you @mvandenburgh ! I think we should start adding some rudimentary "functionality unittests" on web UI so such RFs could be done in a more "informed" fashion.

There's a plan to extend the existing E2E tests to support things like the file browser, which I thought there was an issue filed for already, but I guess there wasn't - filed at #936. Once implemented these tests should prevent stuff like this from happening again

PR #935, opened 2021-10-14
changing the share icon

### LINKED ISSUE #913: share icon panel changes
- list of contributors not necessary, just the title is fine. 
- Change "share this dandiset" to "Share this dandiset with this link"
- For the twitter message we should refer to `@DANDIarchive` rather than `#dandi`

djarecka:
fixes #913

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -2,6 +2,7 @@
   <v-app class="dandi-app">
     <AppBar />
     <v-main>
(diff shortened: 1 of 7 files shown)


djarecka:
@dchiquito - I believe should be better now

satra:
twitter looks good. could we also implement the other comment i had in the issue to use the share link in all viewports?

PR #948, opened 2021-11-01
Remove girder web components dependency

### LINKED ISSUE #670: Remove girder web components dependency
Verify first that we are no longer using it.

mvandenburgh:
Remove the Girder web components dependency from the project and configure Vuetify directly.

This PR does the following:
- Uninstalls GWC
- Installs the `vue-async-computed` package
  - This package was included with GWC, so we must now install it ourselves
  - Note that this dependency will go away after `FileBrowser.vue` is refactored to Composition API.
- Defines the Vuetify theme (colors, mostly) directly in `vuetify.ts`
  - This was another thing that GWC took care of for us. Having direct control over this will be useful as some of the theme may need to change as part of the DLP redesign.

Closes #670

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -12,6 +12,7 @@ import * as Integrations from '@sentry/integrations';
 // Import plugins first (order may matter)
 import '@/plugins/composition';
 import '@/plugins/asyncComputed';
+import vuetify from '@/plugins/vuetify';
 
 // Import custom behavior
 import '@/title';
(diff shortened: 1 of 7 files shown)


src/main.ts
```diff
@@ -22,6 +21,7 @@ import App from '@/App.vue';
 import router from '@/router';
 import store from '@/store';
 import { dandiRest } from '@/rest';
+import vuetify from './plugins/vuetify';
```
dchiquito, line 24: Shouldn't this go with the rest of the plugin imports?

dchiquito (approved):

PR #949, opened 2021-11-02
DLP Redesign - Metadata Overview Display

### LINKED ISSUE #951: Implement overview panel from DLP redesign

mvandenburgh:
TODO
- [x] hide the tab bar so that the overview is the only thing visible on the DLP
- [x] add "funding information" display to DLP overview

Closes #951


mvandenburgh:
One question this raises about the design is what we should do about overflow on each of the boxes. For example, look at this dandiset https://deploy-preview-949--gui-dandiarchive-org.netlify.app/#/dandiset/100268/draft. There's nothing stopping users from putting enough metadata in to make the boxes really long like that.

I see three options:
1) Make the boxes a fixed height and make them scrollable (seems ugly, but very simple to code)
2) Make tall boxes collapsed initially, and allow the user to extend them by clicking it (would look nice but would be complex to code, since vuetify doesn't seem to support this)
	- This introduces another consideration about what to do about the boxes next to the box being extended
3) Same as 2), except clicking opens up a popup modal that contains all of the boxes info (and is scrollable if necessary)

I'm guessing 2 and 3 are the best options UI/UX wise. ping @waxlamp @jtomeck

satra:
@mvandenburgh - i think this would really depend on the content of the boxes and how we represent it. this was the part i think that was left to us to figure out .

for example access should not be one of those columns. and should really go up to the same level as license and keywords. 

subject matter are terms of certain type they are like keywords but with more information (i.e. they may point to an external resource, etc.,. from a ui perspective subject matter could simply be a collection of balloons with the name and a click on the name can bring up the additional information. 

resources are similar, they should have names and relationships, but not all pieces in the record needs to be displayed on the DLP

also there is always likely to be an imbalance in the number of things. if so does it make sense to say display first 3 for each and then have a `+5 more` box to show the full list in a separate modal.

mvandenburgh:
I've implemented the changes discussed in last weeks UI meeting -

- Put subject matter under keywords
- Move license + access information to header next to `Created` and `Last update`
- Remove roles from funding info (they are implicit)
- Don't break cards into columns on small screens/mobile devices

mvandenburgh:
@satra the `name` field on subject matter and related resource is not required currently in the schema, but the redesign uses this field as the header for each card. So it's currently possible to create a related resource or subject matter without a name, which leads to this UI https://deploy-preview-949--gui-dandiarchive-org.netlify.app/#/dandiset/200041/draft. Is there a reason not to ha

satra:
technically the identifier is more important, but it was done to be minimalistic (either identifier or name can be provided). for purpose of the card, i would set the value to be the same as identifier if name is empty. in the future we can interrogate the identifier and fill in the corresponding label.

jtomeck:
Chiming in pretty late on this, but I think if there's a lot of data in each box, we could set a sane maximum to show, with text afterwards saying `XX more entries | link: show more / less` or something like that. Obviously it would be show more until clicking, then it would say show less.

waxlamp (approved):
Looks great! Please make a note of Jared's feedback somewhere, so that we don't lose it going forward.

PR #965, opened 2021-11-10
Update the route on publish

### LINKED ISSUE #579: DJANGO_API: still highlights released if go to "draft" right after publishing
probably because datetime is still matching a (just) released/published version to prevent "re-publishing"

image

dchiquito:
Previously, when the publish button was pressed, the store was modified
directly to show the newly published dandiset. This meant the route was
not updated, which puts the page in a weird state.

Use the router to navigate to newly published dandisets instead of
modifying the state directly.

Fixes #579

### DIFF
--- a/src/views/DandisetLandingView/DandisetPublish.vue
+++ b/src/views/DandisetLandingView/DandisetPublish.vue
@@ -305,6 +305,7 @@ import moment from 'moment';
 
 import { dandiRest, loggedIn as loggedInFunc, user as userFunc } from '@/rest';
 import store from '@/store';
+import router from '@/router';
 import { User, Version } from '@/types';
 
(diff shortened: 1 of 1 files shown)


src/views/DandisetLandingView/DandisetPublish.vue
```diff
@@ -305,6 +305,7 @@ import moment from 'moment';
 
 import { dandiRest, loggedIn as loggedInFunc, user as userFunc } from '@/rest';
 import store from '@/store';
+import router from '@/router';
```
mvandenburgh, line 308: I didn't know you could just import the router like this and use it, that seems cleaner then using the `ctx.root.$router` variable https://github.com/dandi/dandiarchive/blob/master/src/views/MetadataView/Meditor.vue#L310.

mvandenburgh (approved):
LGTM

PR #977, opened 2021-11-17
Add empty and draft toggles to DandisetsPage

dchiquito:
These toggles control whether or not empty or draft dandisets are
displayed in the search. They both default to off.

This requires https://github.com/dandi/dandi-api/pull/607 before the branch preview will work.

### DIFF
--- a/src/components/DandisetsPage.vue
+++ b/src/components/DandisetsPage.vue
@@ -1,13 +1,39 @@
 <template>
   <div v-page-title="pageTitle">
     <v-toolbar color="grey darken-2 white--text">
-      <v-toolbar-title class="d-none d-md-block">
-        {{ title }}
-      </v-toolbar-title>
-      <v-divider
-        class="d-none d-md-block"
-        vertical
-      />
+      <v-menu
+        offset-y
+        :close-on-content-click="false"
+      >
+        <template #activator="{ on, attrs }">
+          <v-icon
+            dark
+            v-bind="attrs"
+            v-on="on"
+          >
+            mdi-cog
+          </v-icon>
+        </template>
+        <v-list>
+          <v-list-item>
+            <v-list-item-title>Show:</v-list-item-title>
+          </v-list-item>
+          <v-list-item>
+            <v-switch
+              v-model="showDrafts"
+              label="Drafts"
+              dense
+            />
+          </v-list-item>
+          <v-list-item>
+            <v-switch
+              v-model="showEmpty"
+              label="Empty Dandisets"
+              dense
+            />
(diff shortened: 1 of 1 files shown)


src/components/DandisetsPage.vue
```diff
@@ -115,6 +141,8 @@ export default defineComponent({
     // https://next.router.vuejs.org/api/#useroute
     const route = ctx.root.$route;
 
+    const showDrafts = ref(false);
+    const showEmpty = ref(false);
     const sortOption = ref(Number(route.query.sortOption) || 0);
     const sortDir = ref(Number(route.query.sortDir || -1));
     const page = ref(Number(route.query.page) || 1);
```
mvandenburgh, line 148: I noticed a bug, to reproduce in the deploy preview:
- go to public dandisets
- turn on "Drafts" and "Empty Dandisets"
- Go to a page higher than 2
- turn off "Drafts" and "Empty Dandisets"

I think the fix is to just reset the `page` variable to 1 whenever `showDrafts` or `showEmpty` change.

mvandenburgh (changes requested):

src/components/DandisetsPage.vue
```diff
@@ -115,6 +141,8 @@ export default defineComponent({
     // https://next.router.vuejs.org/api/#useroute
     const route = ctx.root.$route;
 
+    const showDrafts = ref(false);
```
satra, line 144:
```suggestion
    const showDrafts = ref(true);
```

dchiquito: @satra This will show all draft dandisets with no published versions, including any test dandisets that had data uploaded to them. Is that what you want?

dchiquito: @satra This will show all draft dandisets with no published versions, including any test dandisets that had data uploaded to them. Is that what you want?

satra: by default, for now yes.

satra: once our published count hits 100, we can change the settings. we do need some way of indicating to users that they should use the settings to change display characteristics. perhaps we can do this when we get to a bit more faceted search.

dchiquito:
@jtomeck This PR is making some tweaks to the search page: https://deploy-preview-977--gui-dandiarchive-org.netlify.app/#/dandiset
I added a settings cog to enable/disable draft and empty dandisets.
Any quick changes I should make now before it goes into production?

mvandenburgh (approved):
LGTM after fixing linting errors

PR #979, opened 2021-11-22
Implement redesigned ownership management ui

### LINKED ISSUE #951: Implement overview panel from DLP redesign

mvandenburgh:
Closes #951.


src/components/DLP/DandisetOwnersDialog.vue
```diff
@@ -0,0 +1,302 @@
...
+      type: Array,
+      required: true,
+    },
+  },
+  setup() {
+    const currentDandiset = computed(() => store.state.dandiset.dandiset);
+    const owners = computed(() => store.state.dandiset.owners);
+
+    const searchQuery = ref('');
+    const loadingUsers = ref(false);
```
dchiquito, line 187: I don't see that this is used anywhere.

src/components/DLP/DandisetOwnersDialog.vue
```diff
@@ -0,0 +1,302 @@
...
+      _searchResults.value = users;
+      loadingUsers.value = false;
+    }, 200);
+
+    watch(() => owners.value, () => (
+      owners.value ? Object.assign(newOwners.value, owners.value) : null
+    ),
+    { immediate: true });
+
+    const isPrimaryOwner = (user: User) => newOwners.value?.length && newOwners.value[0] === user;
```
dchiquito, line 220: What does being a primary owner entail? It looks like they get a special icon and they can't be deleted. Does that mean there's no way for the dandiset creator to abdicate ownership?

dchiquito (approved):
couple of questions, overall I think this is OK

src/components/DLP/DandisetOwnersDialog.vue
mvandenburgh: My bad, I meant to put a comment explaining that. It's in Jared's design, but we don't support it yet. But, I think it might be a good idea.
Currently we don't have any notion of a "primary owner", but we do have a constraint that a dandiset must have at least one owner at all times. So right now, attempting to remove the last owner of a dandiset will fail. Maybe instead, it would be cleaner/better UX to have an explicitly designated "primary owner" that can never be removed. (any thoughts @waxlamp @satra @yarikoptic ?)

satra: i would for the moment keep it the way it is. i.e. owners have no differences in status. hence there is no concept of a primary owner, just a list of owners. any primariness is social within the group/project  that created the dandiset. i.e. it's just like our roles as admins on the archive.

waxlamp: I agree with Satra; we shouldn't have the "primary owner" concept and treat all owners equally. This does mean we have to check at runtime and disallow removing a sole owner, but that also means transferring ownership is as simple as adding the transferree, then removing oneself from ownership.

mvandenburgh: Sounds good, I removed the UI elements related to this

PR #983, opened 2021-11-29
Fix pip caching in CI

### LINKED ISSUE #982: Fix pip wheel caching in CI
The GitHub actions CI workflow doesn't cache pip wheels properly; doing so would save us about a minute and a half per CI run (which I think is valuable considering how long the E2E tests take already)

mvandenburgh:
The GitHub actions CI workflow doesn't cache pip wheels properly; currently, everything is downloaded and rebuilt from scratch on every PR or push to master. This PR fixes the CI config file so it caches the pip environment properly. This saves about a minute and a half for each CI run.

Comparison without/with caching:
https://github.com/dandi/dandiarchive/actions/runs/1517816471/attempts/1
https://github.com/dandi/dandiarchive/actions/runs/1517816471/attempts/2

Closes #982

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,32 +61,24 @@ jobs:
         with:
           python-version: '3.8'
 
-      - name: Upgrade pip
-        run: pip install -U pip
-
-      - name: Install Python packages
+      - name: Install latest version of pip
         run: pip install --upgrade pip
 
       - name: Clone dandi-api
         run: git clone https://github.com/dandi/dandi-api.git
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -106,7 +98,6 @@ jobs:
           done
 
           # activate virtual env and start the dandi-api server
```
dchiquito, line 100:
```suggestion
          # start the dandi-api server
```

dchiquito (approved):

PR #984, opened 2021-11-29
Fix e2e tests

mvandenburgh:
Tests are failing due to https://github.com/dandi/dandi-api/pull/590. This PR fixes the tests by accounting for the new form that was added.

### DIFF
--- a/test/src/util.js
+++ b/test/src/util.js
@@ -18,6 +18,9 @@ export function uniqueId() {
   return Date.now().toString();
 }
 
+export const TEST_USER_FIRST_NAME = 'Test';
+export const TEST_USER_LAST_NAME = `User_${uniqueId()}`;
+
 /**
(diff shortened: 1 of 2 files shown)


test/src/util.js
```diff
@@ -18,6 +18,9 @@ export function uniqueId() {
   return Date.now().toString();
 }
 
+export const TEST_USER_FIRST_NAME = 'Test';
+export const TEST_USER_LAST_NAME = `User_${uniqueId()}`;
```
dchiquito, line 22: You might consider adding a `TEST_USER_INITIALS` so you don't have to sprinkle `${TEST_USER_FIRST_NAME.charAt(0)}${TEST_USER_LAST_NAME.charAt(0)}` everywhere

mvandenburgh: Ah good catch, I'll add that in #985

PR #994, opened 2021-12-15
Add test for dandiset identifier incrementing

### LINKED ISSUE #993: Create test to test for dandiset id increment bug
This would be an e2e test that creates a bunch of dandisets through the web ui while checking that identifiers are being assigned correctly.

mvandenburgh:
Closes #993.

This creates 1000 dandisets right now across two users. It might be a good idea to lower that number to decrease time it takes to run CI; I'm going to let it run first and see how long it takes.

### DIFF
--- a/test/src/tests/dandisetIdentifiers.test.js
+++ b/test/src/tests/dandisetIdentifiers.test.js
@@ -13,7 +13,7 @@ describe('ensures dandiset identifiers are incremented correctly', () => {
   it('creates a bunch of new dandisets while verifying that their IDs are correct', async () => {
     jest.setTimeout(60000 * 30); // Temporarily increase timeout to 30 mins for this test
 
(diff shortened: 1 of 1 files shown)


mvandenburgh:
As discussed, we're not going to make this test one of our regularly running E2E tests. Instead, I will run this once against the staging instance and see if any more increment errors happen.

waxlamp:
> As discussed, we're not going to make this test one of our regularly running E2E tests. Instead, I will run this once against the staging instance and see if any more increment errors happen.

Could you report your findings here?

mvandenburgh:
> > As discussed, we're not going to make this test one of our regularly running E2E tests. Instead, I will run this once against the staging instance and see if any more increment errors happen.
> 
> Could you report your findings here?

I executed this test on the staging instance and everything looks good, all identifiers were correct.

PR #1000, opened 2021-12-21
Add embargoed/award number options to create dandiset page

### LINKED ISSUE #995: Update dandiset creation view with embargo options
Offer the user a checkbox and a text field for the NIH award number, and send this data to the backend.

[design doc](https://github.com/dandi/dandi-api/blob/96c85ca5b3c1932b450cb00d4c375f0a176e2261/doc/design/embargo-mvp.md?plain=1#L10-L12)

mvandenburgh:
Adds checkbox to embargo a new dandiset + a text field to enter the award number in. 

Resolves #995


waxlamp:
This should probably be targeted to an embargo-mvp feature branch. Thoughts on that @dchiquito?

dchiquito:
I think it wouldn't really be an issue to just leave this PR unmerged until we're ready to release embargo (we can merge the other PRs without breaking anything, since users wouldn't be able to create embargoed dandisets without this PR).

The only reasons I think we'd want a separate branch is to close this PR (just to keep the PR list shorted), or so we have a branch that we can easily check out to create embargoed dandisets for development purposes. We can always just check out this branch `create-embargoed-dandiset`, so I don't see a great reason for a feature branch.

src/rest.ts
```diff
@@ -162,10 +163,31 @@ const dandiRest = new Vue({
       const response = await client.get('dandisets/', { params });
       return response;
     },
-    async createDandiset(name: string, description: string): Promise<AxiosResponse<Dandiset>> {
-      const { schema_version } = await this.info();
-      const metadata = { name, description, schemaVersion: schema_version };
-      return client.post('dandisets/', { name, metadata });
+    async createDandiset(
+      name: string, description: string, embargoed: boolean, awardNumber: IdentifierForAnAward,
+    ): Promise<AxiosResponse<Dandiset>> {
```
dchiquito, line 168: Generally speaking, I think creating new functions like `createEmbargoedDandiset` is a better design pattern than adding switch arguments like `embargoed` to existing functions. More complicated functions are less testable and harder to read, and adding more arguments scales complexity exponentially.

In this case there is some duplicated code, so to keep things DRY perhaps `createDandiset(...)` could take optional `metadata` and `params` arguments that are empty if unspecified. Then `createEmbargoedDandiset` would call `createDandiset` with appropriately initialized `metadata` and `params`.

waxlamp:
> I think it wouldn't really be an issue to just leave this PR unmerged until we're ready to release embargo (we can merge the other PRs without breaking anything, since users wouldn't be able to create embargoed dandisets without this PR).
> 
> The only reasons I think we'd want a separate branch is to close this PR (just to keep the PR list shorted), or so we have a branch that we can easily check out to create embargoed dandisets for development purposes. We can always just check out this branch `create-embargoed-dandiset`, so I don't see a great reason for a feature branch.

I agree if this is the only branch in the frontend that will be needed for embargo MVP. This is a complex enough venture that I'm not comfortable standing by such a claim.

There's a subtle reason to prefer merging PRs that are ready to go, rather than just leaving them hanging out: it reduces uncertainty in what is a rather sprawling set of coordinated tasks that need to land together to get the MVP done. That is, we don't have to go hunting for unmerged PRs when we are ready to deliver. This is more of a PHB concern than an engineer one, so I'll defer to you guys, but even so my advice is to get it merged (somewhere) if possible.

src/views/CreateDandisetView/CreateDandisetView.vue
```diff
@@ -3,31 +3,59 @@
...
+          outlined
+          class="my-4"
+        />
+        <v-checkbox
+          v-model="embargoed"
+          hide-details
+          class="shrink mr-2 mt-0"
+        >
+          <template #label>
+            Embargo this dataset?
```
dchiquito, line 33: We need a blurb here, perhaps in a mouseover tooltip. Casual users need enough information to understand that they should not check this unless they have an NIH grant.
We should also clarify that any award number can be specified at this time if there are multiple, and that you can change your answer later.

@waxlamp any suggestions?

src/views/CreateDandisetView/CreateDandisetView.vue
```diff
@@ -42,29 +70,57 @@
...
+    const embargoed = ref(false);
+    const awardNumber = ref('');
+    const saveDisabled = computed(
+      () => !name.value
+      || !description.value
+      || (embargoed.value && !awardNumberValidator(awardNumber.value)),
+    );
+
+    const awardNumberRules = computed(
+      () => [(v: string) => awardNumberValidator(v) || VALIDATION_FAIL_MESSAGE],
```
dchiquito, line 100: Is there any reason not to just define `awardNumberValidator` as `return NIH_AWARD_REGEX.test(awardNumber) || VALIDATION_FAIL_MESSAGE` so you don't need this wrapper?
Also, I don't see anything that would necessitate this being a computed property. I don't actually know if you can just return normal objects from `setup` or if it has to wrapped in a `ref`.

dchiquito (changes requested):

src/rest.ts
waxlamp: Also, I couldn't pick just one, but this is worth [thinking about](https://www.google.com/search?q=don%27t+use+boolean+parameters&oq=dont+use+boolean+parameters).

src/views/CreateDandisetView/CreateDandisetView.vue
mvandenburgh: > Also, I don't see anything that would necessitate this being a computed property. I don't actually know if you can just return normal objects from `setup` or if it has to wrapped in a `ref`.

The docs for `v-text-field` specify it as a computed property. If `computed` is removed validation no longer works.

> Is there any reason not to just define `awardNumberValidator` as `return NIH_AWARD_REGEX.test(awardNumber) || VALIDATION_FAIL_MESSAGE` so you don't need this wrapper?

Yes, because both the `awardNumberRules` and `saveDisabled` computed vars need to use it.

src/views/CreateDandisetView/CreateDandisetView.vue
waxlamp: Did I not ever respond to this? I could have sworn I wrote out a full answer...

Here's a first attempt at a blurb:

```
Embargoed datasets are hidden from public access until a specific time period has elapsed. Uploading data to the DANDI archive under embargo requires a relevant NIH award number, and the data will be automatically published when the embargo period expires.
```

waxlamp (commented):
LGTM. I'm deferring to Daniel for a binding review.

mvandenburgh:
I've wrapped the embargo UI elements in a `v-if=false` so we can merge this before we combine the repos, as @roni and I discussed.

waxlamp (approved):

PR #1005, opened 2021-12-29
Use correct casing of `dandi` in download instructions

### LINKED ISSUE #957: Web interface listing invalid download commands for specific versions
If one views the page on gui.dandiarchive.org for a Dandiset with one or more published versions (e.g., https://gui.dandiarchive.org/#/dandiset/000023), clicks on the "Download" button, selects "Download a different version", selects "Other version" and then any non-draft version, a command like the following is then displayed:

    dandi download DANDI:000023/0.210914.1900

However, this command will not work, as the client attempts to resolve `DANDI:*` resource identifiers by prepending `https://identifiers.org/` to them, and https://identifiers.org/DANDI:000023/0.210914.1900 does not exist.

EDIT: [According to @satra](https://github.com/dandi/helpdesk/issues/37#issuecomment-962037772), the identifiers.org URL is invalid due to a bug on their end, and it becomes valid by lowercasing "DANDI:".

waxlamp:
Fixes #957.

(@satra, @jwodder: is this the correct fix?)


satra:
this would be the fix. but identifiers.org has fixed resolving both upper and lowercase. so in that sense the original issue is a non-issue. i don't know which one would break in the future.

waxlamp:
> this would be the fix. but identifiers.org has fixed resolving both upper and lowercase. so in that sense the original issue is a non-issue. i don't know which one would break in the future.

I would be fine with abandoning this PR, as long as we also close the linked issue. Let me know if that's ok, @satra.

Issue #1008, opened 2022-01-03
Finish typescript refactor


mvandenburgh:
This supersedes #855 by outlining specifically what is left in terms of converting the existing codebase to typescript.

- [ ] Convert file browser component to typescript
- [ ] Convert "utility" js files (`router.js`, `migrate.js`, etc) to typescript
- [ ] Enable type checking in vue templates (#987)
  - This is blocked on converting the file browser component

(I had a TODO written down to file this a long time ago but I can't find an existing issue for it, I guess I never got around to it)

PR #1009, opened 2022-01-03
Convert outstanding .js files to typescript

mvandenburgh:
Partly addresses #1008.


mvandenburgh:
Netlify is reporting build failures for the deploy preview :thinking: .... I'll mark this as draft for now

src/rest.ts
```diff
@@ -6,19 +6,25 @@ import {
...
+
 // Ensure contains trailing slash
 const dandiApiRoot = process.env.VUE_APP_DANDI_API_ROOT.endsWith('/')
   ? process.env.VUE_APP_DANDI_API_ROOT
   : `${process.env.VUE_APP_DANDI_API_ROOT}/`;
 
 const client = axios.create({ baseURL: dandiApiRoot });
 
-let oauthClient: OAuthClient|null;
+let oauthClient: OAuthClient|null = null;
```
dchiquito, line 20:
```suggestion
let oauthClient?: OAuthClient = null;
```
I think this is the magic syntax

dchiquito (approved):

src/rest.ts
mvandenburgh: Unfortunately `| null` looks like the only way to do it. https://github.com/microsoft/TypeScript/issues/13321

mvandenburgh:
Marking this as draft - I need to transfer this over to the dandi-api repo

src/rest.ts
dchiquito: sad :'(

